Issue #73, opened 2017-12-05
Handling owner/username in portal API calls


noahmulfinger:
When working with user content, its common that we have to use an endpoint similar to
```
https://www.arcgis.com/sharing/rest/content/users/${username}/...
```
which requires getting the user's username. The `items` package handles this by having the user input an `owner` parameter for functions that require it. However, in the `feature-services-admin` package I've been working on, I've been assuming that the authentication is a UserSession and pulling the username from `requestOptions.authentication.username`. 

Is it always the case that *user* authentication is needed when calling an endpoint like `https://www.arcgis.com/sharing/rest/content/users/${username}/...`? If so, I think pulling from UserSession makes more sense. However, this would make `requestOptions` required, which I'm not sure matches the expected use of `requestOptions`.

I want to get people's thoughts on this so we have a consistent pattern. @patrickarlt @dbouwman @tomwayson @jgravois @araedavis 

As a side note, issues like this make me think it would be good to set up an API design reference in order to keep track of cross-package design considerations.

jgravois:
I had the same reaction you did when i saw that `owner` was a discrete parameter in the items package.

> Is it always the case that user authentication is needed when calling an endpoint like `https://www.arcgis.com/sharing/rest/content/users/${username}/...`?

the REST documentation for _some_ item operations (like the ones pertaining to [resources](http://resources.arcgis.com/en/help/arcgis-rest-api/index.html#/Item_Resources/02r3000002r3000000/) that i mucked around with the other day in #70) state that operations can be performed by either their owner **or** organization administrators. i'll leave it to @dbouwman to confirm whether the API design aims to purposely exploit that idiosyncrasy or whether there's an opportunity to DRY them up.

either way, a guide page that discusses API design would be rad.

patrickarlt:
I originally intended modules to override `IRequestOptions` if they needed to add additional options to a request. So we could write things like this:

```ts
import { request, IRequestOptions, IParams } from "@esri/arcgis-rest-request";
import { IItem } from "@esri/arcgis-rest-common-types"

interface AddItemToFolderOptions extends IRequestOptions {
  owner?: string;
}

export function addItemToFolder (item: IItem, folder: string, params: IParams,  requestOptions: AddItemToFolderOptions): Promise<IItem> {
  const owner = requestOptions.owner || requestOptions.session.username;
  return request(/** ... **/)
}
```

The general design for request is as follows (we should probally put this in an API design page somewhere):

```ts
[...]

PR #74, opened 2017-12-06
refactor(items): make passing an owner optional in item methods

jgravois:
one option to address @noahmulfinger's excellent feedback in #73

AFFECTS PACKAGES:
@esri/arcgis-rest-items

### DIFF
--- a/packages/arcgis-rest-auth/src/AuthenticatedRequestOptions.ts
+++ b/packages/arcgis-rest-auth/src/AuthenticatedRequestOptions.ts
@@ -1,0 +1,25 @@
+import { ApplicationSession } from "./ApplicationSession";
+import { UserSession } from "./UserSession";
+
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+
+/**
+ * Used internally by packages for requests that require authentication.
+ */
+export interface IAuthenticatedRequestOptions extends IRequestOptions {
+  authentication: UserSession | ApplicationSession;
+}
+
+/**
+ * Used internally by packages for requests that require user authentication.
+ */
+export interface IUserRequestOptions extends IRequestOptions {
+  authentication: UserSession;
+}
+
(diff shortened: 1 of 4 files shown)


patrickarlt:
@jgravois I actually think this sets us up with a bad pattern for routes which will require additional params for the URL but no auth: because then you have to do this:

```ts
getItemFromFolder("itemId", null, null, "folder");
```

since you will have to pass `null` or `{}` for `requestOptions` and maybe params just to get to pass in a folder.

jgravois:
> Seems like `IUserSessionRequestOptions` will be needed in other packages too, right?

i assume it will. my first re-useable instinct would be to move it into `auth` because `common-types` doesn't currently have any external dependencies of its own.

@patrickarlt as far as i know, all the methods with `owner/owner?` definitely  _require_ the request to be authenticated.

patrickarlt:
if we are going to require each thing in the URL segment things could get out of hand really fast and we should just consider switching to passing in a single object:

```ts
createItem({
  item: IItem,
  owner: "patrickarlt",
  params: {
    /* any other params */
  },
  authentication: mySession
})
```

and switch `request` to 

```ts
request({
  url: "http://",
  params: {
    /* whatever */
  }
  authentication: /* whatever */,
  httpMethod: "POST"
})
```

patrickarlt:
@jgravois for example [reassignItem](https://resources.arcgis.com/en/help/arcgis-rest-api/#/Reassign_Item/02r300000097000000/) needs

* `tagetUsername` (required)
* `targetFoldername` (optional)
* `username` (required, part of URL)
* `itemId` (required, part of URL)
* `authentication` (current user may or may not be owner)

I really think we might be better off switching to objects for all of this stuff...

jgravois:
@patrickarlt makes some good points. closing in favor of #78

Issue #19, opened 2017-08-09
Setup headless Chrome and Firefox tests on Travis


patrickarlt:
Once we open source the repo we should have Travis CI run our Node tests on Node 6 and Node 8 and our browser tests through Sauce Labs https://wiki.saucelabs.com/display/DOCS/Setting+Up+Karma+for+JavaScript+Unit+Testing for all browsers that we support.

It would also be nice to use [CodeCov](https://codecov.io/) or [Coveralls](https://coveralls.io/) to automatically upload the coverage files generated by Karma.

patrickarlt:
You now have to apply for SauceLabs to get it for free so it might be better off just using headless Firefox and Chrome on travis to automate browser testing. https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai#running_it_all_on_travis_ci

patrickarlt:
There are also additional docs at https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-the-Firefox-addon-in-headless-mode.

jgravois:
i'm particularly interested in wiring up a bot that replies to each open PR indicating the impact on test coverage so i'll look into that part.

patrickarlt:
@jgravois [Coveralls](https://coveralls.io/) (what I use for Acetate) will do that https://github.com/patrickarlt/acetate/pull/70#issuecomment-341254040.

patrickarlt:
Updated the title of this issue since @jgravois got coveralls hooked up.

PR #76, opened 2017-12-06
convince Travis to just run the test suite in Node.js

jgravois:
and leave browser tests to SauceLabs (as per #19)


jgravois:
i like seeing 📗 📗 📗 📗 📗 📗 📗 📗 📗 📗 📗

Issue #66, opened 2017-11-28
smart support for GET (which switches to POST automatically)?


jgravois:
our current implementation of esriRequest defaults to POST and allows a GET to be issued manually.

@patrickarlt's preference is to recommend that packages default to POST and skip implementing logic to switch from a GET to a POST after the commonly accepted ~2000 character limit has been exceeded.

this issue is intended to track interest from other developers who would like to piggyback off a _smart_ request method instead of defaulting to POST (or implementing the logic above in their own package).

anecdotally, it didn't take me long to identify an ArcGIS Server operation that _appears_ to only support GET

* http://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/dynamicLayer
* http://resources.arcgis.com/en/help/arcgis-rest-api/index.html#/Dynamic_Layer_Table/02r3000000q0000000/

tomwayson:
I just bumped into an issue where koop services don't support POST. Also, debugging the POST query to figure that out was not as easy as debugging a GET - not hard, just inconvenient.

I really think defaulting to POST is going to cause more problems than it's worth. I know _these libraries_ don't support non-CORs browsers, and I guess we won't officially support services that don't have CORs on or accept POST, _but_ these libraries will be used in applications that do have to support those scenarios. I worry that there will be dozens of consuming libraries that will have to implement the same < 2000 char limit logic, and it seems like that's better dealt w/ in a centralized place.

I'm still unsure if we should have `request()` do this, or have the individual libs handle it. I prefer the former. Whichever we decide, I'm willing to PR.

tomwayson:
@patrickarlt do you agree that this should be settled before v1 release?

jgravois:
personally, i'm okay with shipping a v1 that defaults to `POST` and allows devs to `GET` or incorporate their own logic to switch between the two.

with something 'smart' i assumed we'd be aiming for (another?) wrapper to switch between the two for them automagically so i don't envision it being a breaking change.

tomwayson:
So after some back and forth, w/ @jgravois, it seems the [2k limit might be on the server side](https://stackoverflow.com/questions/15090220/maximum-length-for-url-in-chrome-browser). Looking at https://github.com/Esri/cedar/pull/346, the service in question is a hosted feature service.

I think the hassle of debugging POST requests and the awkward HTTP verb semantics of POSTing for queries, and warrant defaulting to GET for those requests.

[...]

PR #77, opened 2017-12-08
Smart usage of FormData

patrickarlt:
This PR implements logic for what I proposed in https://github.com/Esri/arcgis-rest-js/issues/66#issuecomment-350107219 which uses form encoded were possible so thing display nicely in the debugger.

screenshot 2017-12-07 13 56 00


support/encoded-param-helper.ts
```diff
@@ -0,0 +1,3 @@
+export function encodedParam(key: string, value: any) {
```
jgravois: lets include this in `request/utils` instead. i can handle it if you don't get to it first.

packages/arcgis-rest-request/src/utils/process-params.ts
```diff
@@ -1,6 +1,42 @@
 /* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
+/**
+ * Checks parameters to see if we should use FormData to send the request
+ * @param params The object whose keys will be encoded.
+ * @return A boolean indicating if FormData will be required.
+ */
+export function requiresFormData(params: any) {
+  return Object.keys(params).every(key => {
```
tomwayson: Maybe I'm reading this wrong, but it seems like this should be `.some()` instead of `.every()`, no?

packages/arcgis-rest-request/test/utils/process-params.test.ts
```diff
@@ -97,4 +97,71 @@ describe("processParams", () => {
...
+    ).toBeFalsy();
+
+    expect(
+      requiresFormData({
+        falsy: null
+      })
+    ).toBeFalsy();
+  });
+
+  it("should require form data for multipart requests", () => {
```
tomwayson: I think a test like the following would fail, but should pass b/c `requiresFormData()` uses `.every()` instead of `.some()`:

```js
    expect(
      requiresFormData({
        string: "string",
        binary: binaryObj
      })
    ).toBeTruthy();
```

patrickarlt:
@tomwayson good catch. I added a new test for this and fixed the implementation.

@jgravois done. That makes things much easier.

PR #80, opened 2017-12-21
fix(release-automation): fix issues uncovered by 1st release

patrickarlt:
I ran into a whole bunch of issues with `npm publish` trying to get 1.0.0 out:

1. Scoped packages are private by default which requires a paid account, which I did not have, this meant some changes in `package.json`s to make Lerna publish the packages as public.
2. `npm publish` runs `npm prepare` which did not work for some reason, maybe because `process.cwd` changed? This has been replaced by a global `lerna bootstrap` across all packages before `npm run release:prepare`.
3. Fixes a few other things in the publish script like making sure the `package.json` version is included in the release tag and the zip file gets cleaned up after `gh-release`.

---

* Remove NPM  prepare scripts, which run before npm publish but break because we are setup w/ lerna.
* Add `publishConfig` so lerna publishes the scoped packages as public, otherwise scoped packages are private by default and will fail to publish.
* `lerna bootstrap` and `npm test` will now be run before `npm run release:prepare` to build and test all the libraries.
* Some small tweaks were made to the publish script to pick up the `package.json` versioning and the remove the ZIP archive after it is published.


tomwayson:
But looks like Travis disagrees w/ me - perhaps related to removing `prepublish` from packages b/c that used to get run when `lerna bootstrap` linked them?

support/publish.sh
```diff
@@ -1,27 +1,35 @@
...
 git add --all
 git commit -am "Prepare v$VERSION" --no-verify
 
 # incriment the package.json version to the lerna version so gh-release works
-npm version $VERSION
+npm version $VERSION --allow-same-version
+
+# amend the changes from `npm version` to the release commit
+git add --all
+git commit -am "Prepare v$VERSION" --no-verify --amend
```
noahmulfinger: I think these two lines (14,15) are unnecessary. `npm version $VERSION --allow-same-version` already adds a commit named `$VERSION`, so there are no uncommitted files to clean up. When lines 14 and 15 are run, that commit is renamed `Prepare v$VERSION`, resulting in two `Prepare v$VERSION` commits.

jgravois: good catch @noahmulfinger!

PR #82, opened 2017-12-22
Fix more release automation issues

patrickarlt:
While testing release automation I accidentally published v1.0.1 to NPM. This PR updates all teh necessary files. Once this gets merged I'll publish 1.0.2 and deprecate 1.0.1 to finish cleaning stuff up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "precommit": "lint-staged",
     "bootstrap": "lerna bootstrap",
     "postinstall": "npm run bootstrap",
-    "prerelease:prepare": "lerna bootstrap && npm test",
+    "prerelease:prepare": "git fetch --all && lerna bootstrap && npm test",
     "release:prepare": "lerna publish --skip-git --skip-npm --yes && node ./support/changelog.js",
     "release:review": "git --no-pager diff --word-diff",
     "release:publish": "./support/publish.sh",
--- a/support/publish.sh
+++ b/support/publish.sh
@@ -21,22 +21,6 @@ git tag v$VERSION
 git push https://github.com/Esri/arcgis-rest-js.git master
 git push --tags
 
-# checkout temp branch for release
-git checkout -b release-v$VERSION
-
-# add built files to the release commit
-git add packages/*/dist -f
-
-# commit the built files
-git commit -am"Publish v$VERSION" --no-verify
-
-# tag the release
-git tag v$VERSION
-
-# push the release commit and tag to github
-git push https://github.com/Esri/arcgis-rest-js.git release-v$VERSION
-git push --tags
-
 # publish each package on npm
 lerna publish --skip-git --yes --repo-version $VERSION
 
@@ -52,8 +36,3 @@ gh-release --t v$VERSION --repo arcgis-rest-js --owner Esri -a $TEMP_FOLDER.zip
 
 # Delete the ZIP archive
 rm $TEMP_FOLDER.zip
-
-# checkout master and delete release branch locally and on GitHub
-git checkout master
-git branch -D release-v$VERSION
-git push upstream :release-v$VERSION
--- a/support/publish.sh
+++ b/support/publish.sh
@@ -5,14 +5,16 @@ VERSION=$(node --eval "console.log(require('./lerna.json').version);")
 
 # commit the changes from `npm run release:prepare`
 git add --all
-git commit -am "Prepare v$VERSION" --no-verify
+git commit -am "v$VERSION" --no-verify
 
 # incriment the package.json version to the lerna version so gh-release works
-npm version $VERSION --allow-same-version
+npm version $VERSION --allow-same-version --no-git-tag-version
 
 # amend the changes from `npm version` to the release commit
 git add --all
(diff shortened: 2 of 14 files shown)


support/publish.sh
```diff
@@ -5,19 +5,38 @@ VERSION=$(node --eval "console.log(require('./lerna.json').version);")
...
+git commit -am "v$VERSION" --no-verify
 
 # incriment the package.json version to the lerna version so gh-release works
-npm version $VERSION --allow-same-version
+npm version $VERSION --allow-same-version --no-git-tag-version
 
 # amend the changes from `npm version` to the release commit
 git add --all
-git commit -am "Prepare v$VERSION" --no-verify --amend
+git commit -am "v$VERSION" --no-verify --amend
```
jgravois: the commit messages have been tweaked, but i'm fairly certain that the manual commit @noahmulfinger pointed out in https://github.com/Esri/arcgis-rest-js/pull/80/files#r158411095) is still superfluous.

now we have:

> `npm version` bumps the root package.json based on what is found in lerna.json (and doesn't need to be followed with an additional `git commit`.

```
commit 738a97723ef381314bcd005d95cd9114ac593039 (HEAD -> fuckingaround)
Author: john gravois <jgravois@esri.com>
Date:   Fri Dec 22 09:37:55 2017 -0800

    v1.0.1
```

> `npm run release:prepare` bumps the version of the individual packages and demos. i think we'd be better off leaving 'Prepare' in the commit message 

```
commit ad10444ef828bb3e5e6ce0ad4ca01b1e52f2d83a
Author: john gravois <jgravois@esri.com>
Date:   Fri Dec 22 09:37:30 2017 -0800

    v1.0.1

File: support/publish.sh [hunk 2]
```diff
@@ -5,19 +5,38 @@ VERSION=$(node --eval "console.log(require('./lerna.json').version);")
...
 
-# push the changes and tag to github
+# tag this version
+git tag v$VERSION
+
+# push everything up to this point to master
 git push https://github.com/Esri/arcgis-rest-js.git master
```
jgravois: i didn't know you could point at a bare url instead of using `upstream`, `origin` or another moniker. very cool!

support/publish.sh
noahmulfinger: @jgravois The `--no-git-tag-version` flag should prevent that from happening. I manually ran everything up to before the push to master, and I only end up with one commit.

support/publish.sh
```diff
@@ -5,19 +5,38 @@ VERSION=$(node --eval "console.log(require('./lerna.json').version);")
...
+git checkout -b release-v$VERSION
+
+# add built files to the release commit
+git add packages/*/dist -f
+
+# commit the built files
+git commit -am"Publish v$VERSION" --no-verify
+
+# tag the release
+git tag v$VERSION
```
noahmulfinger: @patrickarlt Not sure if this is because I didn't run the `git push` steps when testing, but this second `git tag` command fails for me due to there already being an existing tag.

support/publish.sh
patrickarlt: Yeah this makes it so you dont have to have `origin` or `upstream` remotes setup correctly

support/publish.sh
patrickarlt: @noahmulfinger you need to run `npm run release:prepare` which will incriment the `version` key in `lerna.json` otherwise it will read the current version which should have a tag already.

support/publish.sh
patrickarlt: @jgravois right the script now manually runs `git tag` a little later.

package.json
```diff
@@ -90,7 +90,7 @@
     "precommit": "lint-staged",
     "bootstrap": "lerna bootstrap",
     "postinstall": "npm run bootstrap",
-    "prerelease:prepare": "lerna bootstrap && npm test",
+    "prerelease:prepare": "git fetch --all && lerna bootstrap && npm test",
```
patrickarlt: @jgravois @noahmulfinger I also just added this to make sure all changes are pulled down locally before making a release.

support/publish.sh
noahmulfinger: I ran `npm run release:prepare` before running the publish commands. The issue I'm seeing is due to the fact that there's a `git tag v$VERSION` on line 18 and line 34.

jgravois: agreed. i was wrong.

package.json
jgravois: but fetching them doesn't integrate anything into the currently checked out branch...

support/publish.sh
patrickarlt: @noahmulfinger should be fixed now.

package.json
patrickarlt: but it will fetch the remote tags, which is needed for the changelog script to function since the changelog script assumes that you have all the tags locally. So if I publish 1.0.2 and you publish 1.0.3 you need to have to 1.0.2 tag locally otherwise the changelog script will pull commits from the wrong version.

jgravois: ahh. gotcha. thanks for clarifying.

PR #84, opened 2017-12-23
docs(fix dead links): fixed broken links within declaration page content

### LINKED ISSUE #81: add README to each individual package (for npm)
[ref:](https://www.npmjs.com/package/@esri/arcgis-rest-request)

### @esri/arcgis-rest-request <kbd>public</kbd>
Sadly, this package has no readme. Go write one!

jgravois:
fixes a few dozen dead links within the page body of individual declarations.

#### full disclosure: i just blindly added `BASE_URL +` to everything that looked like a link.

after selectively staging a few files and running `npm run c` i noticed that it bundles its own `git add --all` command. i think it'd be preferable if it didn't.

also including READMEs to resolve #81 and fixed a couple typos


jgravois:
@noahmulfinger you just convinced me that we're better off not duplicating that (hopefully) growing list in so many places.

just pushed another commit which trims them from the individual package doc and added `common-types` to the README in the root.

PR #87, opened 2017-12-26
prepend base url to API reference search results

jgravois:
STR:
1. launch https://esri.github.io/arcgis-rest-js/api/
2. search for `h` and select `HTTPMethods` from the list of suggestions

if anyone has a more _elegant_ solution than checking `window.location.origin` in a Vue component, i'm all 👂 s


noahmulfinger:
@jgravois @patrickarlt Changed this slightly so the base url is pulled from acetate instead of recreated in the Vue component.

jgravois:
much cleaner. thanks @noahmulfinger!

Issue #88, opened 2017-12-27
Automatically deploy to gh-pages when merging into master


patrickarlt:
Details are in https://docs.travis-ci.com/user/deployment/pages/. This should make it do we don't have to build the site locally to to update GH pages just merge it into master. This also means that the travis CI command will have to build the docs in addition to running the tests.

PR #90, opened 2017-12-27
have travis push doc updates when a release is tagged

jgravois:
second attempt at #88 

it seems more appropriate to only have travis build the doc and push updates to the live site when a release is tagged.

this will require us to update the site manually when a doc bug has been patched. but i think this is preferable to having new API documentation land too soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,13 @@ before_script:
   - npm run bootstrap
 script:
   - npm run test:node
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
+  on:
+    branch: master
+  local-dir: docs/build
+  target_branch: gh-pages # default
+before_deploy:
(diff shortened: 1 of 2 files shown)


patrickarlt:
This is actually kinda redundant since `postrelase:publish` will deploy the docs after a release so we don't need travis to do it https://github.com/Esri/arcgis-rest-js/blob/master/package.json#L100

patrickarlt:
I just liked the idea of the docs site building whenever we merged to `master` so we don't have to build fixes manually.

jgravois:
> I just liked the idea of the docs site building whenever we merged to `master`

thats good enough for me. especially because most of the PRs lately have been doc fixes.

patrickarlt:
@jgravois do you want to remove `postrelase:publish` too since travis will now build the docs when the publish script pushes to `master`?

jgravois:
> do you want to remove `postrelease:publish` too?

done (and squashed)

PR #94, opened 2017-12-28
hook up coveralls

jgravois:
this PR gets headless chrome tests running on travis and should hook up a coveralls integration.

i had trouble with the relevant encrypted environment variables when i tested against [my fork](https://github.com/jgravois/arcgis-rest-js/pull/1), but i figured i'd take one more crack at it. 

ref: https://coveralls.io/github/Esri/arcgis-rest-js

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,2 @@
 git service_name: travis-ci
-repo_token: $COVERALLS_TOKEN
+repo_token: avn7rbDP1caleMbxtVvI84YrAQxlQaBsp
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+git service_name: travis-ci
+repo_token: $COVERALLS_TOKEN
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,0 @@
-git service_name: travis-ci
-repo_token: avn7rbDP1caleMbxtVvI84YrAQxlQaBsp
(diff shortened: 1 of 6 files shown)


.coveralls.yml
```diff
@@ -0,0 +1,2 @@
+git service_name: travis-ci
+repo_token: avn7rbDP1caleMbxtVvI84YrAQxlQaBsp
```
jgravois: to be perfectly honest, based on my reading of the coveralls doc, this configuration file [shouldn't be required at all](https://docs.coveralls.io/supported-ci-services). 

that said, i'd be satisfied to let the mystery lie if i could just get a successfull build with:
```yaml
repo_token: $COVERALLS_TOKEN` 
```

patrickarlt: I ended up not setting this up like this. I used https://www.npmjs.com/package/coveralls for acetate and just dumped the coverage output to it at the end of the travis CI run like this https://github.com/patrickarlt/acetate/blob/master/.travis.yml#L11

The standard `npm run test:chrome` should already emit a `lcovonly` file in the coverage folder you can pipe to `coveralls`.

patrickarlt: @jgravois I think `karma-coveralls` wants it to be `COVERALLS_REPO_TOKEN` see their usage section https://github.com/caitp/karma-coveralls#usage.

jgravois: > I think karma-coveralls wants it to be COVERALLS_REPO_TOKEN 

right you are. i've recycled the original token and removed the .yaml configuration file entirely.

Issue #101, opened 2018-01-22
Incorrect path for "browser" field in package.json


jluna:
### Problem
When building a bundle with `browserify`, I received the very unhelpful error message that the module could not be found even though it was installed in `./node_modules` via npm.

`Error: Cannot find module './node_modules/@esri/arcgis-rest-auth'`

### Solution
After much confusion, poking, and prodding. I realized that the "browser" path hint in `package.json` was pointing to the wrong location.  After fixing this locally browserify was able to find the correct distribution and build the bundle.

### Example
https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/package.json#L6
Presumably it's an issue for all packages.

jgravois:
thanks for the heads up @jluna! are you interested in submitting a Pull Request with the fix?

PR #102, opened 2018-01-22
ensure JS build tools can fetch a valid UMD file

jgravois:
happy to be voted down if others disagree, but i think the fact that we nest the file in a directory called `/umd` makes appending `.umd.` to the filename superfluous.

AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items
@esri/arcgis-rest-request

ISSUES CLOSED: #101

all tests passing with 100% coverage locally.


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: node_js
-sudo: false
+sudo: required
```
tomwayson: travis strikes again!

jgravois:
> I don't have strong opinions either

given that there are now two of us that don't feel strongly about it, and potentially someone else in a tree waiting to jump out at me, and travis mad for reasons i don't understand, lets just let sleeping dogs lie w/respect/to the filenaming convention. 😇

tomwayson:
Yes, apparently Travis has strong opinions on this.

.travis.yml
itrew: Added a pull request #103 to fix the chrome issue but still continue to use the container-based infrastructure.

noahmulfinger:
@jgravois FYI I'm seeing the same issue as travis locally when I run `npm run bootstrap`. So its not just a travis issue.

noahmulfinger:
I followed the [link in the build error](https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module), which led me to add namedExports to the rollup commonjs plugin:
```
commonjs({
  namedExports: {
    "../arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js": [
      "request",
      "ArcGISAuthError",
      "ArcGISRequestError",
      "getPortalUrl"
    ]
  }
}),
```

(Those 4 exports are the only ones used outside of `arcgis-rest-request`. )

With the above added, `npm run bootstrap` works successfully. However, I don't think this is a great solution since we would have to add to this array whenever another function is used externally.

itrew:
Another solution is a fix to the rollup.config. 

`resolve({ browser: true }),`
can be changed to
`resolve(),`

When the browser field was changed in the different packages it threw a wrench in things.

[Results](https://travis-ci.org/itrew/arcgis-rest-js/builds/332106024)

jgravois:
i think @itrew's suggestion makes the most sense to ensure rollup creates UMD's in the location that is publicized packages to downstream consumers, but thank you both for the help!

i'm gonna rebase this PR. would be good if @patrickarlt can confirm that he's okay with the update.

jgravois:
thx @patrickarlt 

i've rebased the branch. awaiting @tomwayson's confirmation that he's satisfied with the changes and that his friends won't jump out of a bush when we publish `v1.0.3` after this lands.

tomwayson:
If anyone jumps out of a tree/bush, send them to me, I'll deal w/ 'em.

PR #103, opened 2018-01-23
Fix Travis build. Continue to use containers.

itrew:
Updated the build config to use the --no-sandbox flag to fix the issue with chrome.

Just ran it [here](https://travis-ci.org/itrew/arcgis-rest-js).


itrew:
Primarily performance benefits as described [here](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure).

About a week ago Travis pushed an update that caused some headaches, many found [here](https://github.com/travis-ci/travis-ci/issues/8836). They have since updated their [docs](https://docs.travis-ci.com/user/chrome#Sandboxing).

jgravois:
awesome. thanks @itrew!

PR #109, opened 2018-01-31
Fix #108

### LINKED ISSUE #108: IE11 breaks w/o a fetch() polyfill even when supplying my own fetch() implementation
Based on #39, I would expect that if my app is running in some 💩 browser that doesn't support the frickin Internet, I could pass in an implementation of `fetch()` _without actually having to polyfill it_.

However the call to `.bind()` on this line requires a global `fetch()` to exist, regardless of whether or not you intend to use it:
https://github.com/Esri/arcgis-rest-js/blob/f12822cf57e8ab7e41867f17b243d676045588d5/packages/arcgis-rest-request/src/request.ts#L106

You're probably asking, "why not just conditionally polyfill `fetch()` for those unfortunate users?" [as I have suggested before](https://github.com/Esri/arcgis-rest-js/issues/34#issuecomment-326812183). Well, somehow, our app (the Hub), which has a metric ton of dependencies, has been supporting IE11 w/o a `fetch()` polyfill up until this point. That a) really surprises me, and b) makes me pause before adding them. I'll probably end up doing that in the short term, but wanted to raise this issue here and see if we wanted to allow consumers to use this library w/o a polyfill if they have a `fetch()` implementation

If so, do we want to do the same for for `Promise`?

patrickarlt:
@tomwayson this SHOULD prevent `request` from throwing an error when calling `request` with a custom `fetch` with no global `fetch` defined. I didn't test this in IE but I can get to it tomorrow if you can't do it today.

Fix #108

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -103,7 +103,7 @@ export function request(
   requestOptions: IRequestOptions = { params: { f: "json" } }
 ): Promise<any> {
   const options: IRequestOptions = {
-    ...{ httpMethod: "POST", fetch: fetch.bind(Function("return this")()) },
+    ...{ httpMethod: "POST", fetch },
     ...requestOptions
   };
 
@@ -133,6 +133,10 @@ export function request(
         ", "
       )} modules at the root of your application to add these to the global scope. See http://bit.ly/2BXbqzq for more info.`
     );
+  }
+
+  if (options.fetch === fetch) {
+    options.fetch = fetch.bind(Function("return this")());
   }
 
   const { httpMethod, authentication } = options;
--- a/packages/arcgis-rest-request/test/request.test.ts
+++ b/packages/arcgis-rest-request/test/request.test.ts
@@ -199,20 +199,52 @@ describe("request()", () => {
     beforeEach(() => {
       Promise = undefined;
       FormData = undefined;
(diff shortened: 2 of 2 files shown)


tomwayson:
LGTM on paper. I will test later today in IE when I can move back to my fork of the app that passes in ember's `fetch()`.

tomwayson:
@patrickarlt 

FYI - in the end I was not able to successfully use Ember's `fetch` after passing it in. It was just easier for me to solve the bug in our app by "polyfilling" the global `Promise`/`fetch` with Ember's implementations like:

```js
// get Ember's Promise/fetch
import { Promise } from 'rsvp';
import fetch from 'fetch';

export function initialize() {
  // "polyfill" global Promise/fetch for IE
  if (!window.Promise) {
    window.Promise = Promise;
  }
  if (!window.fetch) {
    window.fetch = fetch;
  }
}

export default {
  initialize
};
```

I don't feel great about that, but it works.

patrickarlt:
@tomwayson can't help you there, probably something a little odd about Embers `fetch` vs standards `fetch`. I wonder if the Angular `fetch` is the same.

tomwayson:
yep, we'll be looking into it further in one of the other ember libraries where it's a little less convoluted than my current use case (Hub -> ember-cli-cedar -> cedar -> arcgis-rest-request).

PR #111, opened 2018-02-01
throw ArcGISAuthError instead of Error when unable to refresh a token

### LINKED ISSUE #56: Throw authentication error when unable to refresh user's session
Currently when UserSession is unable to refresh the session, it just throws a generic error. It should throw an ArcGISAuthError.

https://github.com/ArcGIS/arcgis-rest-js/blob/7e5f2e5e862e7f2aaffcbc76a19ca37a44621c47/packages/arcgis-rest-auth/src/UserSession.ts#L513-L523

patrickarlt:
Fix #56


jgravois:
and it even had the correct number of @coveralls replies!

Issue #46, opened 2017-09-20
Design API for @esri/arcgis-rest-feature-layers module


patrickarlt:
API design for a module with support for creating, updating and deleting feature both feature layers and their features.

tomwayson:
finally working on this. - I'm thinking of going w/ `arcgis-rest-features` what say ye?

tomwayson:
I was thinking that would have been a nice terse name for a module that works w/ features (query, CRUD, etc). However, I wasn't thinking about the `getServiceInfo()`, `getLayerInfo()`, etc, which I think should be grouped w/ the query/CRUD fns, but seem a bit out of place under that name.

So I'm going to go back to something more along the lines of @noahmulfinger's proposal, but just make it: `arcgis-rest-feature-service` (drop the `s`).

tomwayson:
@noahmulfinger @jgravois 

After #126 the last couple functions I want to add before the initial release of a "read-only" `@esri/arcgis-rest-feature-service` are the functions to get the feature service info and (i.e. `/?f=json`) and the layer info (i.e. `/0?f=json`).

I think we should figure out the naming convention for these types of functions.

I've actually already written `getLayerInfo(url, requestOptions)`, and was planning to write `getServiceInfo(url, requestOptions)`, but then I noticed that @noahmulfinger suggested above to use the name `getService(url, requestOptions)`. Not to be confused w/ @jgravois' `serviceInfo(url, requestOptions)`, which I suggest we rename to include the term "geocode" somewhere (#122). Also I suspect there will be more of these types of functions as we support more types of services, so maybe we need to be more specific than `getService()` or `getServiceInfo()` (i.e. `getFeatureServiceInfo()`). Beyond that, it seems like the general naming convention questions are:
- include `Info` suffix?
- prefix w/ `get`?

Which brings me to my next question - do we even need these functions? I know in the case of @jgravois' `getService()` it defaults to the world geocoder if no URL is passed, so that definitely has a reason to exist. However, the only things my `getLayerInfo()` does beyond `request()` is: 
- default to `GET` (which implicitly depends on #127)
- defines a `ILayerInfo` for the response

I got about 30 lines into that `ILayerInfo` definition before I began to question my life choices:

```ts
export interface ILayerInfo {
  currentVersion: number;
  id: number;
  name: string;
  type: string;
  // TODO: how much of this is optional?
  serviceItemId?: string;
  displayField?: string;
  description?: string;
  copyrightText?: string;
[...]

PR #115, opened 2018-02-18
New arcgis-rest-features package w/ `getFeature()` to get a feature b…

tomwayson:
…y id

begins to address #46 

Started w/ `getFeature()` just to wire up the new package and run the names by you as I shortened what I proposed in https://github.com/Esri/arcgis-rest-js/issues/46#issuecomment-347365351 from `getFeatureById()` to `getFeature()` b/c that seemed more in line w/ `getItem()`.


tomwayson:
Oh, and @patrickarlt - I can't run the docs site, even on master, so this PR doesn't include any changes (if any) that are needed over there. It seems to hang on `docs:typedoc`:

```bash
➜  arcgis-rest-js git:(master) npm run docs:serve 

> @esri/arcgis-rest-js@1.0.2 predocs:serve /Users/tom/code/arcgis-rest-js
> npm run docs:typedoc


> @esri/arcgis-rest-js@1.0.2 docs:typedoc /Users/tom/code/arcgis-rest-js
> node docs/build-typedoc.js


Using TypeScript 2.4.1 from /Users/tom/code/arcgis-rest-js/node_modules/typedoc/node_modules/typescript/lib
```

tomwayson:
good point @jgravois - I was just working off the spec we laid out in #46 which predated the one arg to rule them all discussion in #78. I forgot about that. I'll update accordingly.

noahmulfinger:
@tomwayson What's the reasoning for naming the package `arcgis-rest-feature-service` without the *s*? It seems like `arcgis-rest-feature-services` is more consistent with `arcgis-rest-items` and `arcgis-rest-groups`

Also, I think `package-lock.json` should get added to this PR.

tomwayson:
Good question @noahmulfinger.

When I stared on this package, I thought a lot about the name. The first conclusion I came to was that I don't think we should avoid appending `-service` where possible, the exception being the package you are planning to actually administer services.

So, I was thinking that the best name for this package, and the best way to be consistent w/ `arcgis-rest-items` and `arcgis-rest-groups` would be `arcgis-rest-features`, but [I worried that `getLayerInfo` or `getServiceInfo` would not fit well  under that name](https://github.com/Esri/arcgis-rest-js/issues/46#issuecomment-366524883). I'm glad to go back to `arcgis-rest-features` if y'all think that's not a problem.

Once I made the name about the _service_ rather than the _features_, the only other peer out of the other pacakges (-auth, -common-types, -geocoder, -request) was really `arcgis-rest-geocoder`, which is kinda it's own thing. So no real help there.

Also, back in #46 you and I both flip flopped back and forth btw/ `arcgis-rest-feature-service` and `arcgis-rest-feature-services` - you originally proposed the singular, then we settled on the latter. However, now that we have -items and -groups, I thought the singular would best.

noahmulfinger:
@tomwayson Okay, thinking about it more, `arcgis-rest-feature-service` sounds better to me as well. I kinda think going singular with *all* the packages would be better, but we can think about in the future.

tomwayson:
Thanks @noahmulfinger. I'm still torn, I like `arcgis-rest-features`. I'm still worried that we're going to end up w/ `-service` on a lot of package names.

A way we could standardize on non-pulral names would have been: 
`arcgis-rest-items` -> `arcgis-rest-content`
`arcgis-rest-groups` -> `arcgis-rest-sharing`

noahmulfinger:
@tomwayson Yeah, I see what you mean. However, I don't think we will need to append `-service` to any of the services that represent specific actions or tasks, for instance `arcgis-rest-geoenrichment`, `arcgis-rest-spatial-analysis`, `arcgis-rest-geometry`.

And, at least in this case, I think the full term `feature-service` is necessary to have getServiceInfo() and getLayerInfo(), as you mentioned in the issue.

tomwayson:
Good point.

As per a discussion w/ @jgravois we're going to hold on merging this until he can cut a release w/ the fixes since 1.0.2 since this will require a minor bump.

Also, FYI I've got `getLayerInfo` and `queryFeatures` queued up locally and will push PRs for those once this gets merged.

Issue #107, opened 2018-01-29
FYI: existing type definitions


JeffJacobson:
A while ago I contributed the [@types/arcgis-rest-api type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/arcgis-rest-api) to the DefinitelyTyped repository. If they meet your needs you could use these instead of creating them from scratch. (Available via NPM [here](https://www.npmjs.com/package/@types/arcgis-rest-api))

I have another version with additional types, but that hasn't been contributed yet. I've attached that version below, in case you want to use it.
[arcgis-rest-api.d.ts.zip](https://github.com/Esri/arcgis-rest-js/files/1674776/arcgis-rest-api.d.ts.zip)

(The only reason that all of the types are in a single `index.d.ts` file is simply because that's the way the DefinitelyTyped project expects them to be.)

jgravois:
very cool! 

this is really similar to the [`arcgis-rest-common-types`](https://github.com/Esri/arcgis-rest-js/blob/f12822cf57e8ab7e41867f17b243d676045588d5/packages/arcgis-rest-common-types/src/index.ts) package we started stubbing out.

we could certainly link out and promote your additional typings, but my first instinct is that it'd be good to combine forces. would you be interested in that?

JeffJacobson:
Yeah, that sounds good. Seems like it would be best to just copy the types I made into your project rather than linking out to the `@types` ones, so they would be kept all in a single location.

tomwayson:
@JeffJacobson this is awesome - I'm starting to recreate some of these - and I don't want to.

No matter what we decide, I think the main thing is to eliminate any duplication of error.

I'll be the first to admit that I'm mystified by TS best practices, I thought DefinitelyTyped was out, and `npm i @types/whateves` was in. Now it appears you're telling me _they are the same thing_. 

 -->Mind blown<--

Also, we've prefixing all interfaces w/ `I`, but you haven't done that. What gives bro? I like yours better. Is that cool? Can we do that?

[...]

Issue #104, opened 2018-01-24
Inconsistent signatures...


dbouwman:
`searchItems` and `searchGroups` signatures are inconsistent;

`searchItems(search: string | ISearchRequestOptions) : Promise<ISearchResult>` has one arg...
and `ISearchRequestOptions` contains a `searchForm`...

where as `searchGroups` takes the `searchForm` as the first arg.

`searchGroups(searchForm: IGroupSearchRequest, requestOptions: IRequestOptions)`

Either is fine, but consistency makes the lib much easier to work with.

As we port the ember-arcgis-portal-services over to using this lib, I'm sure we will find other things like this :)

jgravois:
assigning to myself. i'd like to reorganize the way actual search parameters are supplied entirely.

jgravois:
tracking in #137

PR #120, opened 2018-02-22
fix(items): dont override user supplied parameters when updating items

jgravois:
AFFECTS PACKAGES:
@esri/arcgis-rest-items

ISSUES CLOSED: #117 

to be honest, mixing in a properly serialized item like this seems a little _strange_.

```ts
requestOptions.params = {
    ...requestOptions.params,
    ...serializeItem(requestOptions.item)
  };
```

PR #121, opened 2018-02-23
add more common types

jgravois:
porting a whole load of interfaces and types originally written by @JeffJacobson :tada: 

AFFECTS PACKAGES:
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items
batch-geocoder

BREAKING CHANGE:
no longer prefacing interface names with an `I`

maybe its just the ArcObjects phobia talking, but 5 minutes into this port i started to agree with [@tomwayson](https://github.com/Esri/arcgis-rest-js/issues/107#issuecomment-367217416). A quick look at the TypeScript Language [Specification](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#7.1) made it clear that its not a universally agreed upon convention.

i'm planning on tackling #104 soon so it makes sense to me to rip off the proverbial esri-loader band-aid here too and just release a `2.0.0` if everyone else agrees.

for now, the associated linter rule is turned [`off`](https://palantir.github.io/tslint/rules/interface-name/) because we're still using the `I` outside common-types.

### DIFF
--- a/packages/arcgis-rest-groups/src/groups.ts
+++ b/packages/arcgis-rest-groups/src/groups.ts
@@ -7,10 +7,10 @@ import {
   getPortalUrl
 } from "@esri/arcgis-rest-request";
 
-import { PagingParams, Item } from "@esri/arcgis-rest-common-types";
+import { IPagingParams, IItem } from "@esri/arcgis-rest-common-types";
 
 export interface IPagingParamsRequestOptions extends IRequestOptions {
-  paging: PagingParams;
+  paging: IPagingParams;
 }
 
 export interface IGroupIdRequestOptions extends IRequestOptions {
@@ -32,7 +32,7 @@ export interface IGroupRequestOptions extends IRequestOptions {
   group: IGroup;
 }
 
-export interface IGroupSearchRequest extends PagingParams {
+export interface IGroupSearchRequest extends IPagingParams {
   q: string;
   sortField?: string;
   sortOrder?: string;
@@ -56,7 +56,7 @@ export interface IGroupContentResult {
   start: number;
   num: number;
   nextStart: number;
-  items: Item[];
+  items: IItem[];
 }
 
 export interface IGroupUsersResult {
--- a/packages/arcgis-rest-groups/src/groups.ts
+++ b/packages/arcgis-rest-groups/src/groups.ts
@@ -3,7 +3,7 @@
 import {
   request,
   IRequestOptions,
-  IParams,
+  // IParams,
   getPortalUrl
 } from "@esri/arcgis-rest-request";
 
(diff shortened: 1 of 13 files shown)


patrickarlt:
@jgravois @tomwayson while prefixing interfaces with `I` isn't a universal standard by any means (it is borrowed from .Net) but think it makes sense to differentiate between classes in the JS API `Extent`, `Point`, `SpatialReference` and `Item` and the interfaces that we use in these libraries, `IExtent`, `IPoint`, `ISpatialReference` and `IItem`.

I would hate to end up with something like this:


```ts
import { Extent } from "@esri/rest-js-common-types";

require("esri/Extent", (Extent) => {
  // now you have 2 problems...
});
```

Prefixing with `I` quickly tells you that the thing you are looking at is an interface and not a class. Just because the [TypeScript Guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) discourage it is no reason not to do since the guidelines are for _contributing to TypeScript_ not writing TypeScript code in general.

packages/arcgis-rest-groups/src/groups.ts
```diff
@@ -230,9 +231,9 @@ export function updateGroup(
 export function removeGroup(
   requestOptions: IGroupIdRequestOptions
 ): Promise<any> {
-  const url = `${getPortalUrl(
-    requestOptions
-  )}/community/groups/${requestOptions.id}/delete`;
+  const url = `${getPortalUrl(requestOptions)}/community/groups/${
+    requestOptions.id
```
tomwayson: TIL - you can have line breaks w/in `${}` in template literals w/o adding line breaks to the output.

tomwayson:
If we don't rename the existing interfaces, then this would not have to be a breaking change, right?

If not, I say we lump these in w/ either a 1.0.3 release, or a 1.1.0 release w/ #115

tslint.json
```diff
@@ -8,6 +8,7 @@
     "strict-type-predicates": false,
     "ordered-imports": ["any"],
     "only-arrow-functions": [false],
-    "object-literal-sort-keys": false
+    "object-literal-sort-keys": false,
+    "interface-name": [false, "always-prefix"]
```
tomwayson: I think we want to remove this now that we're back to `I`ing everything, right?

jgravois: definitely. fixed.

tomwayson:
This LGTM now, so merge and include in 1.0.3, or wait for 1.1.0?

jgravois:
> merge and include in 1.0.3, or wait for 1.1.0?

I'm fine with either. i just want to make sure and include #120 in `v1.0.3`

tomwayson:
If 1.0.3 is eminent, I'd say let this be the first PR of 1.1.0

PR #129, opened 2018-02-27
Set externals properly for UMD builds

### LINKED ISSUE #128: peerDependencies should treat arcgis-rest-request as external
As far as I understand, we're implementing packages like -geocoder, -items, -groups, -feature-service, etc as plugins of arcgis-rest-request using `peerDependencies`, right? However, the UMD builds currently do **not** treat arcgis-rest-request as external, so it is bundled in the UMD build of each package. I discovered this while creating the demo for arcgis-rest-feature-service:

https://github.com/Esri/arcgis-rest-js/blob/c961602ac002264e55da5d76514bd2ad64f247ac/demos/feature-service-browser/index.html#L40-L41

I think we need to [modify the rollup config to treat arcgis-rest-request as `external`](https://rollupjs.org/guide/en#peer-dependencies).

patrickarlt:
Close #128.

@tomwayson @jgravois I did verify that this makes bundles smaller but you should verify that the demos still work. I probably inadvertently removed this while setting up the build system.

### DIFF
--- a/demos/ago-node-cli/README.md
+++ b/demos/ago-node-cli/README.md
@@ -13,7 +13,7 @@ Like all the other demo apps, run `npm run bootstrap` from the root.
 ### Running
 If you use this demo as a starting point for your own command line package, you would publish it to npm, then on the target systems run `npm install <your-cli-package>`, and it would be available as a command.
 
-But, this is demo code, and thus the package is not "installed" via `npm install ...`, before we can call it as `ago <command> <query>` we need to run `npm link` in the `/demos/node-cli` folder. After you do that, the command should work.
+But, this is demo code, and thus the package is not "installed" via `npm install ...`, before we can call it as `ago <command> <query>` we need to run `npm link` in the `/demos/ago-node-cli` folder. After you do that, the command should work.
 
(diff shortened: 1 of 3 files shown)


tomwayson:
Demos that I've tested so far:
- [x] gecode-browser
- [x] ago-node-cli

I think if we test one of the auth ones we can merge. I have to run, but will try that later.

jgravois:
> I think if we test one of the auth ones we can merge.

our only _real_ auth demo uses Express (and consequently doesn't load UMDs). that said, my own browser geocoding demo has its own OAuth implementation for batch geocoding. i confirmed locally that (a) the UMDs did indeed shrink and (b) didn't bust anything auth related.

all said, nice catch @tomwayson and _superb_ fix @patrickarlt. i was stuck on the idea that we'd need to omit the package we actually wanted to bundle from the `externals: []`.

thinking about it more, it makes sense that you can do that because internally a package wouldn't include an import that referenced its own name.

tomwayson:
duh - forgot this repo did it's own CJS builds. so glad I tested ago-node-cli! thanks @jgravois

PR #131, opened 2018-02-28
check for errors before trying to parse response

### LINKED ISSUE #110: check for error _before_ parsing response as JSON/text/whateves
It seems to me that we need a `response.ok` check before this `switch`:

https://github.com/Esri/arcgis-rest-js/blob/f12822cf57e8ab7e41867f17b243d676045588d5/packages/arcgis-rest-request/src/request.ts#L175-L190

The same way ember-arcgis-portal-services does here:

https://github.com/Esri/ember-arcgis-portal-services/blob/master/addon/mixins/service-mixin.js#L90-L108

What happens now is that the underlying HTTP error (404, 500, etc) gets buried by a JSON parsing error like `SyntaxError: Unexpected token < in JSON at position 0`, at least in the case of `f=json`.

See the first error on https://esri.github.io/ember-cli-cedar/#/error-handling as an example.

If y'all agree, I can PR the logic from ember-arcgis-portal-services.

tomwayson:
resolves #110

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -193,6 +193,11 @@ export function request(
       return options.fetch(url, fetchOptions);
     })
     .then(response => {
+      if (!response.ok) {
+        // server responded w/ an actual error (404, 500, etc)
+        const { status, statusText } = response;
+        return Promise.reject(new Error(`${status}: ${statusText}`));
+      }
       switch (params.f) {
         case "json":
           return response.json();
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -5,6 +5,7 @@ import { checkForErrors } from "./utils/check-for-errors";
 import { encodeFormData } from "./utils/encode-form-data";
 import { encodeQueryString } from "./utils/encode-query-string";
 import { requiresFormData } from "./utils/process-params";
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -193,6 +193,11 @@ export function request(
       return options.fetch(url, fetchOptions);
     })
     .then(response => {
+      if (!response.ok) {
+        // server responded w/ an actual error (404, 500, etc)
+        const { status, statusText } = response;
+        return Promise.reject(new Error(`${status}: ${statusText}`));
```
patrickarlt: @tomwayson should this be an `ArcGISRequestError`? Mainly so users can avoid situations like having to check for `Error` vs `ArcGISRequestError`:

```js
request(/* ... */).catch(error => {
  if(error instanceof Error) {
    // ok I have an error so I only have error.message...
  }

  if(error instanceof ArcGISRequestError) {
    // ok I have an ArcGISRequestError so I can access lots of other stuff...
  }
})
```

It might be easier if users can do this:

```js
request(/* ... */).catch(error => {
  // error is always a ArcGISRequestError
})
```

This could be something like: 

```ts
return Promise.reject(new ArcGISRequestError(statusText, `HTTP ${status}`));
```

which woudl result in error messages like `HTTP 404: Not Found`

tomwayson: I'll give that a go later today

Issue #132, opened 2018-02-28
Remove tsconfig.json from published npm packages


richardhinkamp:
Hi I installed several packages through npm and they all contain a `tsconfig.json` which contains `"extends": "../../tsconfig.json"`. This points to `node_modules/tsconfig.json` which does not exist. I guess the `tsconfig.json` should not be included in the npm package?

PR #133, opened 2018-02-28
chore(npm packages): remove test files and tsconfig.json from npm tar…

jgravois:
…balls

since we are distributing built types, UMDs, ES6 modules (with types) and common JS modules for node, it seems safe to remove the tsconfig.json entirely, especially since its incomplete.

there's no need to specify "README.md" or "package.json" or whatever is listed as "main". they are [_always_](https://docs.npmjs.com/files/package.json) included. this change drops the tests too to further shrink the bundles.

this will also solve problems like unintentionally shipping cruft like this:
https://unpkg.com/@esri/arcgis-rest-request@1.0.3/.rpt2_cache/

AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items
@esri/arcgis-rest-request

ISSUES CLOSED: #132

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -24,3 +24,5 @@ Using TypeScript allows us to add type information to request params and respons
 TypeScript compiles to JavaScript so you can use @esri/arcgis-rest-js in any JavaScript project. However if you use TypeScript you will get the benefits of type checking for free.
 
 We also _really_ like TypeScript because it supports exporting to both [ES 2015 modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) (use `import`/`export`) and [UMD](http://davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/) modules. This allows us to support a wide variety of module loaders and bundlers, including Browserify, Webpack, RequireJS, and Dojo 1 and 2.
+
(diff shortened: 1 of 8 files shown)


jgravois:
@JeffJacobson (or any other TypeScript developer), is it sufficient when a lib you depend on only exposes ES6 modules and types and doesn't include raw `.ts` files?

https://unpkg.com/@esri/arcgis-rest-request@1.0.3/src/
https://unpkg.com/@esri/arcgis-rest-request@1.0.3/dist/esm/

tomwayson:
I am far from a "TypeScript developer", but, FWIW, I _have been_ including them in other TS projects [like esri-loader](https://github.com/Esri/esri-loader/blob/e647ace333498fc375c56daf7dc1af4f94d6b950/package.json#L7), but I think that's a mistake.

I'm almost certain that they are not needed because `main`, `module`, `types`, `browser` etc all point to files under `dist`, the only way a consuming TS app could access the src files would be `import { request } from 'node_modules/@esri/arcgis-rest-request/src/request.ts`, and even then unless they were using the same tsconfig as us, they could have compile errors.

I say we drop `src` too. If we do, I'll plan to do the same in esri-loader and my other TS packages.

patrickarlt:
@jgravois TypeScript uses the files under `dist` like @tomwayson said. This is why we publish all the `.d.ts` files in the `esm` folder.

Issue #122, opened 2018-02-23
rename geocoder's serviceInfo()


tomwayson:
`serviceInfo()` is too generic a name for something that's going to default to the world geocoder url:

https://github.com/Esri/arcgis-rest-js/blob/2c05a283eac89a94732394ffc3aca8ddcbc18065/packages/arcgis-rest-geocoder/src/geocoder.ts#L387

I suggest `getGeocoderService()` or `getGeocoderServiceInfo()`.

We can rename it and stub in a new fn w/ the old name that shows a deprecation warning until the next breaking change.

PR #134, opened 2018-03-01
refactor(geocoder): use a more descriptive method to fetch metadata

jgravois:
and set up previous method to be deprecated in next major release

AFFECTS PACKAGES:
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-request

ISSUES CLOSED: #122

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -108,7 +108,8 @@ export function request(
   requestOptions: IRequestOptions = { params: { f: "json" } }
 ): Promise<any> {
   const options: IRequestOptions = {
-    ...{ httpMethod: "POST", fetch },
+    httpMethod: "POST",
+    fetch,
     ...requestOptions
   };
 
--- a/packages/arcgis-rest-request/src/utils/check-for-errors.ts
+++ b/packages/arcgis-rest-request/src/utils/check-for-errors.ts
@@ -54,3 +54,12 @@ export function checkForErrors(
 
   return response;
 }
+
+/**
+ * Method used internally to surface messages to developers.
+ */
+export function warn(message: string) {
+  if (console && console.warn) {
+    console.warn.apply(console, [message]);
+  }
+}
--- a/packages/arcgis-rest-request/test/utils/check-for-errors.test.ts
+++ b/packages/arcgis-rest-request/test/utils/check-for-errors.test.ts
@@ -92,8 +92,10 @@ describe("warn", () => {
 
 describe("warn", () => {
   it("should carry on gracefully when no console is available", () => {
+    const realConsoleWarn = console.warn;
     console.warn = undefined;
     warn("Danger Will Robinson!");
     expect(console.warn).toBe(undefined);
+    console.warn = realConsoleWarn;
   });
 });
--- a/packages/arcgis-rest-request/test/utils/check-for-errors.test.ts
+++ b/packages/arcgis-rest-request/test/utils/check-for-errors.test.ts
@@ -1,5 +1,6 @@
 import {
   checkForErrors,
+  warn,
   ArcGISRequestError,
   ArcGISAuthError
 } from "../../src/index";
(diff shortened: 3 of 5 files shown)


packages/arcgis-rest-request/src/utils/check-for-errors.ts
```diff
@@ -54,3 +54,12 @@ export function checkForErrors(
 
   return response;
 }
+
+/**
+ * Method used internally to surface messages to developers.
+ */
+export function warn(message: string) {
+  if (console && console.warn) {
+    console.warn.apply(console, [message]);
```
tomwayson: just curious, why `apply()` here instead of just calling `console.warn(messge)`?

packages/arcgis-rest-request/test/utils/check-for-errors.test.ts
```diff
@@ -80,3 +81,19 @@ describe("checkForErrors", () => {
...
+  it("should bubble up deprecation warnings", () => {
+    console.warn = jasmine.createSpy("warning");
+    warn("Danger Will Robinson!");
+    expect(console.warn).toHaveBeenCalledWith("Danger Will Robinson!");
+  });
+});
+
+describe("warn", () => {
+  it("should carry on gracefully when no console is available", () => {
+    console.warn = undefined;
```
tomwayson: shouldn't we restore this after the test is run?

tomwayson:
Also, in https://github.com/Esri/arcgis-rest-js/issues/46#issuecomment-368326432 I finally meandered around to suggesting we drop the `Info` from the fn name and the response type. So if we wanted to make the name consitent here, it would be:

`getGeocoderService(requestOptions): Promise<IGeocodeService>`

But I could be convinced to go the other way and append `Info` to the fns and response types over there, so:

```ts
getFeatureServiceInfo(url, requestOptions): IFeatureServiceInfo;
getFeatureLayerInfo(url, requestOptions): IFeatureLayerInfo;
```

jgravois:
> as long as we're changing names, let's try to do it only once.

i _totally_ agree (and also forgot initially to refactor the code to stop POSTing at all times.

my vote for a method name is just `geocodeService()` because folks can.

```ts
geocodeService({ httpMethod: "POST" })
```

packages/arcgis-rest-request/src/utils/check-for-errors.ts
jgravois: i assure you my level of understanding here doesn't exceed yours. this is just a blind TypeScript port from [esri-leaflet](https://github.com/Esri/esri-leaflet/blob/82860cab3696fb5889523d8bb3dee49b14006ee6/src/Util.js#L167-L171) that sidesteps a type and linter error
```js
export function warn () {
  if (console && console.warn) {
    console.warn.apply(console, arguments);
  }
}
```
* [`tslint` Rule: no-console bans the use of specified console methods.](https://palantir.github.io/tslint/rules/no-console/)
* [`TypeError:` CreateListFromArrayLike called on non-object](https://stackoverflow.com/questions/41354099/getting-error-createlistfromarraylike-called-on-non-object-when-trying-to-use-a)

packages/arcgis-rest-request/src/request.ts
```diff
@@ -108,7 +108,8 @@ export function request(
   requestOptions: IRequestOptions = { params: { f: "json" } }
 ): Promise<any> {
   const options: IRequestOptions = {
-    ...{ httpMethod: "POST", fetch },
+    httpMethod: "POST",
+    fetch,
     ...requestOptions
```
jgravois: totally unrelated tweak to avoid some unnecessary destructuring in `request()` itself.

tomwayson:
I still lean _slightly_ toward `getGeocodeService()` even if people can override w/ POST. Mainly b/c in -feature-service `getFeatuerLayer()` and `getFeatureService()` sound more consistent w/ `getFeature()`, but also b/c `geocode` is a verb and it sounds a bit like you're going to geocode a service.

No hair is too small to split! Especially the week before DevSummit.

I've said my final piece on the matter. You just do you and I'll follow suit.

jgravois:
you win. 

pushed another commit to use `getGeocodeService()` and `IGetGeocodeServiceResponse`.

you also win an all expense paid trip to take the first crack at writing a style guide. 😉

Issue #100, opened 2018-01-12
refactor demos/vanilla sample to demonstrate popup:false as well


jgravois:
i did a little bit of head scratching this afternoon and i still can't figure out exactly how we want browser developers to implement OAuth2 when they _don't_ want to use a popup.

it'd be cool to tweak our [existing sample](https://github.com/Esri/arcgis-rest-js/tree/94edc2a1d3110d448fd7abe8367812deccbc7647/demos/vanilla) so that it can be toggled to demo both or just write a new one.

```js
// no popup
var session = arcgisRest.UserSession.completeOAuth2({
    clientId: ClientId,
    redirectUri: "http://localhost:8080/post-sign-in.html",
    popup: false,
    portal: "https://www.arcgis.com/sharing/rest"
  })

session // how can i pass a session like this from the redirect page back to the original app?

Issue #11, opened 2017-08-06
Example Apps


dbouwman:
In order to ensure that the api works for our desired use-cases, I propose we build some simple applications to validate the library.

The following demos seem useful:
- jquery app
- dojo app
- ember js app
- angular js app
- react app
- node cli app

For the browser demos, the app should use oAuth to get a token, then instantiate a session and execute a simple, authenticated search @ AGO. 

The node example should take the creds as command-line args, and iterate over all the items the user owns, listing the titles.

At this point, lets keep these as vanilla as possible - just bootstrap + minimal code needed. We can build idiomatic examples down the road.

I'll sign-up to create the `arcgis-rest-js-demos` repo, and build the jquery, ember and node examples. Need some Dojo/Angular/React people to jump in as well @tomwayson @mjuniper??

jgravois:
more `demos/` are **always** welcome.

PR #138, opened 2018-03-01
Example using VueJS

itrew:
Wanted to get my feet wet with this lib and thought it would be helpful to contribute in the process. This is a VueJS app that uses the authentication and request packages. 

Initial template created with [webpack-simple](https://github.com/vuejs-templates/webpack-simple) then added on the vue-router plugin to handle the authentication callbacks.

image


jgravois:
rock and roll 🎼! it looks like this even resolves #100.

itrew:
Aside from #11, #100 was in my mind when I was creating this. This type of implementation would require some form of routing framework to work, though. 

I'm not entirely sure how to pass a session through a redirect like you mentioned. Perhaps a localStorage implementation? I can mess around with that a bit and see what I find.

patrickarlt:
@itrew do you want to update this with the localstorage solution from https://github.com/Esri/arcgis-rest-js/pull/140 and then I can review it?

itrew:
@patrickarlt Thanks for the feedback, everything makes sense 👍 I'll spend some time in the next couple of days updating #140 with both your's and @jgravois 's input on this.

jgravois:
i agree. thanks for both your contributions and for commenting your code so thoroughly!

PR #140, opened 2018-03-01
Updated the vanilla JS demo.

### LINKED ISSUE #100: refactor demos/vanilla sample to demonstrate popup:false as well

itrew:
Per discussion https://github.com/Esri/arcgis-rest-js/pull/138#issuecomment-369724689 this is an updated version of the vanilla js demo. Addresses #11 and should close #100. This implementation stores the session in localStorage and de-serializes it into a new session when not using a popup.

The general gist of OAuth is included, but with this being vanilla I wasn't going to go through the trouble of adding the same search functionality that I had added to the VueJS demo.

image


itrew:
Realized when the session was being placed into local storage the expiration date was being stringified. Last commit now creates a new date from that string when parsing the serialized session on page load.

demos/oauth2-browser/authenticate.html
```diff
@@ -0,0 +1,28 @@
...
+    <script src="node_modules/@esri/arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js"></script>
+    <script src="node_modules/@esri/arcgis-rest-auth/dist/umd/arcgis-rest-auth.umd.js"></script>
+    <script>
+      const match = window.location.href.match(
+        /clientID=(.+)#/
+      );
+      const clientId = match[1];
+      let session;
+      function processAuthentication() {
+        session = arcgisRest.UserSession.completeOAuth2({
```
patrickarlt: This should probably come AFTER the local storage and window calls since it will call `window.close()` in somecases.

jgravois:
pushed another commit to incorporate @patrickarlt's (good) suggestions.

demos/oauth2-browser/authenticate.html
jgravois: i moved the window call, but we can't call `session.serialize()` if the object isn't instantiated yet.

Issue #136, opened 2018-03-01
upgrade rollup to v0.56.3


jgravois:
ref: https://github.com/Esri/esri-leaflet/pull/1054/commits/76236f440d26f0e078a7e69d706eeabb1012b8cb

Issue #135, opened 2018-03-01
distribute unminified files


tomwayson:
anyone object? I think pulling in the minified files in ember addons is causing slow builds. we're still working on configuring ember to not try and minify the pre-uglfied files, but I wondered if anyone had a reason to **not** distribute them?

cc @patrickarlt @dbouwman @jgravois

jgravois:
fine by me.

patrickarlt:
@tomwayson the ESM and Node files aren't minified anyway only the UMD files. Is there any way to try to pull the compiled ESM modules into Ember? The only downside I can see to not minifying the UMD modules is the filesizes on the CDN would be larger.

jgravois:
@tomwayson you were suggesting shipping _both_ right?

https://unpkg.com/leaflet@1.3.1/dist/leaflet-src.js
https://unpkg.com/leaflet@1.3.1/dist/leaflet.js

tomwayson:
So, I looked at [specifying `output` as an array](https://rollupjs.org/guide/en#configuration-files) to generate multiple files from the same config, but it turns out we're quite a few versions behind of rollup, and we don't even use the `input`/`output` options. I don't think it's a good idea to mess w/ that now. Of course we could do what I do in cedar, which I think I stole from leaflet, which is just run rollup 2x for prod builds. All of that is sounding kinda hairball right now w/ all the devsummit stuff looming.

@dbouwman how bad is this needed? You still having long build times?

jgravois:
> the downside would be that the dist is larger, so slower npm installs.

ember needs or not, i think we should be shipping an un-minified UMD too to make downstream debugging easier. even with a sourcemap, variable names don't resolve when loading the obfuscated build.

![screenshot 2018-03-01 10 14 55](https://user-images.githubusercontent.com/3011734/36861667-67b97b5a-1d39-11e8-8c95-b8e453034bfc.png)

dbouwman:
@tomwayson this is not critical path at all - I was really just using this time to poke at this before we all convene in Palm Springs, so if I hit any ZOMG show-stoppers we can puzzle it out face-to-face.

That said - right now the build time on the `torii-provider-arcgis` dummy app is > 5 minutes, and it just has the `request` and `auth` modules included.

I'll also echo @jgravois that debugging against the minified stuff is 😱 - imho - we def need to ship un-minified umd's

dbouwman:
In contrast - before adding those modules, the build is 16s

[...]

PR #141, opened 2018-03-02
chore(rollup): bump to the latest version of rollup

jgravois:
to support multiple output files (and make sure to ship our copyright notice)

i _thought_ i was laying the groundwork for #135, but https://github.com/TrySound/rollup-plugin-uglify/issues/1 indicates that we'll need to spin up two different instances of rollup to create both built and unbuilt output. 

ISSUES CLOSED: #136


jgravois:
pushed up another commit that kicks off two rollup commands for each package so that we can ship a second unminified UMD.

tomwayson:
I will make that change and test in cedar now.

tomwayson:
I was going to point `browser` to the unminified file as per @patrickarlt's suggestion, but it seemed odd to have that pointing to a file w/ "debug" in the name. I think it's more common to specify the minified files w/ .min.js - take [bootstrap for example](https://unpkg.com/bootstrap@4.0.0/dist/js/), so I updated the profiles accordingly. This meant that I didn't need to update the `browser` field of all the package.json files.

I tested this w/ both the features sservice demo app, and a branch of cedar configured to treat arcgis-rest-* as external, and it works w/ both the unminified and minified files.

PR #146, opened 2018-03-03
Fix incorrect types in arcgis-rest-feture-service and get branch coverage back to 100%

tomwayson:
Looks like I noticed `IFeatureSet` in common types before, but forgot to use it.


tomwayson:
@jgravois 

If this LGTY we can release 1.1.0 after it's merged AFAIC. LMK if you want me to try the release.

PR #163, opened 2018-03-27
scrollbar for long words in <tables> instead of overflow

jgravois:
the only downside i can see to the change i'm proposing is that it collapses other tables when they don't have enough content.

screenshot 2018-03-27 16 30 37

### DIFF
--- a/docs/src/_layout.html
+++ b/docs/src/_layout.html
@@ -71,7 +71,7 @@
       </div>
     </footer>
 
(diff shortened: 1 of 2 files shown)


noahmulfinger:
@jgravois Unfortunately, another issue with this fix is that the anchor links get partially hidden:
screen shot 2018-04-05 at 1 55 07 pm

PR #169, opened 2018-04-04
fix(OAuth2 options): add 'locale' and 'state' parameters for browser based OAuth2

jgravois:
In porting [`torii-provider-arcgis`](https://github.com/dbouwman/torii-provider-arcgis/) to use this lib i realized that we are missing support for a couple useful parameters exposed by [oauth2/authorize?](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)

* locale - The locale assumed to render the login page.
* state - An opaque value used by applications to maintain state between authorization requests and responses.

AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-demo-vanilla

### DIFF
--- a/demos/oauth2-browser/authenticate.html
+++ b/demos/oauth2-browser/authenticate.html
@@ -10,13 +10,14 @@
     <script src="node_modules/@esri/arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js"></script>
     <script src="node_modules/@esri/arcgis-rest-auth/dist/umd/arcgis-rest-auth.umd.js"></script>
     <script>
-      // in a production app, clientID would be hardcoded. we're uusing a regex so that developers can pass one in at runtime.
+      // in a production app, clientID would be hardcoded. we're using a regex so that developers can pass one in at runtime.
+      debugger;
       const match = window.location.href.match(
-        /clientID=(.+)#/
+        /&state=(.+)/
       );
       const clientId = match[1];
       let session;
-      function processAuthentication() {        
+      function processAuthentication() {
         window.location.href = '/';
         session = arcgisRest.UserSession.completeOAuth2({
           clientId,
--- a/demos/oauth2-browser/authenticate.html
+++ b/demos/oauth2-browser/authenticate.html
@@ -10,8 +10,10 @@
     <script src="node_modules/@esri/arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js"></script>
     <script src="node_modules/@esri/arcgis-rest-auth/dist/umd/arcgis-rest-auth.umd.js"></script>
     <script>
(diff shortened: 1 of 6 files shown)


demos/oauth2-browser/index.html
```diff
@@ -162,6 +162,8 @@ <h2>
             clientId,
             redirectUri: `${redirect_uri}authenticate.html?clientID=${clientId}`,
             popup: true,
+            locale: `en`,
+            state: "abc123"
```
noahmulfinger: @jgravois Just a thought, what if we set the `state` as the clientId here? That way we wouldn't need to add `?clientID=${clientId}` to the end of the `redirectUri`

demos/oauth2-browser/authenticate.html
```diff
@@ -10,13 +10,14 @@
     <script src="node_modules/@esri/arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js"></script>
     <script src="node_modules/@esri/arcgis-rest-auth/dist/umd/arcgis-rest-auth.umd.js"></script>
     <script>
-      // in a production app, clientID would be hardcoded. we're uusing a regex so that developers can pass one in at runtime.
+      // in a production app, clientID would be hardcoded. we're using a regex so that developers can pass one in at runtime.
+      debugger;
       const match = window.location.href.match(
-        /clientID=(.+)#/
+        /&state=(.+)/
```
tomwayson: maybe I'm missing something, but if the consumer overrides `state` (which defaults to `clientID`) then this won't actually be the `clientId`, right?

demos/oauth2-browser/authenticate.html
```diff
@@ -10,13 +10,14 @@
     <script src="node_modules/@esri/arcgis-rest-request/dist/umd/arcgis-rest-request.umd.js"></script>
     <script src="node_modules/@esri/arcgis-rest-auth/dist/umd/arcgis-rest-auth.umd.js"></script>
     <script>
-      // in a production app, clientID would be hardcoded. we're uusing a regex so that developers can pass one in at runtime.
+      // in a production app, clientID would be hardcoded. we're using a regex so that developers can pass one in at runtime.
+      debugger;
```
tomwayson: we want to remove this `debugger`, right?

jgravois: i could have _sworn_ i caught that on friday and pushed up a commit with a fix.

demos/oauth2-browser/authenticate.html
jgravois: > if the consumer overrides state (which defaults to clientID) then this won't actually be the clientId.

thats true, i've fleshed out the code comment a little more to explain this in more detail. let me know what you think.

jgravois:
still waiting on a green 🚥 from _someone_...

Issue #159, opened 2018-03-16
IUser


jpeterson:
It would be useful to have a method/primitive that expresses an **ArcGIS User**. This feels similar to the `IPortal` and `getSelf()` method in `arcgis-rest-request`.

Note that an `IPortal` returned by `request.getSelf()` includes a `user` property, but this object doesn't appear to be a fully-qualified **User** per the REST API spec.

**User spec**
https://developers.arcgis.com/rest/users-groups-and-items/user.htm

The following works fine today:
```javascript
import { request } from '@esri/arcgis-rest-request';

// session === UserSession();

// use request module to fetch user profile
request(`${session.portal}/community/users/${session.username}`, {
  authentication: session
})
```

tomwayson:
Yeah, we should expose a `getUser()` method, and define an `IUser` interface that basically wraps the above call. The question is what package?

jpeterson:
For what its worth, I first went looking for it in `arcgis-rest-auth`'s `UserSession` class.

But as I thought about it, this feels almost directly analogous to `arcgis-rest-request`'s `IPortal` and `getSelf()`/`getPortal()` methods. In the same way, I would expect to be able to fetch the currently logged in user as well as other users (by username). Worth noting that the shape of the object returned there would be determined by the role of the currently logged in user.

jgravois:
my first instinct is to put it in `arcgis-rest-auth`.

tomwayson:
I feel like we need a new package (@esri/arcgis-rest-users) to manage users. It would have `getUser()` and other functions for all the operations under the ["Community - User" left nav here](https://developers.arcgis.com/rest/users-groups-and-items/user.htm).

`IUser` itself will probably end up in common types and shared amongst the auth and groups packages.

jgravois:
@tomwayson and i _both_ ultimately got our way.

`v1.1.2` included this:

```js
import { getUser } from '@esri/arcgis-rest-users';

getUser("jsmith") // returns IUser
```
in `v1.4.0` we also introduced support for
```js
import { UserSession } from '@esri/arcgis-rest-auth';

const session = new UserSession({
  username: "jsmith",
  password: "123456"
})

session.getUser() // returns IUser and caches the response
```

PR #170, opened 2018-04-04
feat(new users package): add rest-users with a single method

jgravois:
* didnt implement `getSelf()` because it'd cause a global namespace collision, but the following three signatures are supported

```js
getUser("johnqpublic")
// returns info available to anonymous requests

getUser({ authentication: session })
// returns information about the currently logged in user

getUser({
  authentication: session,
  username: "otheruser"
})
// returns information about an explicitly defined user (based on level of permission of the authenticated session)
```

AFFECTS PACKAGES:
@esri/arcgis-rest-users

ISSUES CLOSED: #159

### DIFF
--- a/packages/arcgis-rest-users/README.md
+++ b/packages/arcgis-rest-users/README.md
@@ -20,9 +20,19 @@ npm install @esri/arcgis-rest-users
 ```js
 import { getUser } from '@esri/arcgis-rest-groups';
 
-getUser("johnqpublic")
+getUser("jsmith")
     .then(response => {
-        console.log(response) // {}
+        /*
+        {
+            firstName: "John",
+            lastName: "Smith",
+            tags: ["GIS Analyst", "City of Redlands"],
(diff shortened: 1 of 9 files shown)


packages/arcgis-rest-users/README.md
```diff
@@ -0,0 +1,61 @@
...
+```bash
+npm install @esri/arcgis-rest-users
+```
+
+```js
+import { getUser } from '@esri/arcgis-rest-groups';
+
+getUser("johnqpublic")
+    .then(response => {
+        console.log(response) // {}
```
noahmulfinger: The comment here should probably have more than `{}`. Really minor though.

tomwayson: Should really remove the whole `console.log` statement, no?

jgravois: agreed (and fixed)

Issue #171, opened 2018-04-05
IItem interface shouldn't require properties except id, owner


noahmulfinger:
The `IItem` interface currently requires:
```
owner
title
type
tags
typeKeywords
```
This is an issue in functions like `updateItem`, which requires an `IItem` be passed in. So any call to `updateItem` must include the listed properties. I think it makes sense to just require `id` and `owner`, because these are generally needed to access the item's endpoint, but the rest should be optional. If a function requires an additional property, the interface can be extended.

jgravois:
> I think it makes sense to just require `id` and `owner`.

that sounds fine to me.

tomwayson:
>  So any call to updateItem must include the listed properties

So is that all of the ones listed above, or just `id` and `owner`?

PR #184, opened 2018-05-02
fix(IItem): make id and owner the only required properties of IItem

jgravois:
AFFECTS PACKAGES:

ISSUES CLOSED: #171

### DIFF
--- a/packages/arcgis-rest-items/src/items.ts
+++ b/packages/arcgis-rest-items/src/items.ts
@@ -24,7 +24,7 @@ export interface IItemIdRequestOptions extends IUserRequestOptions {
    */
   id: string;
   /**
-   * Item owner username (by default authentication session will be used).
+   * Item owner username. If not present, `authentication.username` is utilized.
    */
   owner?: string;
 }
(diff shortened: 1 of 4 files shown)


jgravois:
whoops.

i didn't see any errors locally because i didn't run `npm run build` after making the change.

noahmulfinger:
@jgravois Looks like fixing #171 requires a more fine-grained approach. We should probably just make all `IItem` properties optional and then extend `IItem` whenever a function requires certain properties.

tomwayson:
Good idea @noahmulfinger - I suggest something like`IUpdateItemParams` which inherits from `IItem`

jgravois:
> We should probably just make all IItem properties optional

Agreed.

a perfect example of why is `createItemInFolder()`. up until this morning it ignored `item.owner` entirely in favor of `requestOptions.owner` or `requestOptions.authentication.username`

i've tacked on another commit which includes the value in the decision tree (with more precedence than either of the other two sources).

jgravois:
after chatting w/ @tomwayson briefly i'm convinced that the correct order of precedence is

1. requestOptions.owner
2. requestOptions.item.owner
3. requestOptions.authentication.username

i did a little manual testing and confirmed that regardless of whether `item.owner` is omitted in the request or contains a value that doesn't match the url route, the REST API will overwrite with the appropriate JSON. i'm happy about that.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -34,10 +34,6 @@ export interface IItemJsonDataRequestOptions extends IItemIdRequestOptions {
    * JSON object to store
    */
   data: any;
-  /**
```
jgravois: removed a duplicate property present in the interface being extended.

PR #187, opened 2018-05-09
Oauth ie edge fix

patrickarlt:
IE and Edge will throw errors for any non-string object that comes from another frame or window. In this case passing the date object from the oauth popup window back to the main window will result in `Permission denied` errors when trying to access `tokenExpires` which will result in errors when actually trying to use this for a request.

The solution is to `JSON.stringify` everything that crosses the window boundary and `JSON.parse` it on the other end.

### DIFF
--- a/demos/oauth2-browser/authenticate.html
+++ b/demos/oauth2-browser/authenticate.html
@@ -22,7 +22,7 @@
       function processAuthentication() {
         window.location.href = '/';
         session = arcgisRest.UserSession.completeOAuth2({
-          clientId,
+          clientId: clientId,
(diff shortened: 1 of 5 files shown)


patrickarlt:
Hold on this. Looks like IE 11 doesn't close the window but Edge is OK.

patrickarlt:
@jgravois I can't for the life if me figure out how to make IE 11 close the popup and call the completion method on the original window. This should probably be merged anyway since it makes it work in edge. But in IE 11 popup Oauth basically works like inline redirect Oauth.

Issue #183, opened 2018-05-02
searchItems function ignores some properties in request options


noahmulfinger:
`searchItems` creates its own options object, so properties like `portal` are ignored.

https://github.com/Esri/arcgis-rest-js/blob/20296abf9ec64f89785b6f8766cb2a3f9b0d9cfc/packages/arcgis-rest-items/src/items.ts#L106-L128

PR #189, opened 2018-05-10
make sure methods w/ union types pass through *all* requestOptions

jgravois:
@noahmulfinger noticed `searchItems()` in #183. `geocode()` had the same problem.


jgravois:
> Not sure why the coverage went down again.

Issue #198, opened 2018-05-15
v1.2.0 Broke Semver: UMD output file names changed


dbouwman:
It seems that the umd package naming changed, perhaps at the 1.2 release, and given how the libraray can be used, that should incur a major version bump.

Additionally, the `arcgis-rest-request` and  `arcgis-rest-feature-service` packages still include the OLD umd files

![image](https://user-images.githubusercontent.com/119129/40082104-39d3dee8-584d-11e8-9924-8d46954f3358.png)

I recommend we re-publish, without the old files, an with a 2.0.0 version number.

dbouwman:
Update: we should yank the 1.2.0 release as well

jgravois:
that was my fault. after i tidied up for the release i accidentally ran a build on an old branch and forgot to clear the output before i published to npm.

i'm fine with publishing v2.0.0, but my preference would be to tackle a few of the other breaking changes i listed in #137 before we do, mostly because none of them warrant a breaking change release of their own.

tomwayson:
FYI - I've [deprecated](https://docs.npmjs.com/cli/deprecate) the 1.2.0 packages.

Issue #196, opened 2018-05-15
Peer dependencies out of sync


tomwayson:
https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-feature-service/package.json#L17-L24

This seems like something we need to fix ASAP and cut a new release.

tomwayson:
Hrm.... maybe not https://github.com/lerna/lerna/issues/1018

I guess we should only be updating the peer deps that rely on the changes in 1.2.0?

jgravois:
i wonder what was keeping them in sync before...

jgravois:
i manually [synchronized](https://github.com/Esri/arcgis-rest-js/blob/v1.2.1/packages/arcgis-rest-auth/package.json) in `v.1.2.1`

tomwayson:
I was wondering how they got back in sync.

PR #199, opened 2018-05-15
fix(umd): strip outdated umd files from npm packages

jgravois:
no code is necessary to resolve the problem reported in #198. we just need to push something into `master` to convince lerna to help us publish a `v1.2.1` release.

ISSUES CLOSED: #198 

AFFECTS PACKAGES:
@esri/arcgis-rest-request
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items


jgravois:
> Do we need to handle #196 in this too?

no. it'd be better to wait until i'm in the middle of actually publishing.

PR #200, opened 2018-05-17
Handle missing tags and typeKeywords properties

alukach:
Hi all. I'm just getting started using `arcgis-rest-js` to automate some tooling (let CI upload new WAB Widgets to our Portal upon successful PR merge) and got snagged on a small issue: [The documentation](https://esri.github.io/arcgis-rest-js/api/common-types/IItem/) shows the `typeKeywords` and `tags` properties of an `IItem` to be optional, however the serializer method expects them to be an array. If they are omitted from an `item` object a `Cannot read property 'join' of undefined` error will be thrown ([ex](https://stackblitz.com/edit/js-pmfwlq?embed=1&file=index.js&hideExplorer=1&hideNavigation=1)).  I think this fix would resolve the issue. I can generate a test if need-be.

Is there any particular reason we have `strictNullChecks: false` in the `tsconfig.json`? I believe this would avoid such issues (although I admit to being very green to TypeScript, so I may be missing something).

Btw, digging the library so far! 💯

### DIFF
--- a/packages/arcgis-rest-items/src/items.ts
+++ b/packages/arcgis-rest-items/src/items.ts
@@ -384,8 +384,9 @@ function serializeItem(item: IItem): any {
   // create a clone so we're not messing with the original
   const clone = JSON.parse(JSON.stringify(item));
   // join keywords and tags...
-  clone.typeKeywords = item.typeKeywords.join(", ");
-  clone.tags = item.tags.join(", ");
+  const { typeKeywords=[], tags=[] } = item;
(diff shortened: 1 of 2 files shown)


jgravois:
:wave: nice catch **and** fix. thank you! 

we definitely recommend adding tags (for the sake of search), but failing to pass them certainly shouldn't induce an error.

> Is there any particular reason we have `strictNullChecks: false`

i think its well worth investigating whether that setting can be reverted. lets track in #201

PR #202, opened 2018-05-21
feat(shareItem): add a function to share an item as the item owner

haoliangyu:
This function covers the API
https://developers.arcgis.com/rest/users-groups-and-items/share-item-as-item-owner-.htm

AFFECTS PACKAGES:
@esri/arcgis-rest-items

### DIFF
--- a/packages/arcgis-rest-items/src/items.ts
+++ b/packages/arcgis-rest-items/src/items.ts
@@ -60,6 +60,25 @@ export interface IItemCrudRequestOptions extends IUserRequestOptions {
   folder?: string;
 }
 
+export interface IItemShareRequestOptions extends IItemIdRequestOptions {
+  /**
+   * A comma-separated list of group IDs with which the item will be shared.
+   */
+  groups?: string;
+  /**
+   * If `true`, this item will be shared with everyone, for example, it will be publicly accessible. If explicitly set to false, the item will not be shared with the public.
+   */
+  everyone?: boolean;
+  /**
+   * If `true`, this item will be shared with the organization. If set to false, the item will not be shared with the organization.
+   */
+  org?: boolean;
+  /**
+   * Set to `true` when the item will be shared with groups with item update capability so that any member of such groups can update the item that is shared with them.
+   */
+  confirmItemControl?: boolean;
+}
+
 // this interface still needs to be docced
 export interface ISearchRequest extends IPagingParams {
   q: string;
@@ -374,6 +393,25 @@ export function removeItemResource(
 }
 
 /**
+ * Share an item as the item owner
+ * @param requestOptions - Options for the request
+ * @return A Promise to share an item
+ */
+export function shareItem(
+  requestOptions: IItemShareRequestOptions
+): Promise<any> {
+  const owner = requestOptions.owner || requestOptions.authentication.username;
+  const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
+    requestOptions.id
+  }/share`;
+
+  // this is a POST only API
+  requestOptions.httpMethod = "POST";
+
+  return request(url, requestOptions);
+}
+
+/**
  * Serialize an item into a json format accepted by the Portal API
  * for create and update operations
  *
@@ -384,7 +422,7 @@ function serializeItem(item: IItem): any {
   // create a clone so we're not messing with the original
   const clone = JSON.parse(JSON.stringify(item));
   // join keywords and tags...
-  const { typeKeywords=[], tags=[] } = item;
+  const { typeKeywords = [], tags = [] } = item;
   clone.typeKeywords = typeKeywords.join(", ");
   clone.tags = tags.join(", ");
   // convert .data to .text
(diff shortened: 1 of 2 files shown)


tomwayson:
This is great @haoliangyu!

However I know that @dbouwman had reasons for making [separate fns to share w/ a group vs sharing w/ everyone](https://github.com/Esri/ember-arcgis-portal-services/#sharing-service) and I wonder if we'll end up needing to do that here as well?

haoliangyu:
@dbouwman @tomwayson 

I don't have any particular opinion on it. This function is just a wrapper for the item-share API. In our use case, we do want to share an item to everyone and a specified open data group, but it doesn't matter whether we use one function or two.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -60,6 +60,25 @@ export interface IItemCrudRequestOptions extends IUserRequestOptions {
   folder?: string;
 }
 
+export interface IItemShareRequestOptions extends IItemIdRequestOptions {
+  /**
+   * A comma-separated list of group IDs with which the item will be shared.
+   */
+  groups?: string;
```
jgravois: i think `groups?: string[]` would be a little more ergonomic and i'm fairly certain that if you just passed the value through to `request()` unmolested, it'd _just work_.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -373,6 +392,25 @@ export function removeItemResource(
...
+export function shareItem(
+  requestOptions: IItemShareRequestOptions
+): Promise<any> {
+  const owner = requestOptions.owner || requestOptions.authentication.username;
+  const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
+    requestOptions.id
+  }/share`;
+
+  // this is a POST only API
+  requestOptions.httpMethod = "POST";
```
jgravois: no need to set "POST" manually. its the default for this API

dbouwman:
@haoliangyu @tomwayson @jgravois In the `ember-arcgis-portal-services` addon, we have a separate `sharing-service` which handles some idiosyncrasies of the sharing api in a deterministic manner.

First off, we seprated the concept of "Access" from sharing-to-groups... [setAccess](https://github.com/Esri/ember-arcgis-portal-services/blob/master/addon/services/sharing-service.js#L17)

As for sharing-to-a-group... the API is not idempotent. So... if you have Item A, and you make an API call to share that to Group B... assuming you are the owner, and a member of Group B, it will succeed.
If you make that call again, the response will tell you that it did not share it with Group B. Spiffy.

Thus, the [changeGroupSharing](https://github.com/Esri/ember-arcgis-portal-services/blob/master/addon/services/sharing-service.js#L66) method does a number of checks so our application code can be ignorant of the API's "features". It checks if the item it already shared to the group, then checks if the calling user is a member of the group, and then finally decides what url to use depending if the user is the item owner or not.

So - that's the background. Initially we started w a method on the `items-service`, but as we ran into the weirdness of the sharing api, we devised this solution, which has been working really well, so I'd recommend implementing something similar - either in this package or another `sharing` package.

packages/arcgis-rest-items/src/items.ts
haoliangyu: :+1: Agree. An array of group Ids makes more sense to the user.

jgravois:
thanks for the backstory @dbouwman!

@haoliangyu, if want to tackle all the extra complexity that dave described and it'd be helpful enough to your own team to warrant the effort, that'd be _excellent_.

if not, lets just name your function `shareItemAsOwner()` instead. that will set us up to scaffold out a bulletproof `shareItem()` that can chain methods together and ensure a deterministic response no matter which user is passed through sometime soon.

cc/ @dmfenton

dmfenton:
let's just bite the bullet and do this the right (painful) way.

dbouwman:
@jgravois could you scaffold out an `arcgis-rest-sharing` package to get @haoliangyu rolling? Once that's done it's no more difficult to put the actual code into one package or another. 

Let's try to avoid having `arcgis-rest-sharing` depend on `arcgis-rest-items` or `arcgis-rest-groups` :)

packages/arcgis-rest-items/src/items.ts
dbouwman: As i noted below - the sharing API is really obtuse, so sharing to multiple groups at the same time can return responses that do not make much sense. I'd highly recommend sharing to a single group at a time so you can make things more deterministic :)

dmfenton: > I'd highly recommend sharing to a single group at a time so you can make things more deterministic 

You mean under the hood right?

haoliangyu:
Thanks for all input! I was planning to use this function in a non-critical feature so there is no presure to push it. At this moment I can use some simple request logic as a replacement. Since `arcgis-rest-js` is a public module, it should be designed and coded more carefully.

packages/arcgis-rest-items/src/items.ts
tomwayson: or maybe the first iteration only takes a single group, second iteration takes an array and and aggregates the success/failure info of each to return in a single response.

haoliangyu:
close this in favor of #207

PR #205, opened 2018-05-24
new sharing module with setAccess() method that includes error trapping

jgravois:
supercedes #202 

more or less a straight port from Emberland.

```js
setItemAccess({
  id: "47f07",
  access: "org", // "public" | "private"
  owner: "c@sey",
  authentication: session
})
  .then(response => console.log(response))
```
- [X] includes a selective check to determine the privileges of the authenticated user

### DIFF
--- a/packages/arcgis-rest-sharing/README.md
+++ b/packages/arcgis-rest-sharing/README.md
@@ -9,7 +9,7 @@
 
 # @esri/arcgis-rest-sharing
 
-> Helper utilities for managing access to content via the ArcGIS REST API that run in Node.js and modern browsers.
+> Helper utilities for managing access to ArcGIS content in Node.js and modern browsers.
 
 ### Example
 
@@ -19,18 +19,18 @@ npm install @esri/arcgis-rest-sharing
 ```
 
 ```js
-import { setAccess } from '@esri/arcgis-rest-sharing';
+import { setItemAccess } from '@esri/arcgis-rest-sharing';
 
-const id = "abc123";
+const itemId = "abc123";
 
-setAccess({
-    id,
-    access: "everyone",
+setItemAccess({
+    id: itemId,
+    access: "public",
     authentication: session
 })
 ```
 
-### [API Reference](https://esri.github.io/arcgis-rest-js/api/request/)
+### [API Reference](https://esri.github.io/arcgis-rest-js/api/sharing/)
 
 ### Issues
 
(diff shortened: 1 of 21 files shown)


packages/arcgis-rest-sharing/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "peerDependencies": {
+    "@esri/arcgis-rest-request": "^1.3.0",
+    "@esri/arcgis-rest-auth": "^1.3.0",
+    "@esri/arcgis-rest-users": "^1.3.0"
+  },
+  "devDependencies": {
+    "@esri/arcgis-rest-request": "^1.3.0",
+    "@esri/arcgis-rest-auth": "^1.3.0",
+    "@esri/arcgis-rest-users": "^1.3.0"
```
jgravois: I need to get rid of these.

packages/arcgis-rest-sharing/src/sharing.ts
```diff
@@ -0,0 +1,104 @@
...
+      } else {
+        return updateAccess(sharingUrl, requestOptions);
+      }
+    });
+  } else {
+    return updateAccess(sharingUrl, requestOptions);
+  }
+}
+
+function updateAccess(
```
jgravois: need to confirm that this method is omitted from the doc.

packages/arcgis-rest-sharing/src/sharing.ts
```diff
@@ -0,0 +1,104 @@
...
+  url: string,
+  requestOptions: ISetAccessRequestOptions
+): Promise<any> {
+  requestOptions.params = {
+    org: false,
+    everyone: false,
+    ...requestOptions.params
+  };
+
+  // FYI: i think we're setting 'items' unnecessarily in: https://github.com/Esri/ember-arcgis-portal-services/blob/master/addon/services/sharing-service.js#L26
```
jgravois: if true, I'll open another PR

packages/arcgis-rest-sharing/README.md
```diff
@@ -0,0 +1,67 @@
...
+const id = "abc123";
+
+setAccess({
+    id,
+    access: "everyone",
+    authentication: session
+})
+```
+
+### [API Reference](https://esri.github.io/arcgis-rest-js/api/request/)
```
jgravois: need to update this url, and probably several others.

jgravois:
_should_ be good to go now.

PR #207, opened 2018-05-25
deterministic group un/sharing methods

jgravois:
supercedes #202 

port of [`bb9ced6`#L66-L127](https://github.com/Esri/ember-arcgis-portal-services/blob/bb9ced61ba5157b18f1e48634462d1844a9953c6/addon/services/sharing-service.js#L66-L127)

does **not** depend on `rest-items`/`rest-users`/`rest-groups`
```js
const id = "47f";
const groupId = "b3c";

shareItemWithGroup({
  id,
  groupId,
  owner: "c@sey", // otherwise we assume session.username
  authentication: session
})
  .then(response => console.log(response))

unshareItemWithGroup({
  id,
  groupId,
  authentication: session
})
  .then(response => console.log(response))
```
includes intermediate checks to determine if:
- [X] the authenticated user is an administrator of the organization
- [X] the item was _already_ shared with the group
- [X] the authenticated user is a member of the group (if owner of the item being shared)
- [X] whether the authenticated user is an admin/owner of the group (if they _don't_ own the item being shared)


packages/arcgis-rest-sharing/src/group-sharing.ts
```diff
@@ -0,0 +1,205 @@
...
+      ) {
+        // and send back the same response structure ArcGIS Online would
+        const response = { itemId: requestOptions.id, shortcut: true } as any;
+        response[resultProp] = [];
+        return response;
+      } else {
+        // next check to ensure the user is a member of the group
+        return getUserMembership(requestOptions)
+          .then(membership => {
+            if (membership === "nonmember") {
```
jgravois: this is a deviation from a harmless [flaw](https://github.com/Esri/ember-arcgis-portal-services/blob/bb9ced61ba5157b18f1e48634462d1844a9953c6/addon/services/sharing-service.js#L84-L85) in the EAPS logic.

packages/arcgis-rest-sharing/src/group-sharing.ts
```diff
@@ -0,0 +1,205 @@
...
+              throw Error(
+                `This item can not be ${
+                  requestOptions.action
+                }d by ${username} as they are not a member of the specified group ${
+                  requestOptions.groupId
+                }.`
+              );
+            } else {
+              // if orgAdmin or owner (and member of group) share using the owner url
+              if (owner === username || admin) {
```
jgravois: presupposes that the EAPS code [_comment_](https://github.com/Esri/ember-arcgis-portal-services/blob/bb9ced61ba5157b18f1e48634462d1844a9953c6/addon/services/sharing-service.js#L89-L90) is correct, not the code.

packages/arcgis-rest-sharing/src/helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+  )}/items/${requestOptions.id}/share`;
+}
+
+export function isItemOwner(requestOptions: ISharingRequestOptions): boolean {
+  const username = requestOptions.authentication.username;
+  const owner = requestOptions.owner || username;
+  return owner === username;
+}
+
+export function isOrgAdmin(
```
jgravois: on a bike ride after work two things dawned on me

1. it makes more sense to locate this function inside `arcgis-rest-auth`
2. we could even append the information to the active session (like torii does) to short circuit another request.

dbouwman: def cache this in the session, no need to re-request portal/self or user/self to a role name. We should also consider moving over some of the other torii-session helpers - re: privs. Exp in Hub, privs will become a much larger part of how we control things

patrickarlt: @jgravois I definitely agree with this. 👍

jgravois: completed the user/self refactor in #212

packages/arcgis-rest-sharing/test/access.test.ts
```diff
@@ -13,6 +12,11 @@ const SharingResponse = {
 };
 
 describe("setItemAccess()", () => {
+  // make sure session doesnt cache metadata
+  beforeEach(function() {
+    MOCK_USER_SESSION.userInfo = null;
```
tomwayson: as we mentioned on slack, an alternative would be for `"./mocks/sharing"` to `export` a POJO and have this create a `new UserSession()` from the POJO each time.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -498,7 +518,36 @@ export class UserSession implements IAuthenticationManager {
...
+   * ```js
+   * session.getUserInfo()
+   *   .then(response => {
+   *     console.log(response.role); // "org_admin"
+   *   })
+   * ```
+   *
+   * @returns A Promise that will resolve with the data from the response.
+   */
+  getUserInfo(): Promise<IUser> {
```
tomwayson: Since we're returning an `IUser` I think a better name for this fn will be `getUser()`. As I suggested above I think we should remove the `userInfo` property. If we do, we could use `this._user` to cache the response otherwise update it's definition to `user?: IUser`.

packages/arcgis-rest-sharing/test/access.test.ts
jgravois: yesterday i tried the _POJO approach_ (and failed).

packages/arcgis-rest-sharing/src/helpers.ts
```diff
@@ -61,30 +59,24 @@ export function getUserMembership(
...
-      if (response.admins.includes(username)) {
-        result = "admin";
-      }
-      if (response.users.includes(username)) {
-        result = "user";
+  return session
+    .getUser()
+    .then((user: IUser) => {
+      if (user.groups) {
+        user.groups.forEach(function(group: IGroup) {
```
tomwayson: nitpick, but probably best to short circuit w/ `.some()` instead of `forEach()`.

jgravois: research notes:
* [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) returns a boolean 
* [`find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) return the element, but the method isn't supported in IE

ten minutes wasn't enough time for me to learn how to use rollup's babel plugin to transpile down the reference.

my vote:
- [x] refactor the _other_ call to `find()` i found
- [x] rebase this god forsaken monster
- [ ] **call it a day**.

tomwayson: Here's what I meant by using `.some()`

```js
        user.groups.some(function(group: IGroup) {
          const foundIt = group.id === requestOptions.groupId;
          if (foundIt) {
            result = group.userMembership.memberType;
          }
          return foundIt;
        });
```

jgravois: 💡
👦

i 🙇 to your superior foo.

tomwayson: my foo is not superior, just older

:shipit:

PR #212, opened 2018-06-04
Feat/share groups auth refactor

jgravois:
addresses https://github.com/Esri/arcgis-rest-js/pull/207#discussion_r191445162

```js
import { UserSession, getUserInfo } from "@esri/arcgis-rest-auth";

      username: "jgravois",
      password: "10speed"
    });

session.getUserInfo()
  .then(response => console.log(response));

// subsequent calls will return same info from cache
```

~~the signature below would be better, but i haven't figured out how to accomplish it yet.~~

to do:
- [x] add doc
- [x] get test coverage _back_ to 100%
- [x] make sure the method hangs off session itself.


packages/arcgis-rest-auth/src/user-info.ts
```diff
@@ -0,0 +1,28 @@
...
+    return new Promise(resolve => resolve(session.userInfo));
+  } else {
+    const url = `${session.portal}/community/users/${encodeURIComponent(
+      session.username
+    )}`;
+    return request(url, {
+      authentication: session,
+      httpMethod: "GET"
+    }).then(response => {
+      session.userInfo = response;
```
jgravois: should i serialize the response here?

packages/arcgis-rest-sharing/test/access.test.ts
```diff
@@ -13,6 +13,10 @@ const SharingResponse = {
 };
 
 describe("setItemAccess()", () => {
+  beforeEach(function() {
+    MOCK_USER_SESSION.userInfo = null;
+  });
+
```
jgravois: this is janky, but the alternative below is error inducing.

```js
beforeEach(function() {
  import { MOCK_USER_SESSION } from "./mocks/sharing";
});
// Unable to compile TypeScript: An import declaration can only be used in a namespace or module.
```

Issue #137, opened 2018-03-01
v2.0.0 checklist


jgravois:
a list of breaking changes that we _could_ group together. feedback welcome.

- [ ] ~~drop the 'r' from `arcgis-rest-geocoder`?~~
- [ ] ~~drop the 's' from `arcgis-rest-items`?~~
- [ ] ~~drop the 's' from `arcgis-rest-groups`?~~
- [ ] ~~rename `arcgis-rest-common-types` to `arcgis-rest-common` and move the growing collection of shared utility methods in [`request`](https://github.com/Esri/arcgis-rest-js/tree/e2b05b2a849fd8c84797bbb5264c59965a911b25/packages/arcgis-rest-request/src/utils) (that aren't actually _needed_ by `request`) there along with the types~~
- [ ] ~~`getAsyncStuff`> `fetchAsyncStuff`? (and use `get` for getters instead)~~
- [ ] ~~consolidate `feature-service` and `feature-service-admin`~~
- [ ] ~~start bundling `common` _inside_ upstream packages?~~
- [ ] ~~keep documenting the structure, but stop declaring a named interface, for unique responses.~~
- [x] remove `serviceInfo()` from `arcgis-rest-geocoder`
- [x] `esriGeometryType` > `GeometryType`
- [x] `esriUnits` > `Units`
- [x] `esriFieldTypes` > `FieldTypes`
- [x] `folder` > `folderId` in [`IItemCrudRequestOptions`](https://esri.github.io/arcgis-rest-js/api/items/IItemCrudRequestOptions/) (see #236)
- [x] fix capitalization of [`IOauth2Options`](https://github.com/Esri/arcgis-rest-js/blob/72ed12224b8dcbd53121d59cd0f72d1e3ffe9e39/packages/arcgis-rest-auth/src/UserSession.ts#L56) (to be `IOAuth2Options`)
- [x] replace `"adds"/"updates"/"deletes"` in feature service crud functions with consistent use of `"features"` 
- [x] remove `appendCustomParams()` re-export from feature-service (see #365)
- [x] change signature of `getItemResources()` (for consistency)
- [x] remove `IGeocodeParams` #468
- [x] remove `serializeGroup()`#475
- [x] remove [`IItemResourceAddRequestOptions`](https://esri.github.io/arcgis-rest-js/api/items/IItemResourceAddRequestOptions/)
- [x] consolidate `items`, `groups`, `sharing` and `users` packages into `@esri/arcgis-rest-portal`
[...]

PR #214, opened 2018-06-06
refactor(feature-service): update signatures

jgravois:
to stop leaning on `params` directly

AFFECTS PACKAGES:

addresses @patrickarlt's https://github.com/Esri/arcgis-rest-js/issues/137#issuecomment-393362439

this is one rare instance in which it will 'just work' if developers don't update their code.


packages/arcgis-rest-feature-service/src/features.ts
```diff
@@ -83,22 +89,11 @@ export interface IQueryFeaturesParams extends ISharedQueryParams {
   quantizationParameters?: any;
   returnCentroid?: boolean;
   resultType?: "none" | "standard" | "tile";
-  // TODO: is Date the right type for epoch time in milliseconds?
-  historicMoment?: Date;
+  historicMoment?: number;
```
tomwayson: I think my idea was our API would take in dates, but convert them to epoch time (numbers) before sending the requests. That said, I didn't implement it. 😄 

So I agree w/ this change. We can take in numbers for now and then later add the option to take dates (`historicMoment?: number | Date`).

packages/arcgis-rest-feature-service/src/features.ts
```diff
@@ -43,23 +51,21 @@ export interface ISharedQueryParams extends IParams {
...
+  extends ISharedQueryParams,
+    IRequestOptions {
+  /**
+   * Layer service url.
+   */
+  url: string;
   objectIds?: number[];
   relationParam?: string;
   // NOTE: either time=1199145600000 or time=1199145600000, 1230768000000
   time?: Date | Date[];
```
tomwayson: I think we should change this to `number | number[]` as well to be consistent w/ `historicMoment`.

packages/arcgis-rest-feature-service/src/features.ts
```diff
@@ -384,8 +407,30 @@ export function deleteFeatures(
...
+  appendCustomParams(requestOptions, options);
+
   // mixin, don't overwrite
   options.params.objectIds = requestOptions.deletes;
 
   return request(url, options);
 }
+
+function appendCustomParams(
+  oldOptions: IQueryFeaturesRequestOptions,
```
tomwayson: Shouldn't this be:oldOptions: IQueryFeaturesRequestOptions,

tomwayson:
Hm... I'm confused that link doesn't make its way to the docs: https://esri.github.io/arcgis-rest-js/api/feature-service/IQueryFeaturesRequestOptions/

packages/arcgis-rest-feature-service/src/features.ts
jgravois: They _should_, but this would require adding `[key: string]: any;` to all four interfaces to avoid TypeScript complaints.

```
Element implicitly has an 'any' type because type 'IQueryFeaturesRequestOptions | IAddFeaturesRequestOptions | IUpdateFeaturesRequestOptions | IDele...' has no index signature. (7017)
```

tomwayson: actually, I guess just `oldOptions` should be that union type, is that what you tried?

jgravois: exactly. my current approach is definitely incorrect, but for some reason it doesn't bother TypeScript (or me).

we could also:
* append the generic key to each of the interfaces we'd like to union (not DRY, but that's not a big deal)
* create and export a generic `ICrudRequestOptions` interface and have each of the ones we'd like to union extend _that_ (this would require it to appear in the doc, which feels like overkill)
* create and _not_ export a generic `ICrudRequestOptions` interface and use it to declare the type to the utility method _without_ extending the existing interfaces to inherit from it (which isn't really any better than what i'm doing now).
* just do this:

```ts
export function appendCustomParams(
  oldOptions: { [key: string]: any; },
  newOptions:
    | IQueryFeaturesRequestOptions
    | IAddFeaturesRequestOptions
    | IUpdateFeaturesRequestOptions
    | IDeleteFeaturesRequestOptions
) { /**/ }
```

tomwayson: > for some reason it doesn't bother TypeScript (or me).

sorry, it bothers the shit out of me. It's not you, it's TS.

I suspect the problem is the `[key: string]: any` in `IQueryFeaturesRequestOptions` is what causes the `Element implicitly has any` error when you use a union type for `oldOptions`. Can we remove that?

Also, if you don't want to deal w/ this, just LMK and I'll pull and try to reason through it myself.

PR #218, opened 2018-06-13
fix(amd): add output.amd.id to rollup.config

jgravois:
yesterday i was scratching my head for quite awhile trying to reference this lib in a dead simple application alongside the ArcGIS API for JavaScript.

my research indicated that UMD's register themselves automatically as _anonymous_ AMD modules when a loader is detected, but if that was indeed happening, i sure couldn't figure out how to reference to them.

declaring an explicit [`amd.id`](https://rollupjs.org/guide/en#big-list-of-options) in the rollup build makes the following possible.

```js
require(["@esri/arcgis-rest-request"/*, etc. */], function(arcgis/*, etc. */) {
  arcgis.request(url).then(response => {
      console.log(response);
  });
});
```

### DIFF
--- a/docs/src/guides/amd-requirejs-dojo.md
+++ b/docs/src/guides/amd-requirejs-dojo.md
@@ -7,3 +7,33 @@ group: 1-get-started
 ---
 
 # Get Started with ArcGIS REST JS and AMD
+
+```html
+<!DOCTYPE html>
+<html>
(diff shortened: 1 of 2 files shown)


jgravois:
closing in favor of f6e70a8bcc852377c83a1d32b48d5e44f02e73b6

Issue #208, opened 2018-05-28
create mechanism to share login info w/ JSAPI


jgravois:
for lazy loading workflows. ie. transitioning from non-map > map UI.

@noahmulfinger are you guys already doing this in your vector file editor?

patrickarlt:
@jgravois for they style editor we have a method that looks like:

```ts
class AuthService 
  // ...

  registerToken(session: UserSession) {
    loadModules(["esri/identity/IdentityManager"], {
      url: "https://js.arcgis.com/4.7/"
    }).then(([esriId]: [__esri.IdentityManager]) => {
      esriId.registerToken({
        expires: session.tokenExpires.getTime(),
        server: session.portal,
        ssl: true,
        token: session.token,
        userId: session.username
      });
    });
  }

  // ...
}
```

I do like the idea for bi-directionally that @qlqllu mentioned. I think the rough equivalent would be:

```ts
import { UserSession } from "@esri/arcgis-rest-auth";

loadModules(["esri/identity/IdentityManager"], {
  url: "https://js.arcgis.com/4.7/"
}).then(([esriId]: [__esri.IdentityManager]) => {
  var credential = esriId. getCredential(portalUrl);
  var session = new UserSession({
    clientId,
    portal: credential.server,
    token: credential.token,
    tokenExpires: new Date(credential.expires),
    username: credential.userId
  });
});
```

qlqllu:
I have an idea about the bi-direction sync:

Because the API is lazy load, so when the `IdentityManager` is first loaded, we can call an API in this lib to register the IdentityManager. In this register API, we can sync these two sign in info.

tomwayson:
Maybe we can get some easy wins by starting w/ the low level helper fns that would later support a more comprehensive strategy for this.

For example: the `UserSession` could expose a `getRegisterTokenProperties()` fn that returns an object w/ the properties that `registerToken()` expects. Fortunately those properties seem to be the same between [3.x](https://developers.arcgis.com/javascript/3/jsapi/identitymanager-amd.html#registertoken) and [4.x](https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-IdentityManager.html#registerToken). So in @qlqllu's scenario, we'd call `esriId.registerToken(userSession.getRegisterTokenProperties())` once the `IdentityManager` is loaded.

[...]

PR #219, opened 2018-06-13
add UserSession.getCredential()

jgravois:
first pass at #208

PRing against the `amd` branch for the sake of the demo.

```js
require([
  "@esri/arcgis-rest-auth",
  "@esri/arcgis-rest-items",
  "esri/identity/IdentityManager",
  "esri/WebMap"
], function(
  arcgisAuth, arcgisItems, esriId, WebMap
) {
  const session = new arcgisAuth.UserSession({
    username: "arcgis_python",
    password: "P@ssword123"
  });

  // https://geosaurus.maps.arcgis.com/home/item.html?id=728043ac6a574a00b8f1cd5ee0eae8cd
  arcgisItems.getItemData("728043ac6a574a00b8f1cd5ee0eae8cd", { authentication: session })
    .then(response => {
      esriId.registerToken(session.getCredential());

      new WebMap({
        portalItem: {
          // dig the webmap id out of the response
          id: response.map.itemId
        }
      });
    });
});
```

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -5,7 +5,6 @@ import * as http from "http";
 import {
   request,
   ArcGISAuthError,
-  ArcGISRequestError,
   IAuthenticationManager
 } from "@esri/arcgis-rest-request";
 import { generateToken } from "./generate-token";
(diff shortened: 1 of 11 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -510,6 +520,19 @@ export class UserSession implements IAuthenticationManager {
     });
   }
 
+  // do we need to allow for passing a url?
+  // trap for when session isnt fully hydrated?
+  // should this return a promise?
```
tomwayson: I don't think so - let the consuming app decide if the half baked credential is valid for their use case or not

jgravois:
tacked on another commit to utilize `dojoConfig` to identify the location of our UMDs instead of `<script>` tags and revert the rollup change.

PR #221, opened 2018-06-14
docs(demos): jsapi integration demo shows more typical scenario

tomwayson:
don't load arcgis-rest-js using dojo
call require() _after_ loading item

AFFECTS PACKAGES:
@esri/jsapi-integration


tomwayson:
TBH - I was half tempted to go all the way w/ this and actually use esri-loader and could tack on a follow on commit to do that if you think that's worthwhile.

PR #227, opened 2018-06-19
feat(auth): toCredential/fromCredential

jgravois:
to support passing arcgis-rest-js and jsapi credentials back and forth

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #208


tomwayson:
I mean, it makes sense, that _is_ the server.

Can we be smart about appending `/sharing/rest` if needed?

PR #230, opened 2018-06-20
throw NOT_FEDERATED error when no 'owningSystemUrl' is found

jgravois:
fix to ensure we throw an error when a developer attempts to fetch a token for a standalone instance of ArcGIS Server that doesn't advertise an `owningSystemUrl` (like [this one](http://sampleserver6.arcgisonline.com/arcgis/rest/info?f=json)).

STR:
```js
const session = new arcgisRest.UserSession({
  username: "johnq",
  password: "public"
});

session.getToken("http://sampleserver6.arcgisonline.com/arcgis/rest/services/")
  .catch(err => {
    err.code // "NOT_FEDERATED"
  })
```

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -662,17 +662,19 @@ export class UserSession implements IAuthenticationManager {
 
     this._pendingTokenRequests[root] = request(`${root}/rest/info`)
       .then((response: any) => {
-        // a stand-alone instance of ArcGIS Server won't advertise federation at all
-        return response.owningSystemUrl
-          ? response.owningSystemUrl
-          : "https://foo.bar";
+        return response.owningSystemUrl;
       })
       .then(owningSystemUrl => {
         /**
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -662,7 +662,10 @@ export class UserSession implements IAuthenticationManager {
 
     this._pendingTokenRequests[root] = request(`${root}/rest/info`)
       .then((response: any) => {
-        return response.owningSystemUrl;
+        // a stand-alone instance of ArcGIS Server won't advertise federation at all
+        return response.owningSystemUrl
+          ? response.owningSystemUrl
+          : "https://foo.bar";
```
tomwayson: personally I'd prefer to just return `response.owningSystemUrl` and then below check for `if (!owningSystemUrl || !new RegExp(owningSystemUrl).test(this.portal)) {`

jgravois:
> personally I'd prefer to just return `response.owningSystemUrl`

done (and rebased).

PR #232, opened 2018-06-20
fix(GET): switch to POST for requests > 2048 characters by default

jgravois:
use the maximum value supported by IE as the default maximum length of a GET request.

i think it makes sense to do this in `request`. another option would be to set the value in methods that are specifically likely to overrun the limit (like `queryFeatures`).

AFFECTS PACKAGES:
@esri/arcgis-rest-request

### DIFF
--- a/packages/arcgis-rest-feature-service/src/query.ts
+++ b/packages/arcgis-rest-feature-service/src/query.ts
@@ -100,7 +100,7 @@ export interface IQueryFeaturesResponse extends IFeatureSet {
  * getFeature({
  *   url,
  *   id: 42
- * };)
+ * })
  *   .then(feature => {
  *     console.log(feature.attributes.FID); // 42
  *   });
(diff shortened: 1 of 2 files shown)


tomwayson:
First, I don't think w should cater to the lowest (un)common debrowserinator. 

> the maximum value supported by IE

Are you sure about that?

image

I'm pretty sure that changes every time I google it (about once a year).

Also, it's a bit odd to me to see defaults of `httpMethod: 'POST', maxUrlLength: 2048` right next to each other (I should have named it `maxGetUrlLength`). I bet that'll be a bit confusing in the docs too.

That makes me want to say, let's do it at the individual fn level. `queryFeatures()` can set `httpMethod: 'GET', maxUrlLength: 2048`. However, I think it'll be a bit of pain to figure out which fns are "likely to overrun the limit."

Honestly I say we hold on this until we get complaints about issues in IE.

jgravois:
> Honestly I say we hold on this until we get complaints about issues in IE.

works for me :)

Issue #161, opened 2018-03-20
failure to fetch a token for hosted services associated with a portal


jgravois:
in testing #160 i _think_ i uncovered another bug.

```js
// real portal instance, fake username and password
const session = new UserSession({
  username: "c@sey",
  password: "jones",
  portal: "https://portalhostds.ags.esri.com/gis/sharing/rest"
});

session.getToken("https://serverhostds.ags.esri.com/gis/rest/services/Hosted/restjs/FeatureServer")
  .then(token => {
    console.log(token);
  })
```
snooping the traffic i noticed that its the last parameter in the request below that induces the error.
```
https://portalhostds.ags.esri.com/gis/sharing/rest/generateToken
username=casey
password=jones
f=json
referer=localhost:8080
serverUrl=http://serverhostds.ags.esri.com/arcgis // there's your problem

{
  "error": {
    "code":400,
    "message":"Unable to generate token.",
    "details":["Unable to get 'token' from the request or cookie, for server token."]
  }
}
```

i'm pretty sure this is because hosted datastores are _not_ federated the same way standalone versions of ArcGIS Server might be.

this can be fixed by making the check inside [`UserSession.getToken()`](https://github.com/Esri/arcgis-rest-js/blob/3650bd242185024238cfc4630fc77bdd4bfcef3a/packages/arcgis-rest-auth/src/UserSession.ts#L480-L491) more generic to compare root domains instead of the entire url, but doing that causes all sorts of errors in other tests.

```js
const portalDomain = this.portal.split("/")[2];
// isolate the likely root domain from the request url
const urlDomain = url.split("/")[2];
const urlSplit = urlDomain.split(".");
const urlRootDomain = urlSplit[urlSplit.length - 2] + "." + urlSplit[urlSplit.length - 1]
// ...
// now we're just making sure 'portalhostds.ags.esri.com' includes 'esri.com'
} else if (new RegExp(urlRootDomain).test(portalDomain)) {
```
still scratching my head about this one. but perhaps there is another way a portal announces its built in datastore?

PR #233, opened 2018-06-20
fix(enterprise): ensure a brand new token can be generated for servers federated with ArcGIS Enterprise

jgravois:
> this PR description has been edited _heavily_. previous drafts are available from the <kbd>edited &#9660;</kbd> dropdown 

proposed fix to enable the generation of tokens for ArcGIS Data Store content associated with ArcGIS Enterprise.

the tweak i'm proposing ~~generalizes the comparison of the two relevant urls and gives a ✅ as long as the portal and service url share the same `sub.domain.com`.~~ introspects the current session to see whether a token is already available. if not, and both username/password are present (indicating that the session was _not_ instantiated for an OAuth flow), then a request to fetch one manually is fired.

STR:
```js
// real portal instance, fake username and password
  username: "c@sey",
  password: "jones",
  portal: "https://portalhostds.ags.esri.com/gis/sharing/rest"
});

session.getToken("https://serverhostds.ags.esri.com/gis/rest/services/Hosted/restjs/FeatureServer")
  .then(token => {
    console.log(token);
  })
```

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #161

paging @slibby in the hopes that he can provide a sanity check.

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -598,8 +598,8 @@ export class UserSession implements IAuthenticationManager {
    */
   getToken(url: string) {
     if (
-      (/^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal),
-      /^https?:\/\/\S+\.arcgis\.com.+/.test(url))
+      /^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal) &&
+      /^https?:\/\/\S+\.arcgis\.com.+/.test(url)
     ) {
       return this.getFreshToken();
     } else if (new RegExp(this.portal).test(url)) {
@@ -686,11 +686,24 @@ export class UserSession implements IAuthenticationManager {
         return response.authInfo.tokenServicesUrl;
       })
       .then((tokenServicesUrl: string) => {
-        return generateToken(tokenServicesUrl, {
-          token: this.token,
-          serverUrl: url,
-          expiration: this.tokenDuration
-        });
+        if (this.token) {
+          return generateToken(tokenServicesUrl, {
+            token: this.token,
+            serverUrl: url,
+            expiration: this.tokenDuration
+          });
+          // generate an entirely fresh token if necessary
+        } else {
+          return generateToken(tokenServicesUrl, {
+            username: this.username,
+            password: this.password,
+            expiration: this.tokenDuration
+          }).then((response: any) => {
+            this._token = response.token;
+            this._tokenExpires = new Date(response.expires);
+            return response;
+          });
+        }
       })
       .then(response => {
         this.trustedServers[root] = {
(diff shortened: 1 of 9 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -597,12 +597,22 @@ export class UserSession implements IAuthenticationManager {
...
+    const portalDomain =
+      domainParts[domainParts.length - 3] +
+      "." +
+      domainParts[domainParts.length - 2] +
+      "." +
+      domainParts[domainParts.length - 1];
     if (
-      (/^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal),
-      /^https?:\/\/\S+\.arcgis\.com.+/.test(url))
+      /^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal) &&
```
jgravois: substituting `&&` is actually something that @noahmulfinger caught awhile back.

_both_ of these tests need to be truthy. the `,` has the effect of disregarding the first match entirely.

patrickarlt:
@jgravois I think need to figure out a way to do this without assuming things like hosted data stores are always on the same domain as their portals. Depending on how users have deployed their store and portal this change might be totally meaningless.

Also isn't data store just a backend for a server that ends up getting federated with a portal anyway?

How does the JS API handle these calls?

slibby:
1.  is it even appropriate for us to assume that a service that shares the same base domain as an instance of ArcGIS Enterprise will use the same authentication scheme?
**No, we cannot assume this**
2.  if so, should the check be further weakened to compare only domain.com
**No, this is likely even more complicated**
3.  is there another, superior way to sniff out this type of relationship?
**Generally, we would use** `/rest/info`  **on the AGS endpoint to identify the correct Token URL for that server. Similarly, you can use /portal/sharing/rest/portals/self/servers with a valid Portal token to identify valid federated servers as well.**

jgravois:
thanks for confirming @slibby!

thinking through this more deeply, the problem isn't actually a failure to associate the two resources. we should instead introspect the current session and fetch an entirely new token with the user's credentials and (making sure to omit `serverUrl`) when one isn't already attached.

patrickarlt:
@jgravois I still want to do a deeper review of this and test it out. I'm not sure the right approach is falling back to username and password. We still want o encourage users to us Oauth even on portals I think.

jgravois:
> We still want to encourage users to us Oauth even on portals I think.

i agree that we should promote OAuth as the preferred method of authentication and i'm happy to start working on some doc to explain why/how. 

that said, devs can (and will) attempt to instantiate username/password UserSessions, both in node apps, or even just during testing in the browser, so we need to make sure that works too.

jgravois:
i'm still in favor of landing this fix primarily because the equivalent pattern already works just fine for ArcGIS Online.

```js
// fake username and password
  username: "c@sey",
  password: "jones"
});

session.getToken("https://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer/")
  .then(token => {
    console.log(token); // HWYDXKHz4W-x979NNq6_E...
  })
```

PR #235, opened 2018-06-25
Get, add, update and delete attachment modules.

COV-GIS:
Here's the attachment modules for review. Tests and demo coming as soon as time allows.

### DIFF
--- a/packages/arcgis-rest-feature-service/src/addAttachment.ts
+++ b/packages/arcgis-rest-feature-service/src/addAttachment.ts
@@ -6,7 +6,7 @@ import { request, IRequestOptions } from "@esri/arcgis-rest-request";
 import { IEditFeatureResult } from "./helpers";
 
 /**
- * Request options to for adding a related attachment to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
+ * Request options for adding a related attachment to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
  *
  * @param url - Feature service url.
  * @param featureId - Unique identifier of feature to add related attachment.
@@ -54,9 +54,6 @@ export function addAttachment(
 
   // `attachment` --> params: {}
   options.params.attachment = requestOptions.attachment;
-
-  // force POST
-  options.httpMethod = "POST";
 
   return request(`${options.url}/${options.featureId}/addAttachment`, options);
 }
--- a/packages/arcgis-rest-feature-service/src/addAttachment.ts
+++ b/packages/arcgis-rest-feature-service/src/addAttachment.ts
@@ -1,21 +1,20 @@
-/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { request } from "@esri/arcgis-rest-request";
+import { request, IRequestOptions } from "@esri/arcgis-rest-request";
 
-import { IFeatureRequestOptions } from "./query";
-
-import { IEditFeatureResult, appendCustomParams } from "./helpers";
+import { IEditFeatureResult } from "./helpers";
 
 /**
  * Request options to for adding a related attachment to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
  *
  * @param url - Feature service url.
- * @param id - Unique identifier of feature to add related attachment.
+ * @param featureId - Unique identifier of feature to add related attachment.
  * @param attachment - File to be attached.
- * @param params - Additional parameters to be sent via the request. See reference docs.
  */
-export interface IAddAttachmentOptions extends IFeatureRequestOptions {
+export interface IAddAttachmentOptions extends IRequestOptions {
+  url: string;
+  featureId: number;
   attachment: File;
 }
 
@@ -37,7 +36,7 @@ export interface IAddAttachmentResponse {
  *
  * addAttachment({
  *   url: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/ServiceRequest/FeatureServer/0",
- *   id: 8484,
+ *   featureId: 8484,
  *   attachment: myFileInput.files[0]
  * });
  * ```
@@ -53,14 +52,11 @@ export function addAttachment(
     ...requestOptions
   };
 
-  // `attachment` and any additional parameters --> params: {}
-  appendCustomParams(requestOptions, options);
+  // `attachment` --> params: {}
+  options.params.attachment = requestOptions.attachment;
 
   // force POST
   options.httpMethod = "POST";
 
-  return request(
-    `${requestOptions.url}/${requestOptions.id}/addAttachment`,
-    options
-  );
+  return request(`${options.url}/${options.featureId}/addAttachment`, options);
 }
--- a/packages/arcgis-rest-feature-service/src/addAttachment.ts
+++ b/packages/arcgis-rest-feature-service/src/addAttachment.ts
@@ -1,0 +1,66 @@
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+
+import { IFeatureRequestOptions } from "./query";
+
+import { IEditFeatureResult, appendCustomParams } from "./helpers";
+
+/**
+ * Request options to for adding a related attachment to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
+ *
+ * @param url - Feature service url.
+ * @param id - Unique identifier of feature to add related attachment.
+ * @param attachment - File to be attached.
+ * @param params - Additional parameters to be sent via the request. See reference docs.
+ */
+export interface IAddAttachmentOptions extends IFeatureRequestOptions {
+  attachment: File;
+}
+
+/**
+ * `addAttachment()` request response.
+ */
+export interface IAddAttachmentResponse {
+  /**
+   * Standard AGS add/update/edit result Object for the attachment.
+   */
+  addAttachmentResult: IEditFeatureResult;
+}
+
+/**
+ * Attach a file to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
+ *
+ * ```js
+ * import { addAttachment } from '@esri/arcgis-rest-feature-service';
+ *
+ * addAttachment({
+ *   url: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/ServiceRequest/FeatureServer/0",
+ *   id: 8484,
+ *   attachment: myFileInput.files[0]
+ * });
+ * ```
+ *
+ * @param requestOptions - Options for the request.
+ * @returns A Promise that will resolve with the `addAttachment()` response.
+ */
+export function addAttachment(
+  requestOptions: IAddAttachmentOptions
+): Promise<IAddAttachmentResponse> {
+  const options: IAddAttachmentOptions = {
+    params: {},
+    ...requestOptions
+  };
+
+  // `attachment` and any additional parameters --> params: {}
+  appendCustomParams(requestOptions, options);
+
+  // force POST
+  options.httpMethod = "POST";
+
+  return request(
+    `${requestOptions.url}/${requestOptions.id}/addAttachment`,
+    options
+  );
+}
--- a/packages/arcgis-rest-feature-service/src/deleteAttachments.ts
+++ b/packages/arcgis-rest-feature-service/src/deleteAttachments.ts
@@ -1,0 +1,66 @@
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+
+import { IFeatureRequestOptions } from "./query";
+
+import { IEditFeatureResult, appendCustomParams } from "./helpers";
+
+/**
+ * Request options to for deleting related attachments of a feature by id. See [Delete Attachments](https://developers.arcgis.com/rest/services-reference/delete-attachments.htm) for more information.
+ *
+ * @param url - Feature service url.
+ * @param id - Unique identifier of feature to delete related attachment(s).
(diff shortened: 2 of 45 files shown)


packages/arcgis-rest-feature-service/src/addAttachment.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Request options to for adding a related attachment to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
+ *
+ * @param url - Feature service url.
+ * @param id - Unique identifier of feature to add related attachment.
+ * @param attachment - File to be attached.
+ * @param params - Additional parameters to be sent via the request. See reference docs.
+ */
+export interface IAddAttachmentOptions extends IFeatureRequestOptions {
+  attachment: File;
```
COV-GIS: @jgravois and @tomwayson I'm having an issue mocking up a File for testing. tslint wants a File b/c I've typed it here. But get a `Reference Error` when running tests. Do you have any thoughts on this. Here's the test...
```typescript
it("should return objectId of the added attachment and a truthy success", done => {
  const requestOptions = {
    url: "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0",
    id: 42,
    attachment: new File(["foo"], "foo.txt", {type: "text/plain"}) as File,
    where: "1=1"
  } as IAddAttachmentOptions;
  fetchMock.once("*", addAttachmentResponse);
  addAttachment(requestOptions).then(response => {
    expect(fetchMock.called()).toBeTruthy();
    const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
    expect(url).toEqual(`${requestOptions.url}/${requestOptions.id}/addAttachment`);
    expect(options.body).toContain("where=1%3D1");
    expect(options.method).toBe("POST");
    expect(addAttachmentResponse.addAttachmentResult.objectId).toEqual(1001);
    expect(addAttachmentResponse.addAttachmentResult.success).toEqual(true);
    done();
  });
});
```

patrickarlt: @COV-GIS It is probally because `File` doesn't exist in Node so your getting a `ReferenceError` in the Node tests. Try running the following `npm run test:chrome` this will run only the browser tests. If that passes then run `npm run test:node` if that fails then you probably need to detect if `File` exists and if it doesn't do https://stackoverflow.com/questions/44021538/how-to-send-a-file-in-request-node-fetch-or-nodejs to create a different object to use for the attachment.

COV-GIS: Thank you @patrickarlt.

COV-GIS:
Couple notes on the tests...

1. I added and exported `attachmentFile()` in feature.ts but the code coverage goes down, so moved to features.test.ts.
2. Custom params are not being added to `addAttachment()` or `updateAttachment()`. `expect(options.body).toContain("where=1%3D1");` fails when `where: 1=1` is added to request options and I can't see any reason why.

packages/arcgis-rest-feature-service/src/deleteAttachments.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { IFeatureRequestOptions } from "./query";
+
+import { IEditFeatureResult, appendCustomParams } from "./helpers";
+
+/**
+ * Request options to for deleting related attachments of a feature by id. See [Delete Attachments](https://developers.arcgis.com/rest/services-reference/delete-attachments.htm) for more information.
+ *
+ * @param url - Feature service url.
+ * @param id - Unique identifier of feature to delete related attachment(s).
```
jgravois: another nitpick, but i think it makes more sense to have the generic `id` refer to the attachments specifically for this method than the feature.

~~`id`~~ > `featureId`
~~`attachmentIds`~~ > `ids`

or we could just be _completely_ explicit.

`featureId`
`attachmentIds`

packages/arcgis-rest-feature-service/src/addAttachment.ts
```diff
@@ -0,0 +1,66 @@
...
+    ...requestOptions
+  };
+
+  // `attachment` and any additional parameters --> params: {}
+  appendCustomParams(requestOptions, options);
+
+  // force POST
+  options.httpMethod = "POST";
+
+  return request(
```
jgravois: i don't think what you're doing is an `error`, but since you already mixed in, it'd be more _consistent_ to use `options` exclusively and stop relying on `requestOptions` altogether.

packages/arcgis-rest-feature-service/src/getAttachments.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Request options to fetch `attachmentInfos` of a feature by id. See [Attachment Infos](https://developers.arcgis.com/rest/services-reference/attachment-infos-feature-service-.htm) for more information.
+ *
+ * @param url - Feature service url.
+ * @param id - Unique identifier of feature to request related `attachmentInfos`.
+ * @param params - Additional parameters to be sent via the request. See reference docs.
+ */
+// Use IFeatureRequestOptions directly.
+// export interface IGetAttachmentsOptions extends IFeatureRequestOptions {}
```
jgravois: safe to remove.

packages/arcgis-rest-feature-service/test/features.test.ts
```diff
@@ -5,19 +5,43 @@ import {
...
   updateFeaturesResponse,
-  deleteFeaturesResponse
+  deleteFeaturesResponse,
+  getAttachmentsResponse,
+  addAttachmentResponse,
+  updateAttachmentResponse,
+  deleteAttachmentsResponse
 } from "./mocks/feature";
 
+function attachmentFile() {
```
jgravois: nice work!

packages/arcgis-rest-feature-service/test/features.test.ts
```diff
@@ -173,4 +197,103 @@ describe("feature", () => {
       done();
     });
   });
+
+  it("should return an array of attachmentInfos for a feature by id", done => {
+    const requestOptions = {
+      url:
+        "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0",
+      id: 42,
+      where: "1=1"
```
jgravois: no need to include this parameter, it is the default.

packages/arcgis-rest-feature-service/test/features.test.ts
```diff
@@ -173,4 +197,103 @@ describe("feature", () => {
       done();
     });
   });
+
+  it("should return an array of attachmentInfos for a feature by id", done => {
+    const requestOptions = {
+      url:
+        "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0",
```
jgravois: it'd be good to use a visibly fake url here.

packages/arcgis-rest-feature-service/test/features.test.ts
```diff
@@ -173,4 +197,103 @@ describe("feature", () => {
...
+      expect(options.method).toBe("POST");
+      expect(
+        deleteAttachmentsResponse.deleteAttachmentResults[0].objectId
+      ).toEqual(1001);
+      expect(
+        deleteAttachmentsResponse.deleteAttachmentResults[0].success
+      ).toEqual(true);
+      done();
+    });
+  });
```
jgravois: can you add one more test to trap for the scenario in which a developer tries to add an attachment to a feature service that doesn't support the capability please?

packages/arcgis-rest-feature-service/test/features.test.ts
jgravois: unless i'm missing something, the underlying operation for `getAttachments()` doesn't support a `where` parameter.

packages/arcgis-rest-feature-service/src/addAttachment.ts
```diff
@@ -0,0 +1,66 @@
...
+export function addAttachment(
+  requestOptions: IAddAttachmentOptions
+): Promise<IAddAttachmentResponse> {
+  const options: IAddAttachmentOptions = {
+    params: {},
+    ...requestOptions
+  };
+
+  // `attachment` and any additional parameters --> params: {}
+  appendCustomParams(requestOptions, options);
```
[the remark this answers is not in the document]
jgravois: the reason i used `appendCustomParams()`in the crud operations for feature service is because the interfaces they extend ([`ISharedQueryParams`](https://github.com/Esri/arcgis-rest-js/blob/c0a881bd028eb189ca27fa666ae1089663a563c1/packages/arcgis-rest-feature-service/src/helpers.ts#L16-L23) and [`IEditFeaturesParams`](https://github.com/Esri/arcgis-rest-js/blob/c0a881bd028eb189ca27fa666ae1089663a563c1/packages/arcgis-rest-feature-service/src/helpers.ts#L37-L50)) include quite a few optional parameters that are a little too tedious to test for and append manually.

at this point i'm actually probably more fond of the approach you've taken here, primarily _because_ they make calling `appendCustomParams()` superfluous.

COV-GIS:
@jgravois I'll get to this hopefully soon.

packages/arcgis-rest-feature-service/src/addAttachment.ts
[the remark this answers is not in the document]
COV-GIS: Hey @jgravois . `attachment` is not the only param for the operation. But they are few. Per your comment below about being explicit, why not a just extend `IRequestOptions` for each of the attachment methods with the REST API's params. I suppose I could have an interface with params like `gdbVersion`, but I like the simplicity of not extending several interfaces with these straight forward server operations. Honestly, I find the reference hard to follow when it's 3-4 interfaces deep. Based on this project's description it should be the thinnest of wrappers for the REST API without piling on. Just my two cents.

Let me know what you think either way, and I'll get it fixed up.

```typescript
export interface IAddAttachmentOptions extends IRequestOptions {
  url: string;
  featureId: number;
  attachment: File;
  gdbVersion?: string;
  returnEditMoment?: boolean;
  uploadId?: number;
}

export function addAttachment(
  requestOptions: IAddAttachmentOptions
): Promise<IAddAttachmentResponse> {
  const options: IAddAttachmentOptions = {
    params: {},
    ...requestOptions
  };

  // `attachment` and any additional parameters --> params: {}
  options.params.attachment = requestOptions.attachment;
  options.params.gdbVersion = requestOptions.gdbVersion || null;
  options.params.returnEditMoment = requestOptions.returnEditMoment === true ? true : false;
  options.params.uploadId = requestOptions.uploadId || null;

  // force POST
  options.httpMethod = "POST";

  return request(
    `${requestOptions.url}/${requestOptions.featureId}/addAttachment`,
    options
  );
}
```

packages/arcgis-rest-feature-service/src/addAttachment.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Attach a file to a feature by id. See [Add Attachment](https://developers.arcgis.com/rest/services-reference/add-attachment.htm) for more information.
+ *
+ * ```js
+ * import { addAttachment } from '@esri/arcgis-rest-feature-service';
+ *
+ * addAttachment({
+ *   url: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/ServiceRequest/FeatureServer/0",
+ *   id: 8484,
+ *   attachment: myFileInput.files[0]
```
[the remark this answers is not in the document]
COV-GIS: Yes. I've started an attachment demo.

packages/arcgis-rest-feature-service/src/addAttachment.ts
COV-GIS: What does that pattern look like generally?

packages/arcgis-rest-feature-service/test/features.test.ts
COV-GIS: It does not. But was just testing whether addition params made it through. The crud methods don't either but `where` is passed and tested for.

packages/arcgis-rest-feature-service/src/addAttachment.ts
jgravois: i just meant this:
```js
const options: IAnyAttachmentOptions = {
  params: {},
  ...requestOptions
};

return request(
  `${options.url}/${options.id}/any`, options // instead of requestOptions
);
```

packages/arcgis-rest-feature-service/src/addAttachment.ts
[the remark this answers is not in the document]
COV-GIS: Does the sun rise in the east?

I'll get it all fixed up.

packages/arcgis-rest-feature-service/test/features.test.ts
COV-GIS: Missed this. Just did.

packages/arcgis-rest-feature-service/test/features.test.ts
```diff
@@ -173,4 +198,115 @@ describe("feature", () => {
...
+    } as IAddAttachmentOptions;
+    fetchMock.once("*", addAttachmentResponse);
+    addAttachment(requestOptions).then(response => {
+      expect(fetchMock.called()).toBeTruthy();
+      const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
+      expect(url).toEqual(
+        `${requestOptions.url}/${requestOptions.featureId}/addAttachment`
+      );
+      expect(options.method).toBe("POST");
+      // expect(options.body).toContain("returnEditMoment=true");
```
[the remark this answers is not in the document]
jgravois: this is because the request parameters are passed as `FormData`. the approach below works in browsers, but throws an error in Node.js

ref: https://github.com/form-data/form-data/issues/124

```ts
expect(options.body instanceof FormData).toBeTruthy();
const params = options.body as FormData;
expect(params.get("returnEditMoment")).toEqual("true");
```

all in all, this PR looks great. i haven't checked out your demo yet, but i've pushed a commit of my own to tidy up a few more things i've noticed so far.

COV-GIS:
That all looks good @jgravois. I added polyfills to demo. And hopefully we can get this wrapped up.

COV-GIS:
Thanks @jgravois. Can't wait to put this to work!

I can only foresee adding related records query. I've learned a lot...so hope to hit that one out of the park with a lower commit and comment count.

jgravois:
> a lower commit and comment count.

dialogue and revision means the system is working, not failing 😄.

i'll release as soon as i have a chance to plow through a few other pending PRs. i doubt it'll take more than a week.

Issue #239, opened 2018-07-10
Add support for the "social" authorize to beginOAuth2()


tomwayson:
In the Hub, and soon to be in Hub ready apps, we leverage the seemingly undocumented "social" authorize endpoint to go around the default login screen send users directly to the facebook/google provider sign in/up pages. For example, we first show this modal:

<img width="657" alt="image" src="https://user-images.githubusercontent.com/662944/42490692-820e59c8-83c6-11e8-8f62-72e38cc2dded.png">

Clicking on the Facebook/Google buttons will open an OAuth popup that goes to `/sharing/rest/oauth2/social/authorize?socialLoginProviderName=<google|facebook>&autoAccountCreateForSocial=true&width=800&height=500` along w/ the rest of the [OAuth2Options](https://esri.github.io/arcgis-rest-js/api/auth/IOauth2Options/) like `clientId`  and `redirectUri`. That route in turn redirects to the provider's own page:

<img width="1033" alt="image" src="https://user-images.githubusercontent.com/662944/42490888-458ebb5e-83c7-11e8-90e6-81081c662605.png">

I _think_ once the user does their business over on the provider pages the rest of the flow is the same (the redirect uri is opened with `token`, `state`, `expires_in`, `username` etc).

I'm hoping this can be done fairly easily by to to extend [OAuth2Options](https://esri.github.io/arcgis-rest-js/api/auth/IOauth2Options/) to have a `path` which defaults to `/rest/oauth2/authorize` but in this case would be overridden by `/rest/oauth2/social/authorize` along w/ additional props for `socialLoginProviderName` and `autoAccountCreateForSocial` (and maybe `width` and `height`?).

The relevant code that the Hub uses can be seen here:

https://github.com/ArcGIS/opendata-ui/blob/cfbcc8541de10ca67311375a59745e12c1cb5d3f/packages/opendata-ui/app/components/od-sign-in-modal/component.js#L62-L85

And then here:

https://github.com/Esri/torii-provider-arcgis/blob/8be3bdfe99647d1ac87f7b813553fb5968033b14/app/torii-providers/arcgis-oauth-bearer.js#L65-L162

Who's with me? ⚔️

tomwayson:
I can confirm the above approach works by changing the following line n auth.umd.js in my node_modules:

```js
[...]

PR #240, opened 2018-07-10
feat(:unlock:): add support for an OAuth flow that triggers social me…

jgravois:
…dia login automatically

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #239

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -22,6 +22,8 @@ interface IDeferred<T> {
   reject: (v: any) => void;
 }
 
+export type IAuthenticationProvider = "arcgis" | "facebook" | "google";
+
 /**
  * Represents a [credential]((https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-Credential.html)) object used to access a secure ArcGIS resource.
  */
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -156,6 +163,11 @@ export interface IUserSessionOptions {
    */
   portal?: string;
 
+  /**
+   * The authentication provider to use. Defaults to ArcGIS, but Google and Facebook social media logins can also be used.
```
tomwayson: we may want to say something here like: "Specifying a non-ArcGIS provider will take the user directly to that provider's OAuth page."

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -309,17 +329,24 @@ export class UserSession implements IAuthenticationManager {
...
-    const url = `${portal}/oauth2/authorize?client_id=${clientId}&response_type=token&expiration=${duration}&redirect_uri=${encodeURIComponent(
-      redirectUri
-    )}&state=${state}&locale=${locale}`;
+    let url: string;
+    if (provider === "arcgis") {
+      url = `${portal}/oauth2/authorize?client_id=${clientId}&response_type=token&expiration=${duration}&redirect_uri=${encodeURIComponent(
+        redirectUri
+      )}&state=${state}&locale=${locale}`;
+    } else {
+      url = `${portal}/oauth2/social/authorize?client_id=${clientId}&socialLoginProviderName=${provider}&autoAccountCreateForSocial=true&response_type=token&expiration=${duration}&redirect_uri=${encodeURIComponent(
```
tomwayson: This works for Hub, so you can ignore https://github.com/Esri/arcgis-rest-js/issues/239#issuecomment-403905821 - we can add an option for this later if others don't want this behavior.

tomwayson: now I'm leaning back the other way: https://github.com/Esri/arcgis-rest-js/issues/239#issuecomment-403914891

patrickarlt: My comments https://github.com/Esri/arcgis-rest-js/issues/239#issuecomment-403933222

tomwayson:
LGTM :shipit:!

Issue #242, opened 2018-07-11
#233 breaks the use case where a user is signed in and trying to access a public feature service


tomwayson:
My app queries a public feature https://services.arcgis.com/bkrWlSKcjUDFDtgw/arcgis/rest/services/Hub%20Annotations/FeatureServer/0 and uses this as the portal: https://cityxcommunity.maps.arcgis.com/sharing/rest

It worked up until v1.4.1 (and I'd guess right up to #233).

Since upgrading to 1.5.0 I get 2 errors.

The first is that 

That obscures the following NOT_FEDERATED error 

![image](https://user-images.githubusercontent.com/662944/42584924-27aa75b0-84e9-11e8-9b8e-76d98c6a3991.png)

Which is related to

patrickarlt:
Since you aren't using the default portal it is likely failing this check https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L628-L629 then doing `getTokenForServer()` https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L678-L744 which will make a call to https://services.arcgis.com/bkrWlSKcjUDFDtgw/arcgis/rest/info?f=pjson to get `authInfo.tokenServicesUrl` which doesn't match your portal which means that this check is failing https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L701-L704.

So we have 2 options:

1. Trust org portals on ArcGIS Online like `https://cityxcommunity.maps.arcgis.com/sharing/rest` like we do the default ArcGIS Online portal `https://www.arcgis.com/sharing/rest` and fix https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L628-L629
2. Fix the check on https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L701-L704 to assume that if our portal is on an arcgis.com domain and our `authInfo.tokenServicesUrl` is on arcgis.com domain, then assume they are federated and continue trying to generate a token.

patrickarlt:
@tomwayson also is there ANY reason to use https://cityxcommunity.maps.arcgis.com/sharing/rest over https://www.arcgis.com/sharing/rest? As far as I have been able to tell when using the REST APIs there isn't really a difference.

[...]

PR #245, opened 2018-07-16
fix(:bug:): treat ArcGIS Online orgs as equivalent to www.arcgis.com

jgravois:
when testing portals

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #233

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -625,7 +625,9 @@ export class UserSession implements IAuthenticationManager {
    */
   getToken(url: string) {
     if (
-      /^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal) &&
+      /^https?:\/\/\S+\.maps.arcgis\.com\/sharing\/rest|^https?:\/\/www\.arcgis\.com\/sharing\/rest/.test(
+        this.portal
+      ) &&
       /^https?:\/\/\S+\.arcgis\.com.+/.test(url)
     ) {
       return this.getFreshToken();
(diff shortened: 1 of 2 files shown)


tomwayson:
this actually closes #242, right?

jgravois:
whoops. yes.

i just followed the link in the title instead of looking at the url.

jgravois:
>  I think we may need to tweak the regexes

i _think_ we'd be better served to relegate QA and Dev as custom portals.

tomwayson:
or at the very least defer that to a later PR

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -625,7 +625,9 @@ export class UserSession implements IAuthenticationManager {
    */
   getToken(url: string) {
     if (
-      /^https?:\/\/www\.arcgis\.com\/sharing\/rest\/?/.test(this.portal) &&
+      /^https?:\/\/\S+\.maps.arcgis\.com\/sharing\/rest|^https?:\/\/www\.arcgis\.com\/sharing\/rest/.test(
```
patrickarlt: @jgravois @tomwayson I think we should change this regex to be `https:?\/\/S+\.arcgis.com\/sharing\/rest` which would cover both use cases in 1 Regex AND cover Dev and QA. https://regexr.com/3sils.

I think we SHOULD treat dev and QA the same as we treat production otherwise our applications might have unexpected behavior in those environments that we cant reproduct on production.

jf990:
@patrickarlt you should also escape the second ".", dot matches any char, so this is valid

https://example.gis.arcgisxcom/sharing/rest

jgravois:
i've pushed up another commit which incorporates @patrickarlt (and @jf990's) _excellent_ feedback.

Issue #247, opened 2018-07-17
Inconsistent handling of owner property in item functions


noahmulfinger:
A lot of the functions in `items.ts` use different methods for getting the owner property used for creating the correct url. It would be great if we could standardize this as much as possible. Some examples below:

In the `updateItem` function:
```ts
const url = `${getPortalUrl(requestOptions)}/content/users/${
    requestOptions.item.owner
}/items/${requestOptions.item.id}/update`;
```

In the `updateItemResources` function:
```ts
const owner = requestOptions.owner || requestOptions.authentication.username;
const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
    requestOptions.id
}/updateResources`;
```

In the `createItemInFolder` function:
```ts
const owner =
    requestOptions.owner ||
    requestOptions.item.owner ||
    requestOptions.authentication.username;
const baseUrl = `${getPortalUrl(requestOptions)}/content/users/${owner}`;
let url = `${baseUrl}/addItem`;
```

jgravois:
resolved in #248

Issue #241, opened 2018-07-11
Group interfaces don't match documentation


alukach:
Taking a look at our `IGroup` interface, it doesn't seem to match [the documentation](https://developers.arcgis.com/rest/users-groups-and-items/common-parameters.htm#ESRI_SECTION1_17DCA22A3B4F4623960631498150A28C) at all:
https://github.com/Esri/arcgis-rest-js/blob/2aa15e8b7457940bc7b2fef1543f6ba3a5c095c0/packages/arcgis-rest-groups/src/groups.ts#L20-L29

For example, it shows `owner` as a required parameter however the docs do not show it as a supported parameter and states:

> The user who creates the group automatically becomes the owner of the group.

Furthermore, parameters like `culture` and `categories` are not related to the `Group` objects at all and appear to be likely copied over from the parameters for creating `Item` instances.

My understanding is that the interfaces should be updated to something similar to the following:

```typescript

export interface IGroupCreateRequest {
  title: string;
  access: 'private' | 'org' | 'public';
  description?: string;
  snippet?: string;
  tags?: string;
  phone?: string;
  sortField?: 'title' | 'owner' | 'avgrating' | 'numviews' | 'created' | 'modified';
  sortOrder?: 'asc' | 'desc';
  isViewOnly?: boolean;
  isInvitationOnly?: boolean;
  thumbnail?: string;
  autoJoin?: boolean;
}

// https://developers.arcgis.com/rest/users-groups-and-items/group.htm
export interface IGroup {
  title: string;
  id: string;
  isInvitationOnly: boolean;
  owner: string;
  description?: string;
  snippet?: string;
  tags: string[];
  phone?: string;
  sortField?: 'title' | 'owner' | 'avgrating' | 'numviews' | 'created' | 'modified';
  sortOrder?: 'asc' | 'desc';
  isViewOnly: boolean;
  isFav: boolean;
  thumbnail?: string;
  created: number;
  modified: number;
  access: 'private' | 'org' | 'public';
  protected: boolean;
  autoJoin: boolean;
  // Server responded with the following undocumented attributes:
  // "isReadOnly": false,
  // "capabilities": [],
  // "provider": null,
  // "providerGroupName": null,
}

export interface IGroupCreateResponse {
  success: boolean;
  group: IGroup;
}
```

Having separate interfaces for the Request and Response parameters would make development go much more smoothly.

jgravois:
thanks for taking the time to write up this feedback @alukach!

[...]

PR #248, opened 2018-07-17
constrain what constitutes a valid Item for crud operations, and require auth

### LINKED ISSUE #246: Doc: IItem shows "type" & "title" as optional, but AGOL requires them for createItem calls
```
  arcgisRest.createItem({
    authentication: userSession,
    item: {
    }
  })
```
leads to
image

```
  arcgisRest.createItem({
    authentication: userSession,
    item: {
      title: "copied zip",
      type: "Code Sample"
    }
  })
```
is accepted by AGOL.

jgravois:
resolves #246:
* adds a couple code snippets
* ~~stops using the (unnecessary) interface `IUserRequestOptions`~~
* `createItem()` now requires an Item with 'title' and 'type'
* `updateItem()` now requires an Item with an 'id'
* ~~the methods above, and `removeItem()`, now _all_ require authentication~~

no one loves interface soup more than @dbouwman. hopefully he can find time to review soon as @tomwayson is out until late next week.

### DIFF
--- a/packages/arcgis-rest-common-types/src/group.ts
+++ b/packages/arcgis-rest-common-types/src/group.ts
@@ -1,0 +1,51 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+/**
+ * A [Group](https://developers.arcgis.com/rest/users-groups-and-items/common-parameters.htm) that has not been created yet.
+ */
+export interface IGroupAdd {
+  title: string;
+  owner?: string;
+  tags?: string[];
+  description?: string;
+  access?: "private" | "org" | "public";
+  phone?: string;
+  sortField?:
+    | "title"
+    | "owner"
+    | "avgrating"
+    | "numviews"
+    | "created"
+    | "modified";
+  sortOrder?: "asc" | "desc";
+  isViewOnly?: boolean;
+  isInvitationOnly?: boolean;
+  thumbnail?: string;
+  autoJoin?: boolean;
+  snippet?: string;
+  [key: string]: any;
+}
+
+/**
+ * Existing Portal [Group](https://developers.arcgis.com/rest/users-groups-and-items/group.htm).
+ */
+export interface IGroup extends IGroupAdd {
+  id: string;
+  owner: string;
+  tags: string[];
+  created: number;
+  modified: number;
+  protected: boolean;
+  isInvitationOnly: boolean;
+  isViewOnly: boolean;
+  isOpenData: boolean;
+  isFav: boolean;
+  autoJoin: boolean;
+  userMembership?: {
+    username?: string;
+    memberType?: string;
+    applications?: number;
+  };
+  hasCategorySchema?: boolean;
+}
--- a/packages/arcgis-rest-common-types/src/item.ts
+++ b/packages/arcgis-rest-common-types/src/item.ts
@@ -1,0 +1,45 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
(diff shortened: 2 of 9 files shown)


noahmulfinger:
@jgravois Nice! This also mostly addresses the issue I just put up https://github.com/Esri/arcgis-rest-js/issues/247. What about making `determineOwner` more generic so that if there is a `requestOptions.item.owner`, that will be chosen before `session.username`? That way it could be used for `createItemInFolder` and `updateItem` as well.

jgravois:
> This also mostly addresses #247

i love it when a plan comes together. i've pushed up another commit with your suggestion.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -398,3 +457,16 @@ function serializeItem(item: IItem): any {
   }
   return clone;
 }
+
+/**
+ * requestOptions.owner is given priority, requestOptions.item.owner will be checked next. If neither are present, authentication.username will be assumed.
+ */
+function determineOwner(requestOptions: any): string {
```
jgravois: i'm sure it'd be better to use a [union type](https://www.typescriptlang.org/docs/handbook/advanced-types.html) here instead of casting to `any`, but i was worried that i'd already reached my daily limit of new interfaces for this package today.

noahmulfinger: Yeah, I think using `any` is fine in this case for a simple utility function.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -48,8 +57,7 @@ export interface IItemResourceRequestOptions extends IItemIdRequestOptions {
   resource?: string;
 }
 
-export interface IItemCrudRequestOptions extends IUserRequestOptions {
-  item: IItem;
+export interface IItemCrudRequestOptions extends IRequestOptions {
```
noahmulfinger: @jgravois Why remove `IUserRequestOptions`? That already has a mandatory `authentication` property.

jgravois:
it ended up being easier to address #241 and this issue holistically.

its a more involved refactor, but the approach i was suggesting before just feels like a 🤕.

noahmulfinger:
@jgravois Taking a look now.

packages/arcgis-rest-common-types/src/group.ts
```diff
@@ -0,0 +1,51 @@
...
+  owner: string;
+  tags: string[];
+  created: number;
+  modified: number;
+  protected: boolean;
+  isInvitationOnly: boolean;
+  isViewOnly: boolean;
+  isOpenData: boolean;
+  isFav: boolean;
+  snippet?: string;
```
noahmulfinger: `snippet` could probably be moved to IGroupAdd since it fits in with properties that can be passed in to create a group.

packages/arcgis-rest-common-types/src/item.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Existing Portal [Item](https://developers.arcgis.com/rest/users-groups-and-items/item.htm).
+ */
+export interface IItem extends IItemAdd {
+  id: string;
+  owner: string;
+  tags: string[];
+  created: number;
+  modified: number;
+  protected: boolean;
+}
```
noahmulfinger: What's the reason for requiring these specific properties for `IItem`?

jgravois: i've written the interface this way to acknowledge that these props are all guaranteed to exist (_after_ an item has been created).

jgravois:
final feedback addressed (and followed by a rebase). thank you @noahmulfinger!

PR #249, opened 2018-07-18
differentiate IGroup and IGroupResponse

jgravois:
what do you think of this @alukach?

something like `readonly isOpenData: boolean;` would have been less complicated, but that only prohibits a property from being altered _after_ the object has been instantiated.
 
AFFECTS PACKAGES:
@esri/arcgis-rest-groups

~~ISSUES CLOSED: #241~~

### DIFF
--- a/packages/arcgis-rest-common-types/src/index.ts
+++ b/packages/arcgis-rest-common-types/src/index.ts
@@ -477,8 +477,8 @@ export interface IUser {
 }
 
 export interface IGroup extends IItem {
-  isInvitationOnly?: boolean;
   phone?: string;
+  access?: "private" | "org" | "public";
   sortField?:
     | "title"
     | "owner"
@@ -486,19 +486,11 @@ export interface IGroup extends IItem {
     | "numviews"
     | "created"
     | "modified";
+  sortOrder?: "asc" | "desc";
   isViewOnly?: boolean;
-  isFav?: boolean;
-  access?: "private" | "org" | "public";
-  userMembership?: {
-    username?: string;
-    memberType?: string;
-    applications?: number;
-  };
-  protected?: boolean;
+  isInvitationOnly?: boolean;
+  thumbnail?: "string";
   autoJoin?: boolean;
-  hasCategorySchema?: boolean;
-  isOpenData?: boolean;
-  [key: string]: any;
 }
 
 export type esriUnits =
--- a/packages/arcgis-rest-groups/src/groups.ts
+++ b/packages/arcgis-rest-groups/src/groups.ts
@@ -1,9 +1,9 @@
-/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+/* Copyright (c) 2017-2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
+
 import {
   request,
   IRequestOptions,
-  // IParams,
   getPortalUrl
 } from "@esri/arcgis-rest-request";
 
@@ -28,6 +28,18 @@ export interface IGroupSearchRequest extends IPagingParams {
   [key: string]: any;
 }
 
+export interface IGroupResponse extends IGroup {
+  isFav?: boolean;
+  userMembership?: {
+    username?: string;
+    memberType?: string;
+    applications?: number;
+  };
+  protected?: boolean;
+  hasCategorySchema?: boolean;
+  isOpenData?: boolean;
+}
+
 export interface IGroupSearchResult {
   /**
    * Matches the REST APIs form param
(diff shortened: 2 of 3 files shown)


alukach:
@jgravois Wow, thanks for taking the time to further expand on my suggestion!

> something like `readonly isOpenData: boolean;` would have been less complicated, but that only prohibits a property from being altered after the object has been instantiated.

That's exactly where I started when thinking about how to solve my issue, however I agree that it's not ideal.

Overall, I think this change is a step in the right direction, however I think it could be refined just a bit more.

My understanding of REST is that the standard pattern is to have the API respond with the resource, so with I would argue that the `IGroupResponse` model _is_ the `IGroup`.  If we were to separate the request and response actions, it should be the request object that is the exception, as it is a subset of the full model. I believe this would better allow us to have an interface that represents what the REST server expects/allows (for example, this current code shows us optionally allowing an `id` attribute, which would almost surely be ignored by the server).

To pull from #241, I think the following would exactly mimic the permitted/expected behaviour of the REST service:

```js
export interface IGroupCreateRequest {
  title: string;
  access: 'private' | 'org' | 'public';
  description?: string;
  snippet?: string;
  tags?: string;
  phone?: string;
  sortField?: 'title' | 'owner' | 'avgrating' | 'numviews' | 'created' | 'modified';
  sortOrder?: 'asc' | 'desc';
  isViewOnly?: boolean;
  isInvitationOnly?: boolean;
  thumbnail?: string;
  autoJoin?: boolean;
}

// https://developers.arcgis.com/rest/users-groups-and-items/group.htm
export interface IGroup {
  title: string;
  id: string;
  isInvitationOnly: boolean;
  owner: string;
  description?: string;
  snippet?: string;
  tags: string[];
  phone?: string;
  sortField?: 'title' | 'owner' | 'avgrating' | 'numviews' | 'created' | 'modified';
  sortOrder?: 'asc' | 'desc';
  isViewOnly: boolean;
  isFav: boolean;
  thumbnail?: string;
  created: number;
  modified: number;
  access: 'private' | 'org' | 'public';
  protected: boolean;
  autoJoin: boolean;
}

export interface IGroupCreateResponse {
  success: boolean;
  group: IGroup;
}
```

One significant difference between the above code and the code in this PR is that in the above code, `IGroup` doesn't extend `IItem`.  I like the idea of extending `IItem`, but I think that `IItem` is a bit flawed in that it allows `id` (and other expected attributes) to be optional. I think that the idea of separating the request interface from the response interface could extend to the majority of this library. Implementing this would likely require a significant refactor.  Naturally, this could be done incrementally (updating each resource as needed, removing the extension from `IItem`, and then when all interfaces are updated and nothing extends from `IItem`, we would update `IItem` to properly reflect the expected attributes). Or, it could be done all at once in a major refactor 💥. It's a question of priorities.

I might be acting a bit overly pedantic with all of this stuff.  I can help out some, but I'll leave it to you and the other contributors to decide if such an overhaul of the typings is actually a worthy use your time.  Overall, what's proposed in this PR is better than before so I'd be happy with even this if priorities don't allow for things to be refined further.

jgravois:
> I might be acting a bit overly pedantic with all of this stuff.

i disagree. i think this is helpful insight that will improve our API design.

> I think that `IItem` is a bit flawed in that it allows id (and other expected attributes) to be optional (...) The convenience of extending interfaces is nice, but shouldn't come at the cost of correctness.

we actually already discussed this without coming to a really satisfying conclusion (see #184, #246).

thinking about it more i'm convinced that we can (and should) have the best of both worlds.

```ts

export interface IItemCreate {
  title: string;
  access: 'private' | 'org' | 'public';
  description?: string;
  snippet?: string;
  tags?: string;
}

export interface IItem extends IItemCreate {
  id: string;
  owner: string;
  tags: string[]; // optional in the parent interface
  created: number;
  modified: number;
  protected: boolean;
}

// Group Parameters from https://developers.arcgis.com/rest/users-groups-and-items/common-parameters.htm 
export interface IGroupCreate extends IItemCreate {
  phone?: string;
  sortField?: 'title' | 'owner' | 'avgrating' | 'numviews' | 'created' | 'modified';
  sortOrder?: 'asc' | 'desc';
  isViewOnly?: boolean;
  isInvitationOnly?: boolean;
  thumbnail?: string;
  autoJoin?: boolean;
}

// https://developers.arcgis.com/rest/users-groups-and-items/group.htm
export interface IGroup extends IGroupCreate, IItem {
  isInvitationOnly: boolean;
  owner: string;
  tags: string[];
  isViewOnly: boolean;
  isFav: boolean;
  snippet?: string;
  autoJoin: boolean;
}
```

jgravois:
i spent some more time on this and i have to say, i'm coming around to your point of view.

its not that you _can't_ achieve correctness with `extend`, its just that it introduces more unnecessary complexity than benefit.

my latest stab at reorganizing can be found in b4c5f93. more feedback is always welcome, but i'm going to go ahead and close _this_ PR and continue the discussion in #248.

PR #253, opened 2018-07-23
Add attachment notes for api docs

COV-GIS:
@jgravois I noticed the the Notes in the API docs come from the comments in the interfaces themselves and not from `@param`. Here they be.


COV-GIS:
Just checking @jgravois... Are the `@param`s what TS uses for autofill and descriptions when deving?

PR #254, opened 2018-07-23
Add support for uploading blob item's data

MikeTschudi:
For reasonable browsers, a Blob can be converted into a File and sent up with its filename. For IE 11 & Edge, the File constructor is not available, and even if one adds the missing File properties to the Blob, it's sent up to AGOL as "blob" by these two browsers. Changing encode-form-data to use the filename in the FormData.append call gets past this behavior.

I changed the name of IItemJsonDataRequestOptions from a documentation perspective since it can be used by both the JSON and binary functions.

Working on test code.

### DIFF
--- a/packages/arcgis-rest-items/src/items.ts
+++ b/packages/arcgis-rest-items/src/items.ts
@@ -30,9 +30,9 @@ export interface IItemIdRequestOptions extends IUserRequestOptions {
   owner?: string;
 }
 
-export interface IItemJsonDataRequestOptions extends IItemIdRequestOptions {
+export interface IItemJsonRequestOptions extends IItemIdRequestOptions {
   /**
-   * JSON object to store
+   * Object to store
    */
   data: any;
 }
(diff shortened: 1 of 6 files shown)


MikeTschudi:
@jgravois  thank you for the review and improvements. Travis date puzzled me; switched to constructing date from milliseconds to see if that will work both with npm test and Travis.

MikeTschudi:
@jgravois would you please review the last two commits: changing the Date construction because of the Travis problem and adding a test for encode-form-data?

jgravois:
> would you please review the last two commits

looks great! i'm going to _try_ to merge/release today.

packages/arcgis-rest-items/src/items.ts
```diff
@@ -30,9 +30,9 @@ export interface IItemIdRequestOptions extends IUserRequestOptions {
   owner?: string;
 }
 
-export interface IItemJsonDataRequestOptions extends IItemIdRequestOptions {
+export interface IItemDataOptions extends IItemIdRequestOptions {
```
tomwayson: I would suggest naming this `IItemDataRequestOptions`

jgravois: agreed. i introduced that typo during code review.

i'm going to rebase this branch to clean up the commit history and then merge.

jgravois: actually, just realized i already used `IItemDataRequestOptions` in #251.

my best suggestion for this one is `IItemDataAddRequestOptions`.

tomwayson: Not to be _that guy_, but `IItemAddDataRequestOptions`

jgravois:
🚀

PR #256, opened 2018-07-25
fix(query features): add count and extent to IQueryFeaturesResponse

noahmulfinger:
AFFECTS PACKAGES:

`count` and `extent` properties can be returned depending on input params. See examples here https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm

I didn't include this in this PR, but technically `features` should also be optional in the response type. I didn't want to edit `IFeatureSet` since it makes sense that `features` would be required there, so I just left it.

### DIFF
--- a/demos/attachments/package-lock.json
+++ b/demos/attachments/package-lock.json
@@ -2,22 +2,6 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@esri/arcgis-rest-feature-service": {
-      "version": "1.5.1",
-      "resolved": "https://registry.npmjs.org/@esri/arcgis-rest-feature-service/-/arcgis-rest-feature-service-1.5.1.tgz",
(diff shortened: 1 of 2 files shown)


noahmulfinger:
@jgravois I like it. The only issue is it looks like `count` and `extent` can both be returned at the same time. See `JSON response syntax (when returnCountOnly=true and returnExtentOnly=true) ` https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm. We could add some intersection types like `IQueryCountResponse & IQueryExtentResponse`, but we might need to include the `returnIdsOnly` case as well...

jgravois:
good point.

lets start with one flexible, alternate response interface for the different combinations of `count`, `extent` and `objectIds` that might be present in responses that _don't_ include `features[]`.

noahmulfinger:
@jgravois :+1: Makes sense to me!

Issue #203, opened 2018-05-22
supplying my own fetch implementation doesn't work at node


haoliangyu:
The documentation does not say I cannot provide my own `fetch` implementation at node. So I try

``` javascript
require("isomorphic-form-data");

const fetch = require('node-fetch')
const { request } = require("@esri/arcgis-rest-request");

request("https://www.arcgis.com/sharing/rest/info", { fetch }).then(response => {
  console.log(response);
});
```

then it throws an error

```
/home/linux/Desktop/Untitled Folder/node_modules/@esri/arcgis-rest-request/dist/node/request.js:49
    var options = tslib_1.__assign({ httpMethod: "POST", fetch: fetch }, requestOptions);
                                                                ^

ReferenceError: fetch is not defined
    at request (/home/linux/test/node_modules/@esri/arcgis-rest-request/dist/node/request.js:49:65)
    at Object.<anonymous> (/home/linux/test/index.js:7:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
    at bootstrap_node.js:609:3
```

My node version is v8.11.1 and it does not seem to like the undeclared variable `fetch`.

jgravois:
i can repro this. the only way i could use this lib w/ `node-fetch` successfully was via the global: 

```js
"use strict";
/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
 * Apache-2.0 */
Object.defineProperty(exports, "__esModule", { value: true });
// ...
var fetch = require('node-fetch');
var Headers = fetch.Headers;  // this isn't necessary with `isomorphic-fetch`
```

jgravois:
addressed (at least partially) by #263

jgravois:
addressed (fully) by #264

PR #263, opened 2018-08-01
refactor: use JSON object literal to construct Headers

jgravois:
i have no idea _why_, but `ember-fetch` ignores the real [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) we currently pass through. this approach is also necessary to support the use case described in #203.

i can't see any downside to the proposed change, but i'm happy to be proven wrong.

AFFECTS PACKAGES:
@esri/arcgis-rest-request

### DIFF
--- a/demos/batch-geocoder-node/batch-geocode.js
+++ b/demos/batch-geocoder-node/batch-geocode.js
@@ -1,4 +1,6 @@
 require('isomorphic-fetch');
+// const customFetch = require('node-fetch');
+
 require('isomorphic-form-data');
 const fs = require('fs');
 const Papa = require('papaparse');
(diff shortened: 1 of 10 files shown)


tomwayson:
FYI - w/ as part of testing #264 I pulled the `custom-fetch-node` branch, which includes the changes in this PR, and then ran basically all the demos (couldn't get the vue demo to work, issue forthcoming) _w/o making any changes to the demo code_ and they all seemed to work. In other words, I tested the baseline condition to verify that there were no regressions (and I didn't see any) _before_ I started testing node-fetch and seeing these errors: https://github.com/Esri/arcgis-rest-js/pull/264#pullrequestreview-143642171, so I think _this_ change is 👍 even if there's more work to do in #264.

PR #272, opened 2018-08-08
bump common-types version number manually

jgravois:


jgravois:
not really.

my first attempt to publish the packages on npm failed though (because I recently set up 2FA). that _could_ have had something to do with it.

PR #274, opened 2018-08-09
fix(encodeFormData): append file name based on object type instead of…

noahmulfinger:
… key and name properties

Append a filename to parameters of type File that don't have the key "file". Also append a filename
to blobs (which don't have a "name" property).

My use case for this is uploading dataURLs as Blobs for item and user thumbnails, which use the `thumbnail` property. Currently passing `thumbnail: <File or Blob>` results in the thumbnail being named `blob.png`. This PR would change it to be named  `<filename>.png` (for Files) or `thumbnail.png` (for Blobs).

AFFECTS PACKAGES:
@esri/arcgis-rest-request

### DIFF
--- a/packages/arcgis-rest-request/src/utils/encode-form-data.ts
+++ b/packages/arcgis-rest-request/src/utils/encode-form-data.ts
@@ -15,9 +15,9 @@ export function encodeFormData(params: any): FormData | string {
   if (useFormData) {
     const formData = new FormData();
     Object.keys(newParams).forEach((key: any) => {
-      if (key === "file" && newParams[key].name) {
-        // Pass on the file's name if provided to override defaults such as "blob"
-        formData.append(key, newParams[key], newParams[key].name);
+      if (typeof Blob !== "undefined" && newParams[key] instanceof Blob) {
+        // Pass on the explicit file name to override default name such as "blob"
+        formData.append(key, newParams[key], newParams[key].name || key);
       } else {
         formData.append(key, newParams[key]);
       }
--- a/packages/arcgis-rest-request/test/utils/encode-form-data.test.ts
+++ b/packages/arcgis-rest-request/test/utils/encode-form-data.test.ts
@@ -6,17 +6,35 @@ import {
 import { attachmentFile } from "../../../arcgis-rest-feature-service/test/attachments.test";
 
 describe("encodeFormData", () => {
-  it("should encode in form data for multipart requests", () => {
-    const binaryObj =
-      typeof File !== "undefined"
-        ? new File(["foo"], "foo.txt", {
-            type: "text/plain"
-          })
-        : new Buffer("");
+  it("should encode in form data for multipart file requests", () => {
+    const binaryObj = attachmentFile();
 
     const formData = encodeFormData({ binary: binaryObj });
-
     expect(formData instanceof FormData).toBeTruthy();
+
+    const data = formData as FormData;
+    if (data.get) {
+      expect(data.get("binary") instanceof File).toBeTruthy();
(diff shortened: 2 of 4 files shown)


noahmulfinger:
Naming the file after the key doesn't seem ideal, but I think it's better than passing `blob`.

packages/arcgis-rest-request/src/utils/encode-form-data.ts
```diff
@@ -15,9 +15,15 @@ export function encodeFormData(params: any): FormData | string {
...
-        // Pass on the file's name if provided to override defaults such as "blob"
+      if (typeof File !== "undefined" && newParams[key] instanceof File) {
+        // Pass on the file's name to override defaults such as "blob"
         formData.append(key, newParams[key], newParams[key].name);
+      } else if (
+        typeof Blob !== "undefined" &&
+        newParams[key] instanceof Blob
+      ) {
+        // Pass on the key as the file name to override defaults such as "blob"
+        formData.append(key, newParams[key], key);
```
jgravois: for Blobs, can we continue to use `newParams[key].name` if it has been set and default to using `key` when it hasn't?

packages/arcgis-rest-request/src/utils/encode-form-data.ts
```diff
@@ -15,9 +15,15 @@ export function encodeFormData(params: any): FormData | string {
   if (useFormData) {
     const formData = new FormData();
     Object.keys(newParams).forEach((key: any) => {
-      if (key === "file" && newParams[key].name) {
-        // Pass on the file's name if provided to override defaults such as "blob"
+      if (typeof File !== "undefined" && newParams[key] instanceof File) {
```
jgravois: 👍

packages/arcgis-rest-request/test/utils/encode-form-data.test.ts
```diff
@@ -6,17 +6,40 @@ import {
...
+        ? new Blob([], {
+            type: "text/plain"
+          })
+        : Buffer.from("");
+
+    const formData = encodeFormData({ binary: binaryObj });
+
+    expect(formData instanceof FormData).toBeTruthy();
+
+    if (formData instanceof FormData && formData.get) {
```
jgravois: i like the idea of introspecting the FormData more deeply in browser tests only.

can you update this one too?
https://github.com/Esri/arcgis-rest-js/blob/7ab02af0a6ed86cf0f526609df4f27dac97fde8c/packages/arcgis-rest-items/test/items.test.ts#L690-L693

```ts
// introspect the FormData in the browser only
if (options.body instanceof FormData && options.body.get) {
  expect(options.body.get("token") as string).toEqual("fake-token");
  expect(options.body.get("f")).toEqual("json");
  expect((options.body.get("file") as File).name).toEqual("foo.txt");
}
```

noahmulfinger:
@jgravois Addressed your comments. I realized that since a `File` is a type of `Blob`, I can remove the first if statement and then use `newParams[key].name || key`.

jgravois:
to clarify:

even though Blob doesn't have a `name` property, due to the nature of JavaScript, the property _can_ be set and _will_ be respected when binary data is passed through in a request.

```js
blob.name = "myfile.zip";

arcgisRest.addItemData({
  authentication: session,
  id,
  data: blob
});
```

to recap:

this PR ensures internally that:
* we pass through `Blob/File` filenames appropriately for request parameters other than `file`.
* we lean on the parameter name as a default filename (instead of the word `blob`)

MikeTschudi:
@noahmulfinger , @jgravois , works with my zip file test cases; thanks for the improvement

PR #275, opened 2018-08-10
typo fixes in commitzen template

jgravois:
i don't _think_ we're actually using `support/FormData.d.ts` anywhere.


noahmulfinger:
@jgravois Where is the commit template actually used? It doesn't seem to be used when I run `npm run c` or `git commit`

jgravois:
good point. the template we use comes from [`cz-lerna-changelog`](https://www.npmjs.com/package/cz-lerna-changelog).

PR #285, opened 2018-08-16
Feature: new addItemResource method

jgravois:
i couldn't take it anymore so i rearranged our menagerie of item methods into unique files.

the new stuff is in ebbdc7ea501f79f144fc4c6f870e6ff2f1a554fa


jgravois:
why not just release both at the same time?

PR #288, opened 2018-08-17
fix(portal): allow devs to pass through a bare portal url

jgravois:
a small tweak to make `getPortalUrl()` more resilient in situations where a developer passes through a `portal` without bothering to declare a protocol.

```ts
// old
getPortalUrl({ portal: "bare.com/sharing/rest" } as IRequestOptions)
>> "bare.com/sharing/rest"

// new
getPortalUrl({ portal: "bare.com/sharing/rest" } as IRequestOptions)
>> "https://bare.com/sharing/rest"
```
AFFECTS PACKAGES:
@esri/arcgis-rest-request

### DIFF
--- a/demos/batch-geocoder-node/package.json
+++ b/demos/batch-geocoder-node/package.json
@@ -28,7 +28,7 @@
   "homepage": "https://github.com/Esri/arcgis-rest-js#readme",
   "dependencies": {
     "@esri/arcgis-rest-auth": "^1.8.0",
-    "@esri/arcgis-rest-common-types": "^1.7.1",
+    "@esri/arcgis-rest-common-types": "^1.8.0",
     "@esri/arcgis-rest-geocoder": "^1.8.0",
     "@esri/arcgis-rest-request": "^1.8.0",
     "isomorphic-fetch": "^2.2.1",
--- a/packages/arcgis-rest-auth/package.json
+++ b/packages/arcgis-rest-auth/package.json
@@ -19,7 +19,7 @@
     "@esri/arcgis-rest-request": "^1.7.1"
   },
   "devDependencies": {
-    "@esri/arcgis-rest-common-types": "^1.7.1",
+    "@esri/arcgis-rest-common-types": "^1.8.0",
     "@esri/arcgis-rest-request": "^1.8.0"
   },
   "scripts": {
--- a/packages/arcgis-rest-feature-service/package.json
+++ b/packages/arcgis-rest-feature-service/package.json
@@ -19,7 +19,7 @@
     "@esri/arcgis-rest-request": "^1.7.1"
   },
   "devDependencies": {
-    "@esri/arcgis-rest-common-types": "^1.7.1",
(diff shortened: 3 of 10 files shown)


patrickarlt:
@jgravois would it make more sense to just throw an error here instead. The library doesn't correct me if I make this mistake in other places so why should it here?

jgravois:
> The library doesn't correct me if I make this mistake in other places.

i think it'd be a DX improvement if it did. there are all sorts of menial errors we could easily trap for and fix internally. currently the API feels really picky/brittle.

* did the url include a protocol?
* did it include /sharing/? /rest/? both?
* does it end in a trailing slash?

prior art (of yours :wink:): https://github.com/Esri/esri-leaflet/blob/33363b7aebed6e044f7ca86668c99cbdaffde8cb/src/Util.js#L110-L120

i don't think we need to bend over backwards, but our own REST API is inconsistent in its decision to return the protocol along urls.

because of idiosyncracies like this, my opinion is that a 'fix' would be more useful than an 'error' here.

example: https://edn.maps.arcgis.com/sharing/rest/portals/self?f=json
```js
// 20180817162333
// https://edn.maps.arcgis.com/sharing/rest/portals/self?f=json

{
  "customBaseUrl": "maps.arcgis.com",
  "helperServices": {
  // ...  
  "asyncClosestFacility": {
    "url": "https://logistics.arcgis.com/arcgis/rest/services/World/ClosestFacility/GPServer/FindClosestFacilities",
    "defaultTravelMode": "FEgifRtFndKNcJMJ"
  }
}
```

tomwayson:
> the library doesn't correct me if I make this mistake in other places so why should it here?

If by that you mean `requestOptions.url` - I think that's different scenario than this. This is a fn whose purpose is to return a URL, and to me it's _somewhat_ implicit that it is going to normalize the URL in the process.

I'm torn about whether this fn should normalize _all_ the cases @jgravois mentions in https://github.com/Esri/arcgis-rest-js/pull/288#issuecomment-414011729

I think we should considering adding another fn like:

```ts
function normalizePortalUrl(url: string, options?: {
  addSharingRest?: boolean,
  addTrailingSlash?: boolean
}): string
```

That would w/o any additional `options` passed, at just prepend `https` if the `url` doesn't include a protocol. I also think we should call that _not only_ on the `requestOptions.portal`, but also `requestOptions.authentication.portal` here:

https://github.com/Esri/arcgis-rest-js/pull/288/files#diff-e3757c1050b9a73fb027fb35e31f525fR25

patrickarlt:
@jgravois @tomwayson I agree that we should at least do as Esri Leaflet does and handle leading/trailing whitespace and add a trailing slash wherever we accept a URL as a param (`requestOptions.url`, `*.poral`, ect...). This is mostly for our own sanity since we do ALOT of URL concatenation.

@jgravois I guess for ergonomics sake we could cleanup portal URLs automatically, but I feel like we should at least throw out a `console.warn` if a developer puts in something we have to clean up.

PR #294, opened 2018-08-23
feat(users): add getUserNotifications function

mjuniper:
AFFECTS PACKAGES:
@esri/arcgis-rest-users

Thought I'd take a stab at this. Not really done-done (I would want to add createNotification and removeNotification) but I would like feedback, especially of the "OMG you did this RONG" variety.

### DIFF
--- a/packages/arcgis-rest-common-types/src/index.ts
+++ b/packages/arcgis-rest-common-types/src/index.ts
@@ -465,3 +465,12 @@ export type esriUnits =
   | "esriSRUnit_Kilometer"
   | "esriSRUnit_NauticalMile"
   | "esriSRUnit_USNauticalMile";
+
+export interface INotification {
+  id: string;
+  type: string;
+  target: string;
+  targetType: string;
+  received: number;
+  data: { [key: string]: any };
+}
--- a/packages/arcgis-rest-common-types/src/index.ts
+++ b/packages/arcgis-rest-common-types/src/index.ts
@@ -465,12 +465,3 @@ export type esriUnits =
   | "esriSRUnit_Kilometer"
   | "esriSRUnit_NauticalMile"
   | "esriSRUnit_USNauticalMile";
-
-export interface INotification {
-  id: string;
-  type: string;
-  target: string;
-  targetType: string;
-  received: number;
-  data: { [key: string]: any };
-}
--- a/packages/arcgis-rest-users/src/users.ts
+++ b/packages/arcgis-rest-users/src/users.ts
@@ -69,12 +69,6 @@ export function getUser(
   return request(url, options);
 }
 
-export interface IGetUserNotificationRequestOptions extends IRequestOptions {
-  /**
-   * A session representing a logged in user.
-   */
-  authentication?: UserSession;
-}
 export interface INotificationResult {
   notifications: INotification[];
 }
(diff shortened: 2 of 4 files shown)


packages/arcgis-rest-users/src/users.ts
```diff
@@ -68,3 +68,35 @@ export function getUser(
   // send the request
   return request(url, options);
 }
+
+export interface IGetUserNotificationRequestOptions extends IRequestOptions {
```
jgravois: `IRequestOptions` already includes optional authentication, so no need to extend it with a custom interface.

mjuniper: I was wondering about that but copied it from [here](https://github.com/Esri/arcgis-rest-js/blob/af61e3e9b2a7fbc164b1d3e7ccad8e9b6a2079ea/packages/arcgis-rest-users/src/users.ts#L17). Is that an error or is it not needed there either?

mjuniper: Oh, now I remember the other thing. I actually tried that but TS complained because the `autehentication` on `IRequestOptions` is an `IAuthenticationManager`, not a `UserSession` which I think is what I want. I'm a TS noob so advice is welcome.

mjuniper: Gotcha, thanks John!

packages/arcgis-rest-common-types/src/index.ts
```diff
@@ -465,3 +465,12 @@ export type esriUnits =
   | "esriSRUnit_Kilometer"
   | "esriSRUnit_NauticalMile"
   | "esriSRUnit_USNauticalMile";
+
+export interface INotification {
```
tomwayson: I think this interface could live in @esri/arcgis-rest-users, since it may only be used in that package - or are we putting all interfaces in common-types these days?

packages/arcgis-rest-users/src/users.ts
```diff
@@ -68,3 +68,29 @@ export function getUser(
...
+ * @param requestOptions - options to pass through in the request
+ * @returns A Promise that will resolve with the user's notifications
+ */
+export function getUserNotifications(
+  requestOptions: IUserRequestOptions
+): Promise<INotificationResult> {
+  let url;
+  let options = { httpMethod: "GET" } as IUserRequestOptions;
+
+  const username = requestOptions.authentication.username;
```
tomwayson: is the idea that users can only fetch their own notifications?

packages/arcgis-rest-common-types/src/index.ts
jgravois: i like where it is, but it'd be fine in this package too.

_this is why no one knows where to put interfaces_.

packages/arcgis-rest-users/src/users.ts
mjuniper: That was my thinking.

packages/arcgis-rest-common-types/src/index.ts
mjuniper: That makes sense.

mjuniper:
Added `removeNotification` + tests. Feedback welcome.

packages/arcgis-rest-users/src/users.ts
```diff
@@ -68,3 +68,54 @@ export function getUser(
...
+  const username = encodeURIComponent(requestOptions.authentication.username);
+  const portalUrl = getPortalUrl(requestOptions);
+  const url = `${portalUrl}/community/users/${username}/notifications`;
+  options = { ...requestOptions, ...options };
+
+  // send the request
+  return request(url, options);
+}
+
+export function removeNotification(
```
dbouwman: Pls add comments so the generated doc has something in it :)

PR #296, opened 2018-08-23
Minor error handling fixes

noahmulfinger:
Two commits:
- check for errors function: throw an error for a response with a `failure` status
- ArcGISRequestError: In the case that null or an empty string is passed to the constructor, replace with UNKNOWN_ERROR or UNKNOWN_ERROR_CODE

Pretty minor edge cases that I doubt others will encounter, but shouldn't hurt to handle them.

### DIFF
--- a/packages/arcgis-rest-request/src/utils/ArcGISRequestError.ts
+++ b/packages/arcgis-rest-request/src/utils/ArcGISRequestError.ts
@@ -53,12 +53,15 @@ export class ArcGISRequestError {
    * @param options - The original options and parameters of the request
    */
   constructor(
-    message = "UNKNOWN_ERROR",
-    code: string | number = "UNKNOWN_ERROR_CODE",
+    message?: string,
+    code?: string | number,
     response?: any,
     url?: string,
     options?: IRequestOptions
   ) {
+    message = message || "UNKNOWN_ERROR";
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-request/src/utils/ArcGISRequestError.ts
```diff
@@ -53,12 +53,15 @@ export class ArcGISRequestError {
    * @param options - The original options and parameters of the request
    */
   constructor(
-    message = "UNKNOWN_ERROR",
-    code: string | number = "UNKNOWN_ERROR_CODE",
+    message?: string,
+    code?: string | number,
```
jgravois: based on your refactor, my instinct is that `message` and `code` should remain _required_ parameters.

noahmulfinger: But message and code weren't required before. They had default values if no value was passed in.

jgravois: ah, gotcha. i should have taken a closer look at the existing tests.

https://github.com/Esri/arcgis-rest-js/blob/af61e3e9b2a7fbc164b1d3e7ccad8e9b6a2079ea/packages/arcgis-rest-request/test/utils/ArcGISRequestError.test.ts#L33-L39

PR #298, opened 2018-08-27
fix(sharing): ensure requestOptions.authentication is defined before …

mjuniper:
…accessing

AFFECTS PACKAGES:
@esri/arcgis-rest-sharing

### DIFF
--- a/packages/arcgis-rest-sharing/src/helpers.ts
+++ b/packages/arcgis-rest-sharing/src/helpers.ts
@@ -31,7 +31,8 @@ export interface ISharingResponse {
 }
 
 export function getSharingUrl(requestOptions: ISharingRequestOptions): string {
-  const username = requestOptions.authentication.username;
+  const username =
+    requestOptions.authentication && requestOptions.authentication.username;
   const owner = requestOptions.owner || username;
   return `${getPortalUrl(requestOptions)}/content/users/${encodeURIComponent(
     owner


patrickarlt:
@mjuniper should we just throw an error here instead?

tomwayson:
uggh... good catch @jgravois

@patrickarlt if we keep the current logic, I think this should throw an error if `authentication` (or `requestOptions` for that matter) is not passed in.

That said, I'm not sure we should make it so `authentication` is always required.

The problem is that in the Hub apps, passing in the `UserSession` that we cache on login (`session.authMgr`) was causing issues in other parts of the application, so I changed it so that all fns pass in `portal` and `token` instead of a full blown `authentication`: https://github.com/Esri/ember-arcgis-portal-services/pull/128

It might be better if this fn requires _either_ an `authentication` _or_ an `owner`.

tomwayson:
Alternatively, as a quick(ish) fix that would not require us to make updates to rest-js (yet), I could add an optional `useAuthentication: false` argument to `addOptions()` in EAPS, and we could set it to `true` for this one request for now, since I'm _pretty_ sure it was item fetching requests that were having issues using `authentication`.

patrickarlt:
@tomwayson it feels like we should probably just figure out why you couldn't pass your `session.authMgr` object instead... What were the errors that you were seeing? I don't think we should make this optional OR allow just passing in a bare `token` option.

@tomwayson what you considered just creating an object like this and passing it as `authentication`?

```js
const tempAuthObj = {
   portal: 'whatever the users portal is',
   token: function () {
    return Promise.resolve('the token you want to use')
  }
}
```

But we should really fix this at the source. If we allow skipping `authentication` as a param we also allow users to miss out on all the federation work we have done.

tomwayson:
the errors are on our end b/c Hub has some scenarios where it needs to effectively impersonate another user from another org.

Don't ask.

That plus some URL shenanigans we have to do for enterprise environments mean that there are times when we need to use a different `portal` and/or `token` than those associated w/ the current user.

I think one question we should consider is, is passing a `portal` and `token` a first class supported alternative to passing an `authentication` in rest-js? I understand why it's less _desirable_ than passing an auth (federation, etc). It currently works for _some_ packages/fns, but should we make viable for all of them?

> you considered just creating an object like this and passing it as authentication?

Yeah, I was also thinking something along those lines, though I'm not convinced that's any better than passing `portal` / `token`.

tomwayson:
ok, after thinking this through a bit, I I'm now coming down on the side of `authentication: tempAuthObj` is a better pattern than passing in `requestOptions.portal` and `requestOptions.params.token` - if for no other reason than it's more in line w/ the advertised [requestOptions API](https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/). 

@dbouwman already has a PR open to move us in that direction: https://github.com/Esri/ember-arcgis-portal-services/pull/132

I suggest closing this PR, or changing it so that it's purpose is to throw an error if no `requestOptions` or `requestOptions.authentication` are passed in.

jgravois:
resolved via https://github.com/Esri/ember-arcgis-portal-services/commit/86f903695c3fab2df3e2ffc5e2662f62d8d89f81

PR #299, opened 2018-08-29
Added types for three transpiling problems

MikeTschudi:
@jgravois, opening the repo in VS Code with tsc v3 showed three problems in arcgis-rest-sharing\helpers.ts:

* assumed type `any` for `user` on line 52
* assumed type `any` for `err` on line 84
* unable to assign `Promise<string|void>` to `Promise<string>` on line 63

The changes in this PR remove the problems, but I'm new to this repo's environment and don't want to break anything.

### DIFF
--- a/packages/arcgis-rest-sharing/src/helpers.ts
+++ b/packages/arcgis-rest-sharing/src/helpers.ts
@@ -49,7 +49,7 @@ export function isOrgAdmin(
 ): Promise<boolean> {
   const session = requestOptions.authentication as UserSession;
 
-  return session.getUser(requestOptions).then(user => {
+  return session.getUser(requestOptions).then((user: IUser): boolean => {
     if (!user || user.role !== "org_admin") {
       return false;
     } else {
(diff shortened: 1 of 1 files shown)


tomwayson:
Thanks @MikeTschudi!

I've pulled this locally to test, but just so I understand, this only fixes errors that you see in VS Code, right? Not when you run `npm run lint` at the monorepo root, right, b/c I still see dozens of those on this branch, but I also see what appears to be the same on master (which surprises me).

tomwayson:
looks like we don't run `lint` before tests when commiting or in travis, so that 'splains the lint errors. that raises existential questions that are out of the scope of this PR.

As for what's in this PR, I don't see any errors on for these lines on this branch, nor on master. 

I agree w/ this change: https://github.com/Esri/arcgis-rest-js/pull/299/files#diff-3a3bfbc01d2c46e08d8fdf5150fd79b3R52

But I have to disagree w/ TS on `getUserMembership()` - as near as I can tell the promise _always_ resolves a `string` b/c the `catch()` calls `throw` which means the promise returned by the fn should `reject` not `resolve`, right?

MikeTschudi:
@tomwayson & @jgravois, I propose closing this PR in deference to #301

jgravois:
sounds good.

PR #300, opened 2018-08-30
Added create folder function

MikeTschudi:


packages/arcgis-rest-items/src/create.ts
```diff
@@ -13,6 +16,38 @@ export interface IItemAddRequestOptions extends IItemCrudRequestOptions {
...
+ * import { createFolder } from '@esri/arcgis-rest-items';
+ *
+ * createFolder("Map Collection", {
+ *   authentication: userSession
+ * })
+ * ```
+ *
+ * @param folder - Name to be assigned to new folder
+ * @param requestOptions - Options for the request
+ * @returns A Promise that creates a folder
```
tomwayson: suggest: "A Promise that resolves with folder details once the folder has been created"

packages/arcgis-rest-items/src/create.ts
```diff
@@ -13,6 +16,38 @@ export interface IItemAddRequestOptions extends IItemCrudRequestOptions {
...
+ */
+export function createFolder(
```
tomwayson: Should we call this `createUserFolder()` in case we end up having to add a fn to support [the admin create folder end point](https://developers.arcgis.com/rest/enterprise-administration/server/createfolder.htm).

Otherwise I suppose we could call that `createServerFolder()`. Actually, that's probably better since that is going to be used less frequently, and we can keep this more commonly used fn name shorter.

packages/arcgis-rest-items/src/create.ts
```diff
@@ -13,6 +16,38 @@ export interface IItemAddRequestOptions extends IItemCrudRequestOptions {
...
+  folderName: string,
```
tomwayson: I know our convention on this is not clear, but I think the only time we add named arguments like `folderName` instead of creating a new `IAddFolderRequestOptions` is when there is a possible scenario where we the fn could be called w/o request options. For example, `getItem()` takes an `itemId` as a string b/c you could fetch a public item w/ just that single argument. However, in this case, I think you _always_ need at least the `title` and `authentication` so I think we should use `IAddFolderRequestOptions`.

If we do that, I think we ought to change the name of the option from `folderName` to `title` since [that's what the underlying parameter is called](https://developers.arcgis.com/rest/users-groups-and-items/create-folder.htm).

jgravois: > I know our convention on this is not clear.

@tomwayson is right (both about the convention and lack of clarity). you can check out #78 if you're interested in the backstory.

> I think we ought to change the name of the option

in general i think its the right move for us to use different terminology than the underlying REST API if it will reduce confusion, but `title` and `folderName` are more or less equivalent, so probably simpler to just pass through here.

tomwayson:
Thanks @MikeTschudi!

Issue #236, opened 2018-06-29
createItemInFolder expects a folderId, not a folder name


jgravois:
https://github.com/Esri/arcgis-rest-js/blob/2c52423c9abef5454505274fbe783011aceed8f9/packages/arcgis-rest-items/src/items.ts#L143-L145

i did a little testing this morning and realized you can't pass a folder name (or title) to createItem.

```js
arcgisRest.createItem({
  authentication: session,
  folder: 'foo',
  item: {
    tags: ['no']
  }
})
// Message: User folder does not exist.
```
it'd be a breaking change to rename the property to `folderId`, but it'd be appropriate. in the interim, we should at least clarify this in the inline documentation.

PR #306, opened 2018-09-05
Make moveItem's folder parameter optional

MikeTschudi:


packages/arcgis-rest-items/src/update.ts
```diff
@@ -27,7 +27,7 @@ export interface IItemMoveRequestOptions extends IItemCrudRequestOptions {
    * Alphanumeric id of folder to house moved item. If null, empty, or "/", the destination is the
    * root folder.
    */
-  folder: string;
+  folder?: string;
```
jgravois: my preference would be to use `folderId` consistently instead of `folder`.

see #236

MikeTschudi: @jgravois, I agree, but then it would be different than the corresponding parameter in the REST call (https://developers.arcgis.com/rest/users-groups-and-items/move-item.htm); is that OK?

jgravois: yes. I think the REST parameter name is a footgun.

tomwayson: yeah, we're even _less_ clear about our guidelines (or lack thereof) on trying to stick to the shape (names, etc) of the REST API, except in those cases where it doesn't make sense. Some ambiguous balance of replicating an API that people are already familiar with and building the API we would want to use.

PR #307, opened 2018-09-06
Added create feature service

MikeTschudi:


dbouwman:
@MikeTschudi Awesome work! I've not done enough work with creating feature services to be able to comment on the code itself, but maybe @patrickarlt or @tomwayson can 👀 the details of this so we can get this merged.

Related - we have landed a number of PR's recently, and I'll have one to fix a bug in `UserSession.getUser` later today - I think we should look at cutting a release early next week... unless anyone objects...

noahmulfinger:
Not sure if this has already been discussed offline, but I know originally we were thinking of putting this function in a `feature-service-admin` package (https://github.com/Esri/arcgis-rest-js/issues/46). I know this function fits in well with items functions, but users might expect `createFeatureService` to be in a feature service package instead. I don't feel strongly about it, but something to think about.

tomwayson:
So glad you've started on this @MikeTschudi! We are going to need this very soon too. I haven't reviewed the code yet, but I did want to weigh in on the package question that @noahmulfinger brings up.

I never thought about adding these fn's to the items package. To me that somehow actually makes more sense than to the feature-service package. That said, I still think we should create a new `@esri/arcgis-rest-feature-service-admin` package for this fn.

MikeTschudi:
@noahmulfinger, @tomwayson, at first, items felt like an awkward conceptual fit.  It ended up in items because it shared utility functions with other creation and update functions, and as I worked with it, the fit seemed better.  Perhaps instead of moving the function, it should be renamed to something like create a feature service item.

I'm currently working on a function to update it to add layers. At this time, I'm calling it `updateFeatureService` (seemed clearer than REST's `addDefinition`) and putting it into the `arcgis-rest-feature-service` package.

jgravois:
very cool @MikeTschudi!

👍 for putting this `createFeatureService()` function (and eventually `updateFeatureService()`) in a _new_ `feature-service-admin` package that depends on `items`.

if that sounds good to you, let me know and i can i push a commit or two of my own on this branch to do the re-arranging.

MikeTschudi:
That would be great, @jgravois; thanks!

tomwayson:
I _think_ it's good to create the new the admin package b/c that will be a _very_ small subset of users.

One day I'll learn to trust tree shaking I guess, but in the meantime we've already started down a "no package is too small" path, might as well double down. When that day comes, we'll probably have a big breaking release that pulls everything into a couple packages.

MikeTschudi:
That will be fun!  ;-)

packages/arcgis-rest-feature-service-admin/src/create.ts
```diff
@@ -0,0 +1,195 @@
...
+  IItemCrudRequestOptions,
+  determineOwner
+} from "@esri/arcgis-rest-items";
+import { IExtent, ISpatialReference } from "@esri/arcgis-rest-common-types";
+
+/**
+ * A [`createParameters` JSON object for a new
+ * service](https://developers.arcgis.com/rest/users-groups-and-items/create-service.htm).
+ */
+export interface ICreateServiceParams {
```
jgravois: i took the liberty of renaming this interface (and a few others)

packages/arcgis-rest-feature-service-admin/src/create.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Name of the service to be created. This name must be unique. If the name already exists, the operation will fail. ArcGIS Enterprise does not allow spaces or special characters other than underscores in a service name.
+   */
+  name: string;
```
jgravois: this prop is now _required_

packages/arcgis-rest-feature-service-admin/test/mocks/move.ts
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IItemMoveResponse } from "@esri/arcgis-rest-items";
+
+export const MoveToFolderSuccessResponse: IItemMoveResponse = {
```
jgravois: `any` > `IItemMoveResponse`

packages/arcgis-rest-feature-service-admin/test/mocks/service.ts
```diff
@@ -0,0 +1,21 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { ICreateServiceResult } from "../../src/create";
+
+export const FeatureServiceSuccessResponse: ICreateServiceResult = {
```
jgravois: `any` > `ICreateServiceResult`

Issue #309, opened 2018-09-07
Workflows and UserSession._user


dbouwman:
During the "activation" of an Initiative, we have the following (rough) workflow:

- fetch initiative "template" item
- interpolate values into template, and save as new Initiative "instance"
- create a "collaboration group" (shared editing)
- share newly created Initiative to the "collaboration group"

The tricky bit arises in the sharing at the end. Since the Portal Sharing API has some "idiosyncrasies", the `shareItemWithGroup` function does a lot of pre-flight checking, including a check to see if the current user is a member of the group. It currently does this by checking if the group is in the `UserSession._user`'s array of groups.

The trouble is that `UserSession.getUser()`, tries to be "smart" in that it will only fetch the user from the Portal if it does not already have a cached version and of course, the cached `user` can not possibly have the group because we just created it. 

In my local AGRjs branch, I updated the `getUserMembership` call to fetch the group, and look at it's `.userMembership` property instead of looking in the `UserSession.user.groups`. However, that just seems like a wasted xhr, when the app actually knows the user is a member of the group. 

Question is - how can we inform AGRjs's group sharing logic of this? Should we send another optional param that basically says "trust me, the user is a member"? 

I also sense that this may crop up in more scenarios... thus begging the question of keeping the `UserSession.user` in sync with what's on the server-side? Should we even try?

For this specific scenario, I'm disinclined to make the `addGroup` function inject the group into `UserSession._user.groups` as that seems "side-effecty" and get's into managing state etc... but i'm unclear how else to wrangle this...

Anyhow - open to ideas...

jgravois:
thanks for this write up. i still think our best option is to expose something in the API that guarantees a fresh request for user metadata will be made.

lets continue tracking in #291.

PR #310, opened 2018-09-07
Rework group membership checking, simplify UserSession.getUser

dbouwman:
### Group Membership Checking
Previously this used `UserSession.getUser()` and would check to see if the group in question was in the user's array of groups. However, since `UserSession.getUser()` implements a caching strategy, if you created a group and then attempted to share an item to that group, the `getUserMembership` check would return `nonmember` even though the user is actually the owner. More about this general problem is on #309 

While there is some discussion about possibly adding cache-busting to the `getUser()` call, I opted to simply make a call to the group itself, which will return the same info.

I also fixed `isItemSharedWithGroup` where we mutated an `IRequestOptions.params` but then sent that into a subsequent call, which resulted in an unexpected url. We should be careful not to mutate that as calling apps/libraries will pass the same thing into multiple requests when orchestrating calls

### UserSession.getUser() 
Just simplify `return new Promise(resolve => resolve(this._user)` to `return Promise.resolve(this._user)`

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -603,7 +603,7 @@ export class UserSession implements IAuthenticationManager {
    */
   getUser(requestOptions?: IRequestOptions): Promise<IUser> {
     if (this._user && this._user.username === this.username) {
-      return new Promise(resolve => resolve(this._user));
+      return Promise.resolve(this._user);
     } else {
       const url = `${this.portal}/community/users/${encodeURIComponent(
         this.username
--- a/packages/arcgis-rest-sharing/src/group-sharing.ts
+++ b/packages/arcgis-rest-sharing/src/group-sharing.ts
@@ -1,11 +1,7 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import {
-  request,
-  IRequestOptions,
-  getPortalUrl
-} from "@esri/arcgis-rest-request";
+import { request, getPortalUrl } from "@esri/arcgis-rest-request";
 import { IItem } from "@esri/arcgis-rest-common-types";
 
 import {
--- a/packages/arcgis-rest-sharing/src/group-sharing.ts
+++ b/packages/arcgis-rest-sharing/src/group-sharing.ts
@@ -120,7 +120,7 @@ function changeGroupSharing(
                 }`;
               } else {
                 // if they are a group admin/owner, use the bare item url
-                if (membership === "admin") {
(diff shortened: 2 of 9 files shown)


packages/arcgis-rest-sharing/src/helpers.ts
```diff
@@ -58,35 +59,26 @@ export function isOrgAdmin(
...
+ * @export
+ * @param {string} groupId
+ * @param {IGroupSharingRequestOptions} requestOptions
+ * @returns {Promise<string>}
+ */
 export function getUserMembership(
-  requestOptions: IGroupSharingRequestOptions
+  groupId: string,
+  requestOptions: IRequestOptions
 ): Promise<string> {
```
jgravois: i know you hate `IEverything`, but there's no need to introduce a breaking change to the input parameters.

it'd be kinda cool to be even more explicit about what is returned though.

```ts
export function getUserMembership(
  requestOptions: IGroupSharingRequestOptions
): Promise<"owner" | "admin" | "member" | "nonmember"> {
```

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -603,7 +603,7 @@ export class UserSession implements IAuthenticationManager {
    */
   getUser(requestOptions?: IRequestOptions): Promise<IUser> {
     if (this._user && this._user.username === this.username) {
-      return new Promise(resolve => resolve(this._user));
+      return Promise.resolve(this._user);
```
jgravois: 👍

packages/arcgis-rest-sharing/src/helpers.ts
```diff
@@ -58,35 +59,26 @@ export function isOrgAdmin(
...
-      /* istanbul ignore next */ err => {
-        throw Error(
-          `failure determining membership of ${session.username} in group:${
-            requestOptions.groupId
-          }: ${err}`
-        );
-      }
-    );
+    .catch(ex => {
+      return "nonmember";
```
jgravois: pretty sweet to catch the `400` and just pass on the relevant info 🛌!

packages/arcgis-rest-sharing/src/group-sharing.ts
```diff
@@ -179,15 +179,18 @@ function isItemSharedWithGroup(
     sortField: "title"
   };
 
+  // we need to append some params into requestOptions, so make a clone
+  // instead of mutating the params on the inbound requestOptions object
+  const ro = { ...requestOptions };
```
jgravois: good catch 🥇 

`ro` is a little cryptic, `options` would be my preference.

PR #317, opened 2018-09-11
Update feature service to add layer(s) &| table(s)

MikeTschudi:


packages/arcgis-rest-feature-service-admin/src/update.ts
```diff
@@ -0,0 +1,93 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
+import { ILayer, ITable } from "@esri/arcgis-rest-common-types";
+
+export interface IAddToFeatureServiceRequestOptions
```
jgravois: If you're going to call the method `updateFeatureService()`, i'd recommend using `IUpdateFeatureService...` as interface names too.

packages/arcgis-rest-feature-service-admin/src/update.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  return new Promise(resolve => {
+    request(url, requestOptions).then(
+      response => {
+        resolve(response);
+      },
+      response => {
+        // We're not interested in the full ArcGISRequestError response, nor having an exception thrown
+        resolve(response.response);
```
jgravois: i don't quite grok your code comment. what is it exactly that you're going for here? specifically why you don't want devs to `catch()` errors?

MikeTschudi: @jgravois, the underlying REST call sends an HTTP 200 for success or failure, but the request call converts the failure into an exception. Purpose of this section is to 1) align with style of REST call and 2) reduce the wordy content returned by failed request call; much of that content simply echoes what the app sent.  Not needed by app, and more to define in error response type.

packages/arcgis-rest-feature-service-admin/src/update.ts
MikeTschudi: Good point; how about we rename the function to `addToFeatureService`?

packages/arcgis-rest-feature-service-admin/src/update.ts
jgravois: > the request call converts the failure into an exception.

Thats actually an intentional design choice in this API. We think developers will consider it more intuitive to trust that everything went according to plan if a promise resolves and `catch()` errors from our REST API (even if they were sent with a `200` status code).

MikeTschudi: OK; easy enough to change

packages/arcgis-rest-feature-service-admin/src/update.ts
MikeTschudi: Good resource--thanks!

MikeTschudi:
@jgravois, I think that I addressed your review comments.
Thanks for them.

jgravois:
i pushed up another tweak that further simplifies how errors are handled.

please take a look at 31e56db and let me know what you think.

MikeTschudi:
Sounds good, @jgravois; thanks

jgravois:
@MikeTschudi if it'd be helpful to you for us to publish this to npm, just say the word.

MikeTschudi:
@jgravois, what does publishing to npm do, please?

jgravois:
npm is a package manager for sharing JS code that gets more popular by the day. it includes a cli that allows folks to pull down any bundle they're interested in.

```
npm install @esri/arcgis-rest-feature-service-admin
```
https://www.npmjs.com/package/@esri/arcgis-rest-feature-service-admin

_usually_ when folks use npm, they organize the dependencies for their project in a `package.json` file. you can find some straightforward examples of this in our `/demos/` directory.

https://github.com/Esri/arcgis-rest-js/blob/d3f2553e59235c7657a074160a3694e06144b87e/demos/attachments/package.json#L8-L11

i offered to publish a release to npm because the version thats up there right now has the method you contributed to create feature services, but it doesn't yet include the method you added for modifying them.

tomwayson:
we gotta get me back on the release wagon. I'm gunshy after the releases I made didn't bump all the packages. Maybe @jgravois can hold my hand on Mon while I run the scripts?

MikeTschudi:
Thanks, @jgravois.  I've been using @esri/... without thinking about how we get from the repository to @esri/...

Issue #295, opened 2018-08-23
Unable to determine if a user session should be SSL-only


skitterm:
When authenticating with the OAuth implicit grant, a `ssl=true` key-value pair is returned as part of the OAuth callback's hash, when the organization is HTTPS-only. Based on that I can adjust everything to be strictly HTTPS when the organization requires it.

However, when calling `userSession.toCredential()`, `ssl` is [always `true`](https://github.com/Esri/arcgis-rest-js/blob/0b2d8f8c4b5f177471158a0d165089312459c010/packages/arcgis-rest-auth/src/UserSession.ts#L579).

Unless I'm mistaken, that makes it impossible to determine whether or not a session should be SSL-only (without manually reading the oauth callback hash for that value, of course).

Newbie idea, but a way to let `userSession` know if `ssl` should be `true || false`  would be to allow passing it in as a parameter in `completeOAuth2()`?

tomwayson:
I think the way to do that would be to add add a `ssl: boolean` prop to [UserSession](https://esri.github.io/arcgis-rest-js/api/auth/UserSession/) and [UserSessionOptions](https://esri.github.io/arcgis-rest-js/api/auth/IUserSessionOptions/) and then as you say add it here:

https://github.com/Esri/arcgis-rest-js/blob/af61e3e9b2a7fbc164b1d3e7ccad8e9b6a2079ea/packages/arcgis-rest-auth/src/UserSession.ts#L430

Then update `toCredential()` to pull from that value.

I'd gladly accept such a pull request 😄

skitterm:
PR submitted: #332

jgravois:
resolved via #332. thanks again @skitterm 🎉

PR #332, opened 2018-09-20
Add ssl flag

skitterm:
Addresses #295. 

### Features
* Adds `ssl` field to `UserSession` class and `FetchToken` response
* Reads/writes to that field in `to/fromCredential`, `deserialize` and `toJSON`
* Reads oauth-callback hash for `ssl` property in `completeOAuth2`
* Reads `oauth2/token` response for `ssl` property in `exchangeAuthorizationCode`

### Future work needed (not included in this PR)
* Ideally `arcgis-rest-request` methods (and other requests when authenticated) should look for `ssl` property and force the URL to always be HTTPS in that case
* Does `ApplicationSession` need this `ssl` property as well?
* Does `UserSession.getToken` method need to read `ssl` value?


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -165,6 +165,11 @@ export interface IUserSessionOptions {
    */
   portal?: string;
 
+  /**
+   * Whether requests should be made exlusively over HTTPS.
```
jgravois: _exclusively_ 😈

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -456,10 +469,14 @@ export class UserSession implements IAuthenticationManager {
       Date.now() + parseInt(match[2], 10) * 1000 - 60 * 1000
     );
     const username = decodeURIComponent(match[3]);
+    const ssl =
+      win.location.href.indexOf("&ssl=true") !== -1 ||
```
jgravois: this may just be me, but `.indexOf("&ssl=true") > -1` feels more readable than a double-negative.

packages/arcgis-rest-auth/src/fetch-token.ts
```diff
@@ -12,13 +12,15 @@ interface IFetchTokenRawResponse {
   access_token: string;
   expires_in: number;
   username: string;
+  ssl: boolean;
```
tomwayson: I wonder if this should be `ssl?: boolean`?

packages/arcgis-rest-auth/src/fetch-token.ts
```diff
@@ -38,7 +40,8 @@ export function fetchToken(
       username: response.username,
       expires: new Date(
         Date.now() + (response.expires_in * 60 * 1000 - 60 * 1000)
-      )
+      ),
+      ssl: response.ssl === true
```
jgravois: you do this instead of `ssl: response.ssl` to account for instances where the parameter is missing entirely, correct?

tomwayson:
Nice work @skitterm, thanks for the contribution!

packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -19,6 +19,7 @@ describe("UserSession", () => {
     const session = new UserSession({
       clientId: "clientId",
       redirectUri: "https://example-app.com/redirect-uri",
+      ssl: false,
```
jgravois: unless i'm missing something, there's nothing a user can do in the `UserSession` constructor to dicate what the server is going to respond with so there's no point in including `ssl` as an option.

is that correct? if so, we just need to update our examples and make that clear in the API reference.

packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -703,7 +707,7 @@ describe("UserSession", () => {
       const MockWindow = {
         location: {
           href:
-            "https://example-app.com/redirect-uri#access_token=token&expires_in=1209600&username=c%40sey&persist=true"
+            "https://example-app.com/redirect-uri#access_token=token&expires_in=1209600&username=c%40sey&ssl=true&persist=true"
```
jgravois: it'd be super helpful to have one test where `ssl=true` and another where its `false`/not set (i'm not sure which actually happens in the wild)

packages/arcgis-rest-auth/src/fetch-token.ts
skitterm: Yes, doing this to ensure a boolean value. Since from my experience it either returns `ssl: true` or not at all.

packages/arcgis-rest-auth/test/UserSession.test.ts
skitterm: My thinking is that `ssl` is an attribute of a user session, and something that would be needed by arcgis-rest-request's request methods (since they take in a UserSession/ApplicationSession as a param) to determine whether or not to force HTTPS for a request.

tomwayson: I think the main risk w/ the approach in this PR is that a user can set `ssl` before `beginOAuth()` which may overwrite whatever they set. 

One alternative would be that we don't let consuming apps set this and we have some kind of read-only property/fn like `sslOrg` that _always_ returns what was in the redirect url parameter.

Given that we haven't thought through how `request()` is going to use this parameter yet, it may be safer to do the latter. That would be a much more useful signal to `request()`. For example, it could  so it could only force SSL for requests that are made to that org.

I'm open to suggestions though.

jgravois:
awesome contribution. thank you @skitterm!

skitterm:
Should we then remove the `this.ssl = options.ssl` from the constructor itself? Though if we do, I feel like we'd still want to be able to set `ssl` property via a setter.

jgravois:
> Should we then remove the `this.ssl = options.ssl` from the constructor itself?

i didn't spend long with it, but the way things are set up now its necessary to be able to set the property later.

Issue #284, opened 2018-08-16
Documentation pages navigation: highlight currently selected page


gavinr:
I'll admit this is a subjective request, but I personally find it easier when the page I'm on highlights itself in the navigation. It helps me get my bearings if maybe I arrived on this page from a search engine or outside link. So I would propose changing this:
![image](https://user-images.githubusercontent.com/209355/44232173-e1640f00-a165-11e8-86c5-8377fe39988a.png)

to something like this:
![image](https://user-images.githubusercontent.com/209355/44232190-eb860d80-a165-11e8-8a6b-ad8fa60fc801.png)
(notice the background highlight on the selected page in the nav)

Just throwing it out there to see what you all think! Thanks!

jgravois:
_definitely_ a good suggestion.

noahmulfinger:
The highlighting actually works if you run the docs locally. I think we need to change something in the calcite web configuration so it works in both environments.

gavinr:
I see now - yes - when running the doc locally the current page in the navigation is outlined:
![image](https://user-images.githubusercontent.com/209355/44268873-6ef03f00-a1f8-11e8-9ef0-ac782ac1d333.png)

I'm not too familiar with the docs build system, but FWIW it looks like the `list-item-active` class is just not getting added to the div when hosted in production:

![production](https://user-images.githubusercontent.com/209355/44269000-df975b80-a1f8-11e8-9664-361a775bdaec.png) (production)

vs locally hosted:

![locally hosted](https://user-images.githubusercontent.com/209355/44269025-ecb44a80-a1f8-11e8-81b7-61c2e10c9e92.png)

jgravois:
the problem is worse than that. `is-active` isn't being set on page load in production _anywhere_ on page load.

proposed ~~fix~~ hack in #334 	

![screenshot 2018-09-21 12 01 39](https://user-images.githubusercontent.com/3011734/45900962-22b79200-bd96-11e8-86bb-4affed58b2ba.png)

![screenshot 2018-09-21 12 02 28](https://user-images.githubusercontent.com/3011734/45901000-3fec6080-bd96-11e8-9cc9-aec2e5e001c1.png)

gavinr:
looks great @jgravois!

tomwayson:
I hate to even bring this up, but in FF I can get the nav in a state where _both_ nav links appear active by going from one to the other and then using the back button (or key board shortcut):

[...]

PR #334, opened 2018-09-21
docs: adapt built in acetate helper to fix sticky links

jgravois:
ISSUES CLOSED: #284 

🎩 tip to @paulcpederson for pointing out that its actually a helper built right into acetate that helps us set the `is-active` class automatically when a page loads.

it feels like a wretched hack to me to copy/paste the entire helper (and its dependencies) into this library, but i couldn't think of a more clever way to extend the logic in the comparisons.

```js
${BASE_URL}${currentUrl}` === `finalUrl`

// instead of 
currentUrl === finalUrl
```

@paulcpederson also mentioned that we might should be using `{{ relativePath }}` in this website to make our lives easier, but i couldn't get that option to behave either.

```
{% link relativePath + "api/", "API Reference", class="side-nav-link"%}

// .api/ ?

{% link relativePath + "/api/", "API Reference", class="side-nav-link"%}

// /api/ ??

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -2,10 +2,35 @@ const path = require("path");
 const fs = require("fs");
 const { inspect } = require("util");
 const _ = require("lodash");
-const slug = require("slug");
 
 const IS_DEV = process.env.ENV !== "prod";
-const BASE_URL = process.env.ENV === "prod" ? "/arcgis-rest-js" : "";
+const BASE_URL = IS_DEV ? "" : "/arcgis-rest-js";
+
+const url = require("url");
+const PROTOCOL_REGEX = /^(.+:)?\/\//;
+const HASH_REGEX = /^#/;
+
+function stripSlashes(string = "") {
+  var count = string.length - 1;
+  var index = 0;
+
+  while (string.charCodeAt(index) === 47 && ++index);
+  while (string.charCodeAt(count) === 47 && --count);
+
+  string = string.slice(index, count + 1);
+
+  return string;
+}
+
+function ensureTrailingSlash(baseUrl) {
+  let ext = path.extname(baseUrl);
+
+  if (!ext && baseUrl[baseUrl.length - 1] !== "/") {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/acetate.config.js [hunk 1]
```diff
@@ -211,4 +236,84 @@ module.exports = function(acetate) {
       : [];
     return `npm install ${package.name} ${peers.join(" ")}`;
   });
+
+  acetate.helper(
+    "link",
```
patrickarlt: @jgravois this should already be built into Acetate. http://acetate.io/documentation/built-in-helpers/ > Link Helper. Is something wrong with that version? The follwoing should just work in templates `{% link '/guide/', 'Guide' %}`

paulcpederson:
@jgravois you could also just not use the link helper rather than replicating it and changing it inside your project. 

This seems to work pretty well:

```
<nav class="top-nav-list" role="navigation" aria-labelledby="topnav">
  <a href="{{relativePath}}/guides/" class="{% if url == '/guides/'%}is-active{% endif %} top-nav-link">Guides</a>
  <a href="{{relativePath}}/api/" class="{% if url == '/api/'%}is-active{% endif %} top-nav-link">API Reference</a>
</nav>
```

jgravois:
thanks @paulcpederson. doing that for each of the individual 'link`s in the API reference sidebar probably _is_ a better option than hacking on the built in helper.

@patrickarlt i wouldn't say there's anything _wrong_ with what is built into `acetate`. it just doesn't account for the fact that we added a utility in #83 to prepend our baseurls with `/arcgis-rest-js/` _only_ when we are building for production.

this was the best idea we could come up with to host the doc at https://esri.github.io/arcgis-rest-js/.

there's likely a way to configure the raw source to achieve both lofty goals simultaneously, but the exact path eludes me.

jgravois:
> there's likely a way to configure the raw source to achieve both lofty goals simultaneously

figured it out in #341

Issue #321, opened 2018-09-14
add SRI hashed script tag examples to packages docs page


COV-GIS:
Perhaps under `CDN url:` have...

Script tag:
```<script src="https://unpkg.com/@esri/arcgis-rest-request@1.9.0/dist/umd/request.umd.js" integrity="sha384-Iutv/c8BA+49hi5Y9oNmHeFbrxWU62LKbyb2YV/OK3PwGP2e4XFIq//lFTWDXryO" crossorigin="anonymous"></script>```

I'm sure there's a way to automate it.

Just a thought.

jgravois:
this would be a welcome contribution, but i won't have time to tackle it personally any time soon.

pat's great static site generator [`acetate`](https://github.com/patrickarlt/acetate) already provides a great hook to do this. you'd just need to add another dep to generate the hash when the cdn url is composed.

https://github.com/Esri/arcgis-rest-js/blob/master/docs/acetate.config.js#L200-L204

COV-GIS:
Just getting it out there.

Very well may do @jgravois.

COV-GIS:
I was investigating a bit. Don't see any issue using node's `crypto` package to generate hashes for the dist packages and creating a script tag code sample.

John, Tom and Pat...what say you about either replacing `CDN url:` with script tag code or adding it as an additional code sample as suggested above?

COV-GIS:
Sorry fellas...I'm gonna let this one go for now.

The hashes are easy enough to create. And so is composing a script tag string for the docs with an acetate helper. But creating an SRI hash requires a promise, so hashes would need to be created as part of the build or possibly with an acetate plugin to create and include the hashes in the `typedoc.json` for each package.

Too far out in the rhubarb for me at this time.

I'll keep it in mind and might surprise with a PR at a future date.

jgravois:
> But creating an SRI hash requires a promise

i hadn't considered that but it makes sense, only because we don't have a local reference to built files from the latest release readily available.

one way to sidestep that problem would be to add yet another step to our [release script](https://github.com/Esri/arcgis-rest-js/blob/master/support/publish.sh#L17) to calculate hashes for each of the _local_ packages locally and save them in a config file.

https://github.com/Esri/arcgis-rest-js/blob/master/support/publish.sh#L17

leaving this open in case anyone else is interested in tackling.

COV-GIS:
Here's where I got to @jgravois.

[...]

PR #336, opened 2018-09-23
docs(api): add cdn script tag to docs api page and package pages

COV-GIS:
As requested in https://github.com/Esri/arcgis-rest-js/issues/321.

Adds basic script tags to api and package pages.

SRI hash proof of concept https://github.com/COV-GIS/arcgis-rest-js/commit/579636bafebff4f2b352c70863dc5331d5650d5e.

Requesting assistance and/or ideas as how to best create SRI hashes for each min js file when starting up the docs build process.

### DIFF
--- a/docs/src/api/_package.html
+++ b/docs/src/api/_package.html
@@ -3,7 +3,7 @@
 <p class="trailer-half">{{pkg.description}}</p>
 <h2 class="font-size--1 trailer-half">npm install:</h2>
 <pre><code>{% npmInstallCmd pkg %}</code></pre>
-<h3 class="font-size--1 trailer-half">CDN:</h2>
+<h2 class="font-size--1 trailer-half">CDN:</h2>
 <pre><code>{% scriptTag pkg %}</code></pre>
 <hr class="leader-half trailer-half">
 <ul class="list-plain package-contents">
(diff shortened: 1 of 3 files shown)


docs/src/api/_package.html
```diff
@@ -5,6 +5,8 @@ <h2 class="font-size--1 trailer-half">npm install:</h2>
 <pre><code>{% npmInstallCmd pkg %}</code></pre>
 <h2 class="font-size--1 trailer-half">CDN url:</h2>
 <pre><code>{% cdnUrl pkg %}.js</code></pre>
+<h3 class="font-size--1 trailer-half">Script tag:</h2>
+<pre><code>{% scriptTag pkg %}</code></pre>
```
jgravois: ~~`<h2 class="font-size--1 trailer-half">CDN url:</h2>`~~
~~`<pre><code>{% cdnUrl pkg %}.js</code></pre>`~~

if we're going to display a `<script>` tag, lets remove the bare CDN url.

COV-GIS: Done. Retained both helpers as `cdnUrl` is used in AMD example.

COV-GIS:
I was thinking creating a `key:value` hash of `packageName:sriHash` in `acetate.config.js` for access by the helper would work well. I couldn't see a way to get the packages to do so when loading acetate.

COV-GIS:
I just noticed the heading tags were mismatched before and I perpetuated it. Fixed.

PR #337, opened 2018-09-24
docs(api): add sri hashes to script tags

COV-GIS:
I was over thinking this. There's no need somehow create hashes in the actual acetate build process. There's only 9 packages and I know their names. Hashes can be generated and stored in an object for use by the helper.

The hashes are only valid for deployment. The local dist files will always hash differently on account of the copyright timestamp being different when bootstrapping `npm install` or `npm run build`. When deving docs the hashes will be wrong, nor should be used in the demos.

I don't forsee an issue with build and docs deploy. Looks like `publish.sh` should be using the same dist files as will be available via the CDN. You probably want to scrutinize this assumption to be sure.

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -3,11 +3,51 @@ const fs = require("fs");
 const { inspect } = require("util");
 const _ = require("lodash");
 const slug = require("slug");
+const sriToolbox = require("sri-toolbox");
 
 const IS_DEV = process.env.ENV !== "prod";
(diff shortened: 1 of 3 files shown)


jgravois:
getting closer 😸

one complication though is that we don't just push doc updates to production when we tag releases. they go live _every time_ we merge into `master`.

COV-GIS:
Check roger.

I'll start on Plan B as time allows.

PR #344, opened 2018-10-02
Add "access" param to IItemResourceRequestOptions.

ssylvia:
Items now support `private` resource items: https://developers.arcgis.com/rest/users-groups-and-items/add-resources.htm. This PR adds a param to the types to support this.

### DIFF
--- a/packages/arcgis-rest-items/src/add.ts
+++ b/packages/arcgis-rest-items/src/add.ts
@@ -95,6 +95,7 @@ export function addItemResource(
     file: requestOptions.resource,
     fileName: requestOptions.name,
     text: requestOptions.content,
+    access: requestOptions.access,
     ...requestOptions.params
   };
 
--- a/packages/arcgis-rest-items/src/helpers.ts
+++ b/packages/arcgis-rest-items/src/helpers.ts
@@ -31,6 +31,11 @@ export interface IItemResourceRequestOptions extends IItemIdRequestOptions {
    * Text input to be added as a file resource.
    */
   content?: string;
+  /**
+   * Set file resource to be private regardless of the item access level, or revert it by setting
+   * it to inherit which makes the item resource have the same access as the item.
+   */
+  access?: "private" | "inherit";
   resource?: any;
 }
 
--- a/packages/arcgis-rest-items/src/update.ts
+++ b/packages/arcgis-rest-items/src/update.ts
@@ -82,6 +82,7 @@ export function updateItemResource(
 
   // mix in user supplied params
   requestOptions.params = {
+    file: requestOptions.resource,
     fileName: requestOptions.name,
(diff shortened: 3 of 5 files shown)


jgravois:
> I thought  `...requestOptions.params` was adding the rest of the options.

close. `params` is a backdoor that allows developers to pass through _any_ arbitrary parameter to `@esri/arcgis-rest-request` they want.

```ts
// so you can already do this
addItemResource({
  id: "3ef",
  resource: file,
  params: { access: "private" } 
})
```

you raise a good point about being wary of introducing a footgun that allows developers to _accidentally_ revert the new access setting when they make an update to a resource.

but maybe we could have the best of both worlds with something like this?

```ts
// add
requestOptions.params = {
  file: requestOptions.resource,
  fileName: requestOptions.name,
  text: requestOptions.content,
  access: requestOptions.private ? "private" : "inherit",
  ...requestOptions.params
};

// update
requestOptions.params = {
  fileName: requestOptions.name,
  text: requestOptions.content,
  ...requestOptions.params
};

// only set
if (requestOptions.private) {
  requestOptions.params.access = "private";
}
``

**Comment by ssylvia** [comment]
@jgravois But if we do that, in my workflow above,

> 1. Add a resource and set the access
> 2. Update the resource as needed, no reason to add access param
> 3. In a publish workflow, update the resources access

our app would use:

Add: 
```
addItemResource({
  resource: file,
  private: true
})
```

Update:
```
updateItemResource({
  id: "3ef",
  resource: file
})
```

Publish:
```
updateItemResource({
  id: "3ef",
  resource: file,
  params: { access: "inherit" } 
})
```

Seems better to keep access parameter consistent.

**Comment by jgravois** [comment]
thanks for taking the time to explain your exact flow.

i've opened up https://github.com/ssylvia/arcgis-rest-js/pull/1 with a suggestion (and tests) to ensure internally that we do the following:

```ts
// set access: "private"
updateItemResource({
  id: "3ef",
  private: true
})

// set access: "inherit"
updateItemResource({
  id: "3ef",
  private: false
})

// don't set the access property at all
updateItemResource({
  id: "3ef"
})
```

do those tweaks put you at ease? i don't mean to beat a dead horse, it just feels like better DX _not_ to expose the raw request parameter name and values here.

@tomwayson? (or anyone else). just tell me and i'll give it a _rest_ (groan)

ssylvia:
@jgravois, yep that looks like a fine way to do it. I fixed one failing test you had. Also, my last commit adds the "file" to the `updateItemResource` so the file can be updated with the same param as the `addItemResource`.
```
requestOptions.params = {
    file: requestOptions.resource,
```
It was already documented this way, but it practice had to use

```
updateItemResource({
  params: {
    file: [FILE]
  }
})
```

Issue #343, opened 2018-10-01
GeoJSON typing for the feature service package


itrew:
Since GeoJSON is a valid output for `FeatureLayer/query` endpoints, I was thinking it may be helpful to add that as an alternative response type for the `getFeature`, `queryFeatures`, and `queryRelated` methods. Maybe a Typescript generic for those three methods?

The reason I bring this up is that I am working on a library that needs to grab data from a service before processing it. The utility I will be using for processing desires GeoJSON, so my goal was to just overwrite the request options with the output format of 'geojson'.  Right now, I have to assert the response to the GeoJSON type as follows. 

```
import * as fetch from 'isomorphic-fetch'
import * as FormData from 'form-data'
import { queryFeatures, IQueryFeaturesRequestOptions } from '@esri/arcgis-rest-feature-service'

export async function fetchESRIGeoJSON(
  options: IQueryFeaturesRequestOptions
): Promise<GeoJSON.GeoJSON> {
  // Overwrite the output format parameter.
  options.params = {
    ...options.params,
    ...{ f: "geojson" }
  }
  // Requires the assertion of the queryFeatures response as a GeoJSON.
  const features: GeoJSON.GeoJSON = await queryFeatures(options) as GeoJSON.GeoJSON
  return features
}
```

It works fine for now, but just a thought for the future.

jgravois:
we should definitely advertise that /query can return a GeoJSON feature collection.

```ts
// new devDependency / peerDependency
import FeatureCollection from "@types/geojson";

export function queryFeatures(
  requestOptions: IQueryFeaturesRequestOptions
): Promise<IQueryFeaturesResponse | IQueryResponse | FeatureCollection> {
```

itrew:
So returning a union type brings along it's own quirks. After making the change @jgravois suggested, we run into the following scenario:

```typescript
import { FeatureCollection } from 'geojson'
import { queryFeatures, IQueryFeaturesResponse } from './query'

queryFeatures({
  url: 'www.esri.com/some_feature_service_url'
}).then((response) => {
  // Response is inferred to be of type <IQueryFeaturesResponse | IQueryResponse | FeatureCollection>
  
  // Errors because features property does not exist on the type IQueryResponse
  // features is type <any>
  const features = response.features
  
  // Works because of the assertion.
  // features2 (ESRI Feature) is type <IFeature[]>
  const features2 = (response as IQueryFeaturesResponse).features
  
  // Works because of the assertion.
[...]

PR #346, opened 2018-10-02
Address #343. Add GeoJSON types to feature service package.

itrew:
Added GeoJSON output type to several methods in the feature service package.

AFFECTS PACKAGES:

This is my first crack at adding some overloads and generics thanks to the suggestion by @tomwayson in #343. I'm not 100% on this, so please give it a good shake.

One thing I noticed is that it is still a little wonky with the `f` parameter if the type isn't explicit. For example, the first snippet errors and the second does not.

```
const requestOptions = {
      url: serviceUrl,
      id: 42,
      params: {
        f: 'geojson'
      }
    };
getFeature(requestOptions) <-- Errors because f: string is not assignable to type 'geojson'.
```
```
const requestOptions: IGetFeatureRequestOptions<IQueryGeoJSONFormatParams> = {
      url: serviceUrl,
      id: 42,
      params: {
        f: 'geojson'
      }
    };
getFeature(requestOptions) <-- Now works
```
Maybe there is a more graceful way to handle that? I'm all ears.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -29,6 +29,7 @@ module.exports = function(config) {
       tsconfig: "./tsconfig.json",
       bundlerOptions: {
         transforms: [require("karma-typescript-es6-transform")()],
+        ignore: ["geojson"],
         resolve: {
           // karmas resolver cant figure out the symlinked deps from lerna
           // so we need to manually alias each package here.
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,12 +13,6 @@
         "conventional-changelog": "0.0.17"
       }
     },
-    "@types/es6-promise": {
-      "version": "0.0.32",
-      "resolved": "https://registry.npmjs.org/@types/es6-promise/-/es6-promise-0.0.32.tgz",
-      "integrity": "sha1-O89E+x5Cnz33YYjIxth0Rjujcf0=",
-      "dev": true
-    },
     "@types/events": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@types/events/-/events-1.2.0.tgz",
@@ -37,7 +31,7 @@
       "integrity": "sha512-Z5nu9Pbxj9yNeXIK3UwGlRdJth4cZ5sCq05nI7FaI6B0oz28nxkOtp6Lsz0ZnmLHJGvOJfB/VHxSTbVq/i6ujA==",
       "dev": true,
       "requires": {
-        "@types/node": "*"
+        "@types/node": "6.0.111"
       }
     },
     "@types/glob": {
@@ -46,9 +40,9 @@
       "integrity": "sha512-wc+VveszMLyMWFvXLkloixT4n0harUIVZjnpzztaZ0nKLuul7Z32iMt2fUFGAaZ4y1XWjFRMtCI5ewvyh4aIeg==",
       "dev": true,
       "requires": {
-        "@types/events": "*",
-        "@types/minimatch": "*",
-        "@types/node": "*"
+        "@types/events": "1.2.0",
+        "@types/minimatch": "2.0.29",
+        "@types/node": "6.0.111"
       }
     },
     "@types/handlebars": {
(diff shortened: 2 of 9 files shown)


itrew:
Oops, looks like it didn't import the GeoJSON type namespace. I'll take a look at fixing that later today.

itrew:
@tomwayson I think the overloading does work with minimal thought, but it's the generics that make it a little more cumbersome. They way I set it up, however, should be able to hide that if TS is able to infer the parameter types correctly.

Unfortunately it looks like the BS from https://github.com/Esri/arcgis-rest-js/pull/346#issue-219830417 is more due to a limitation of the literal types used for the output formats doesn't match how TS infers strings. Based on the info [here](https://basarat.gitbooks.io/typescript/docs/types/literal-types.html#inference), there are a couple different solutions that seems pretty straightforward.

```
// Works out of the box. The overload defaults to a JSON format and returns the IFeature type.
const requestOptions = {
      url: serviceUrl,
      id: 42
};
getFeature(requestOptions)
    .then(response => {
          console.log(respone.attributes)
    }
```
```
// Throws a compile error because of the string inference.
const requestOptions = {
      url: serviceUrl,
      id: 42,
      params: {
        f: 'geojson'
      }
};
getFeature(requestOptions) <-- Error thrown here like example in previous comment.
    .then(response => {
          console.log(respone.attributes)
    }
```
```
// After a minor tweek to ensure you are inferring the right format type.
const requestOptions = {
      url: serviceUrl,
      id: 42,
      params: {
        f: 'geojson' as 'geojson'   <-- assert the string as the 'geojson' type. Could be 'json' as 'json' too.
      }
};
getFeature(requestOptions) <-- No more error.
    .then(response => {
          console.log(respone.attributes)  <-- A helpful error here, because .attributes does not exist on a GeoJSON feature.
    }
```

All of that being said, if you all would rather just return a union type, I can roll these updates back and go for that. I think you would lose the helpful type errors (like the missing attributes property in the third example) however. 

As far as the dependency goes, you make a very good point. Based on the comment [here](https://github.com/Microsoft/types-publisher/issues/81#issuecomment-234051345) I've moved it to a dependency of the package itself so that they types are installed for those who are using the library with TS.

There are still some issues with Karma and the GeoJSON typings that I need to looks at. It seems something with it trying to read the type definition file is causing Karma to error out on the builds, even though the test are passing with Jasmine.

tomwayson:
Thank you for taking the time to provide such detailed info @itrew.

To be clear, you've done awesome work here, and I don't want to you rollback anything yet.

My problem is w/ TS. This BS just makes me never want to use string literals. `f: 'geojson' as 'geojson'` is just dumb. TS people need to be shown that whenever they are touting the elegance of their system which "only makes you provide types when they're needed." They should be forced to include a slide showing that in every presentation on TS - "Also, there will be times when you need to do something as stupid as: `f: 'geojson' as 'geojson'`".

I'm guessing we can't do this w/o the string literals, b/c there's no way for TS to infer which type it should return?

To me, _personally_, the helpful type error (`console.log(respone.attributes)  <-- A helpful error here, because .attributes does not exist on a GeoJSON feature.`) is not worth having to reply to a single issue w/ "yeah, the problem is, you need to use `f: 'geojson' as 'geojson'`. sorry."

B/c the only downside of returning a union type is that they don't get intellisense on `response.` unless they explicitly cast it to one of the known return types, right? If so, I'm OK w/ that. The benefits of TS are opt-in, but it doesn't put some BS type error in your face, b/c at the end of the day, _it's perfectly fine to call this fn w/ `f: 'geojson'`_.

That's just my 2 cents, but I'm curious how others who likely actually consume this library from a TS app (I typically consume it from JS) feel.

If we decide to go w/ just returning union types, if it's easier to open another PR rather than roll back what you have, that's fine.

itrew:
I hear ya. It's been a little frustrating dealing with the string literals, but in a way they make sense. After reading [this](https://github.com/Microsoft/TypeScript/issues/11270#issuecomment-250957528) comment on a similar issue on the TS repo, I can see why it acts the way it does.

After some more thought, I think I'm going to scrap this PR for now and go with the union types. Another good reason to start fresh is that I just did some testing and apparently both the `getFeature` and `queryRelated` methods are unable to return a geojson. The feature endpoint doesn't support the geojson format, and the query related endpoint returns an error when attempting to retrieve the records in geojson. So I won't add the geojson typings to those methods in a new PR.

itrew:
Closing for now. Moving the discussion back to #343

PR #350, opened 2018-10-08
feat(users): add user invitation functions

mjuniper:
AFFECTS PACKAGES:
@esri/arcgis-rest-users

### DIFF
--- a/packages/arcgis-rest-users/src/invitation.ts
+++ b/packages/arcgis-rest-users/src/invitation.ts
@@ -90,6 +90,12 @@ export function acceptInvitation(
   return request(url, requestOptions);
 }
 
+/**
+ * Decline an invitation.
+ *
+ * @param requestOptions - Options for the request
+ * @returns A Promise that will resolve with the success/failure status of the request
+ */
 export function declineInvitation(
   id: string,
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-users/src/invitation.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Get an invitation for a user by id.
+ *
+ * @param requestOptions - options to pass through in the request
+ * @returns A Promise that will resolve with the invitation
+ */
+export function getUserInvitation(
+  id: string,
+  requestOptions: IUserRequestOptions
```
jgravois: assuming that all these operations _require_ auth, my recommendation would be:

```ts
export interface IInvitationRequestOptions extends IUserRequestOptions {
  invitationId: "string"; // mandatory param with more explicit name
}

// same for accept/decline
export function getUserInvitation (
  requestOptions: IInvitationRequestOptions
)
const url = `${portalUrl}/community/users/${username}/invitations/${requestOptions.invitationId}`;
```

see #78 for more info. i still **really** need to write some CONTRIBUTOR doc for this.

PR #365, opened 2018-10-13
ensure 'first class citizen' geocode request parameters are passed through

jgravois:
if you aren't just passing a string to be geocoded, currently this is how you do it:
```js
geocode({
  params: {
    address: "123 Main St."
  }
})
```
the tweaks here ensure you can do this too
```js
geocode({ address: "123 Main St." })
```
### related:
- [X] moved utility `appendCustomParams()` method that was previously in @esri/arcgis-rest-feature-service into @esri/request

### unrelated:
- [X] made a small tweak to stop the tslint precommit hook from running twice
- [X] added the 20 folks (:100:) that have already contributed to this project to the package.json

cc/ @gavinr

### DIFF
--- a/packages/arcgis-rest-feature-service/src/helpers.ts
+++ b/packages/arcgis-rest-feature-service/src/helpers.ts
@@ -7,13 +7,6 @@ import {
   IGeometry,
   ISpatialReference
 } from "@esri/arcgis-rest-common-types";
-import { IRequestOptions } from "@esri/arcgis-rest-request";
-
-import { IQueryFeaturesRequestOptions } from "./query";
-import { IAddFeaturesRequestOptions } from "./add";
-import { IUpdateFeaturesRequestOptions } from "./update";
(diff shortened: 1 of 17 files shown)


packages/arcgis-rest-feature-service/src/helpers.ts
```diff
@@ -50,31 +43,3 @@ export interface IEditFeaturesParams {
    */
   rollbackOnFailure?: boolean;
 }
-
-/**
- * Used internally by the package to ensure that first order request options are passed through as request parameters.
- */
-export function appendCustomParams(
```
tomwayson: is the idea that this is not a breaking change b/c it was only used internally by this package?

jgravois: good catch.

we don't have to duplicate the method to salvage backwards compatibility. re-exporting directly from @esri/request is sufficient.

3166ee2

PR #366, opened 2018-10-15
chore(docs): add in missing types library from npm install notes

deeg:


jgravois:
better late than never @coveralls!

PR #368, opened 2018-10-15
chore(docs): fix suggest & geocode autocomplete example

deeg:


jgravois:
those snippets were actually more screwed up than i thought.

i fixed them (and simplified them) in 6157ddf00297c0a4b5a612c69ec983192242c104. thanks for the 🦅👀!

Issue #373, opened 2018-10-29
expose new responseType option in IRequestOptions


sfabijanski:
I'm working with Node 8.6.0 in a linux environment. I was trying to request an image export from an image server when I ran into an error: "error: TypeError: response.blob is not a function." When the format is set to *image* rather than resolve with the image buffer, the request throws an error.

I've traced the error down to the format setting. When the parameters are set and format is *image* the promise should resolve and the image buffer should be returned, but it does not, and throws an error in **request.ts**. Based on the structure of that file, and the switch statement on *params.f* I'm guessing that the 'zip' format will also fail.

Investigating a bit further I found #bitinn/node-fetch/issues/78  which references response.blob being browser only. 

I put together a quick test:
``` javascript
require("isomorphic-fetch");
require("isomorphic-form-data");

const { request } = require("@esri/arcgis-rest-request");
const imageServiceURL = 'https://geodata.md.gov/imap/rest/services/Imagery/MD_SixInchImagery/ImageServer';

// if f='json' this will work in Node
// when f='image' request fails with 'response.blob is not a function' error
const params = {
  bbox: '-8529110,4764085,-8528610,4764085',
  bboxSR: '102100',
  size: '400,400',
  imageSR: '102100',
  format: 'png8',
  pixelType: 'U8',
  noDataInterpretation: 'esriNoDataMatchAny',
  interpolation: '+RSP_BilinearInterpolation',
  f: 'image'
};

request(imageServiceURL + '/exportimage', { httpMethod: 'POST', params })
.then( result => {
  console.log({result});
})
.catch( error => { console.log({error})});

```

jgravois:
thanks for taking the time to write up what you found. for the time being you might consider using node-fetch@^2.0.0

```js
const fetch = require("node-fetch");
request(imageServiceURL + '/exportimage', { 
  httpMethod: 'POST', 
  params,
  fetch
})
```

related: https://github.com/matthew-andrews/isomorphic-fetch/pull/157

tomwayson:
alternatively you could try https://github.com/developit/unfetch/tree/master/packages/isomorphic-unfetch

jgravois:
see https://github.com/Esri/arcgis-rest-js/pull/374#issuecomment-435133941

https://developer.mozilla.org/en-US/docs/Web/API/Response#Methods

sfabijanski:
[...]

PR #374, opened 2018-10-29
a few tweaks to make the express demo app more plug and play

jgravois:
AFFECTS PACKAGES:
@esri/arcgis-rest-request

in #373 @sfabijanski discovered that isomorphic-fetch does _not_ [yet](https://github.com/matthew-andrews/isomorphic-fetch/pull/157) implement `response.blob()`.

~~the proposed tweak here adds a simple check within `@esri/arcgis-rest-request` and passes back the raw response to Node.js clients so they can work with a stream instead.~~


jgravois:
another option (and perhaps preferable) would be to just start recommending folks use polyfill for Node.js that uses node-fetch@^2.0.0 under the hood.

tomwayson:
@jgravois looks like isomorphic-unfetch is up to date: https://github.com/developit/unfetch/blob/1670b30b541ee9209c58653c550cf07489609cb9/packages/isomorphic-unfetch/package.json#L15

sfabijanski:
@jgravois I don't think returning the stream is the preferred. To make the library work consistently between browser JS and Node, the request should return the raw buffer--not a stream. I achieved the desired result--the image buffer--by changing the request format from 'image' to 'json', then pulling out the URL to the image from the *href* property on the response and going back to the server with the *request-promise* library, eg `requestPromise( {url: img.href,  encoding: null})`. This approach returns a buffer but requires two trips to the server to get the desired image. 

Perhaps isomorphic-unfetch, mentioned by @tomwayson, provides easier, direct access to the buffer?

sfabijanski:
Did some research and quick hacking and this snippet provides the expected result:
``` javascript
case "image":
    return response.buffer().then(b => b);
```

patrickarlt:
@jgravois I see a few options here, but the goals would be to:

1. We can't break existing browser based code that relies on responding with a `Blob` without a major version update.
2. We should provide as close to the same API as possible in both Node and browsers.
3. Since Node and Browsers ARE different and have different preferences (streams vs blobs) we should try to account for that.

Based on this there are a few options

1. Just recommend that people install something that implements `blog` using `node-fetch@^2.0.0` under the hood. Leave the existing code as it with no changes and close this PR.
2. Implement something like https://github.com/bitinn/node-fetch/issues/78#issuecomment-184759751 inside `polymorphicBlob` that reads the whole stream into a string and then converts that string to an array buffer and respond with that. This means that existing browser code would still work but it would mean that browsers would get a `Blob` and Node users would get a `Buffer` or `string`.
3. Implement a `responseType` option on  `IRequestOptions` that would allow users to control the response type independent of `params.f` for example:
    ```js
    require("isomorphic-fetch");
    require("isomorphic-form-data");
    const fs = require("fs");
    const { request } = require("./packages/arcgis-rest-request/dist/node/request");

    const url = 'https://geodata.md.gov/imap/rest/services/Imagery/MD_SixInchImagery/ImageServer/exportImage';

    const params = {
      bbox: '-8529110,4764085,-8528610,4764085',
      bboxSR: '102100',
      size: '400,400',
      imageSR: '102100',
      format: 'png8',
      pixelType: 'U8',
      noDataInterpretation: 'esriNoDataMatchAny',
      interpolation: '+RSP_BilinearInterpolation',
      f: 'image'
    };

    request(url, { params, responseType: 'stream' })
      .then( response => {
        // response is a readable stream
      })
      .catch( error => { console.log({error})});

    request(url, { params, responseType: 'buffer' })
      .then( response => {
        // response is a node buffer, uses `response.buffer` which is non standard under the hood
      })
      .catch( error => { console.log({error})});

    // no responseType indicates to use the existing behavior based on params.f to keep backward compat
    request(url, { params })
      .then( response => {
        // response is a blob which is an error unless node-fetch>2.0.0
      })
      .catch( error => { console.log({error})});
    ```
   values for `responseType` would be `json`, `text`, `arrayBuffer`, `blob` which are from the fetch spec, and `buffer` (non standard from `node-fetch`) and `stream` (non standard).

---

Personally I'm leaning toward by 1st or 3rd option.

jgravois:
option 1 sounds fine to me. `isomorphic-unfetch` definitely appears to give Node.js developers what they need.

```js
require("isomorphic-unfetch");

request(url, { params })
  .then( response => {
    return response.buffer()
  })
    .then(buffer => {
      fs.writeFile("foo.png", buffer, (err) => {});
    })
```

after confirming that our Node.js demos function with `isomorphic-unfetch`, i pushed another commit to:
* revert my previous suggestion
* use `isomorphic-unfetch` in our own demos
* swap it in as the suggested dependency in our guide docs.

i did _not_ update our [jasmine helper](https://github.com/Esri/arcgis-rest-js/blob/41ac46c3d0cb0e9b2876d65b641ac7d0e0dcc6db/support/test-helpers.js), because we'll have to upgrade fetch-mock@^7.0.0 at the same time, and doing that wrecks all sorts of havoc.

jgravois:
> I think the need to be able to override the response type has come up before.

yes. specifically we need a mechanism to allow developers to avoid passing the `f` parameter entirely _and_ declare the response type.

sfabijanski:
The example provided by @jgravois ,
``` javascript
require("isomorphic-unfetch");

request(url, { params })
  .then( response => {
    return response.buffer()
  })
    .then(buffer => {
      fs.writeFile("foo.png", buffer, (err) => {});
    })
````
Still does not work in a Node 8.6 environment. The call to *response.buffer()* still throws an error, so the response object is still not consistent between environments.

**Comment by jgravois** [comment]
> The example provided by @jgravois ... does not work

ugh. my code snippet from yesterday assumed this API was still passing back the raw response.

i spent awhile researching this morning and couldn't figure out how to save a `Blob` to disk in Node.js. because of this i'm switching my vote to a modified version of @patrickarlt suggestion 3). 

if we stick with exposing **only what's present in the fetch spec**, Node.js devs could do this:
```js
request(url, { 
  params,
  responseType: "arrayBuffer"
 })
  .then( response => {
    fs.writeFile("foo.png", new Buffer(response));
  })
```

sfabijanski:
I also spent considerable time trying to do anything with the returned *Blob*, all without success. Eventually I found [this StackOverflow answer](https://stackoverflow.com/questions/52576973/how-to-convert-fetchapi-parsed-blob-results-to-base64) which confirmed my own suspicions regarding the blob.  I think option three is a reasonable approach but realize that might take some time to happen.  

I also think some form of inspection of the globals and determining if this is running in Node or a browser and returning *response.buffer()* when Node would  be enough to satisfy the needs of Node development (at least in the short-term).

I'm fairly new to the concept of streams in Node, even though my last couple of day's research show I've been working with them regularly in various forms. [This](https://medium.freecodecamp.org/do-you-want-a-better-understanding-of-buffer-in-node-js-check-this-out-2e29de2968e8) provides a fairly straightforward  explanation. Buffers allow Node to work on streams.

jgravois:
i've reverted the changes i suggested related to `isomorphic-unfetch`.

we can continue to discuss exposing additional fetch options in  #373

tannerjt:
I have a module for creating backups from feature service endpoints.  It currently uses the core HTTPS module for streaming feature service geojson directly to disk and avoids storing the buffer into memory.

@jgravois mentioned potentially using arcgis-rest-js and referenced this issue ticket.

I like @patrickarlt's option (3) of explicitly requesting response type as a stream, independence of `params.f`.  I would assume the best method would be to return a readable stream, which is allowed by the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams)

For example, from` arcgis-rest-request` you could return:

```JavaScript
const fetch = require('node-fetch')

function request(path) {
  return new Promise((resolve, reject) => {
    fetch(path).then((resp) => {
      //resp.body is a ReadableStream
      resolve(resp.body)
    })
  })
}

module.exports = request
```

Then it can be consumed as a stream
```JavaScript
const request = require('request')
const fs = require('fs')
const outF = fs.createWriteStream('./outF.json')
const path = 'http://navigator.state.or.us/arcgis/rest/services/Projects/Prep_TransCams/MapServer/3?f=pjson'

request(path).then((resp) => {
  resp.pipe(outF)

  resp.on('data', (chunk) => {
    console.log('Got Chunk', chunk)
  })

  resp.on('end', () => { console.log('Finished') })
})
```

tomwayson:
So to meet @tannerjt's we'd need to at least add a new option like `responseType` to let consumers override the default behavior based on `params.f`. 

Beyond that I'm not fully clear what is needed. If `responseType === 'stream'`, we return `response.body`, regardless of `params.f`?

tannerjt:
Would it make sense to just add `stream?: boolean`  the `IRequestOptions` interface.  If true, return `response.body` before the switch statement that checks for `params.f`.

jgravois:
> Would it make sense to just add `stream?: boolean` the `IRequestOptions` interface

@patrickarlt's original suggestion is more 'forward looking'.

```
responseType === 'stream'
```

you're correct about the logic though. If true, return `response.body` before the switch statement that checks for `params.f`.

tannerjt:
From @patrickarlt's `responseType` suggestions:

> 3\. values for `responseType` would be `json`, `text`, `arrayBuffer`, `blob` which are from the fetch spec, and `buffer` (non standard from `node-fetch`) and `stream` (non standard).

I'm not sure any of these but `stream` are needed.  For requests that expect a `promise` to be returned, defining `params.f` will return the correct anticipated `json`, `text`, or `blob`.  

Both [node-fetch](https://www.npmjs.com/package/node-fetch) and regular [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Body/body) support streaming bodies.  When there are cases where a `stream` is needed (memory management for large responses), there would only really be two possible options `stream` or `buffer`.

`buffer` (node-fetch) and 'arrayBuffer` (fetch) will both read the entire content into memory first, which negates the memory advantage of using streams.  I guess it's possible that there are scenarios where an someone would want the entire binary ([convert jpg to png maybe?](https://stackoverflow.com/questions/11554006/javascript-arraybuffer-whats-it-for), but is this a likely scenario?

Based on this, I propose 2 options:

1. Instead of `responseType`, just have an optional boolean `stream` to `IRequestOptions`.
2. If `buffer` is really a desired, limit `responseType` to only two options; `stream` or `buffer` (or instead `raw` like you mentioned).  If `buffer` is requested, the entire binary would be returned (either as buffer or arrayBuffer) in the promise.

Also, please correct anything I may have gotten wrong above.  I'm no expert in this arena and am figuring this out myself as I go along.

jgravois:
@tannerjt is right that we're already handling the majority of response types appropriately internally.

i'd vote for a third option though.

```ts
{
  // ...
  rawResponse?: boolean;
} as IRequestOptions
```

if the value is truthy we'd short-circuit our params.f check entirely and just pass back the response without manipulating it at all.

this is analogous to a helpful back door @patrickarlt built into [esri-leaflet](https://github.com/Esri/esri-leaflet/blob/1ec2167ed7ea13d67a8536bb62c49a04930a470e/src/Tasks/Query.js#L141) long ago. it'd solve @tannerjt's use case _and_ probably others we haven't even dreamed up yet.

tomwayson:
I like that... but...

@jgravois I assume you mean that if `rawResponse` is truthy, `request()` would return the reponse synchronously, right? I wonder how that would affect the whole deck of cards that assume they are going to get a Promise.

tomwayson:
I mean, we'd have the same problem if we returned `response.body` as @tannerjt suggests.

jgravois:
i'm not sure i follow you. currently we're calling the intermediate promise below _inside_ the API.

```js
fetch(url)
  .then(response => response.json()) // response.text() response.blob(). 
  .then(payload)
```
so to our consumers, they shouldn't notice a difference at all.

```js
request(url, { params: { f: "json" } })
  .then(response) // already json

request(url, { rawResponse: true })
  .then(response) // can call response.json(), response.blob() or access a readable stream
```

tannerjt:
Hey @jgravois & @tomwayson, please take a look at the pull request #457.  I opted to stick with limiting to stream output because buffer and arrayBuffer appeared to be too convoluted for isomorphic support and testing.  @jgravois is correct that consumers will not notice a difference, and the request will always honor the `params.f` format.  If `stream` is requested, the response.body will return a [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) in the `promise`.  I also added a test to make sure either `Readable` or `ReadableStream` is returned and a demo to illustrate streaming a geojson feature service response directly to file.

tomwayson:
duh... ignore me.

Issue #47, opened 2017-09-20
Design API for @esri/arcgis-rest-routing module


patrickarlt:
Design API with support for the routing services https://resources.arcgis.com/en/help/arcgis-rest-api/#/Overview_of_network_analysis_services/02r30000001s000000/ probably using https://github.com/Esri/esri-leaflet-routing as a base.

PR #382, opened 2018-11-07
added routing module

gavinr:
Beginning of the routing module and the the `solveRoute` function, covering the [solve endpoint](https://developers.arcgis.com/rest/network/api-reference/route-synchronous-service.htm) of the REST API.

#47


packages/arcgis-rest-routing/src/solveRoute.ts
```diff
@@ -0,0 +1,124 @@
...
+ *        }
+ *      ],
+ *      authentication: session
+ *   })
+ *   .then((response) => {
+ *     response.routes.features; // => {routes: {features: [{attributes: { ... }, geometry:{ ... }}]}}
+ *   });
+ * ```
+ *
+ * @param address String representing the address or point of interest or RequestOptions to pass to the endpoint.
```
jgravois:
```
@param requestOptions Options to pass through to the routing service.
```

jgravois:
i decided to kill two birds with one stone and reorganize a few of the utility methods and interfaces you copied from the 'geocoder' package and put them in 'common-types' so they can be shared instead.

this should resolve the problem @MikeTschudi reported in https://github.com/ArcGIS/arcgis-clone-js/pull/3#issuecomment-436781419 as well.

can you sanity check me please @tomwayson?

patrickarlt:
@jgravois should we keep `common-type` for just types and make a separate `shared-utils` package for shared utility methods?

jgravois:
>  I think we should go ahead and create @esri/arcgis-rest-common, and start by moving these fns there.

that works for me. i'd be more inclined to go with @patrickarlt's suggestion but the types are already automatically stripped from built `umd` output and i'm hopeful we could do the same in `cjs`. if so, and we can make sure only d.ts files are created for types in the `esm` build, it'd make for another tidy little package.

to do: 

- [ ] revert the changes in my last commit and move the shared functions into a _second_ new package called `@esri/arcgis-rest-common`
- [ ] make sure `routing` includes `common` as a dependency, (but not geocoder)
- [ ] refactor geocoder to use `common` at `v2.0.0`
- [ ] figure out how to help @MikeTschudi with his unrelated problem

tomwayson:
just curious, why not refactor geocoder to start using the fns in 'common' now? is that breaking?

jgravois:
> why not refactor geocoder to start using 'common' now? is that breaking?

exactly.

jgravois:
okay, i finished refactor (2) and rebased to clean up my own tail of the commit history.

PR #387, opened 2018-11-13
chore: include raw /src in common-types npm package

jgravois:
to help downstream karma-typescript users sidestep errors when the bundler doesn't find any JavaScript.

ref: https://github.com/ArcGIS/arcgis-clone-js/pull/8

personally i don't see a downside to doing this for **every** package.

AFFECTS PACKAGES:

### DIFF
--- a/packages/arcgis-rest-common-types/package.json
+++ b/packages/arcgis-rest-common-types/package.json
@@ -6,7 +6,8 @@
   "author": "",
   "license": "Apache-2.0",
   "files": [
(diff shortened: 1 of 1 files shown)


tomwayson:
So something like `import { IFeature, IUser } from '@esri/arcgis-rest-common-types';` won't work in karma-typescript, but w/ this PR it will just start working? It seems like that `import` would need to be changed to point to the files under `src` since we're not changing `package.types`, right?

PR #389, opened 2018-11-14
Add Code to Request.ts to set Referer

will7692:
If Window is undefined set referer to the same value generate token uses if window is undefined
@jgravois


jgravois:
thx @will7692! did you confirm this resolves things on your end?

will7692:
@jgravois yes i have tested this, and it does solve the token issue

PR #394, opened 2018-11-19
refactor(decodeValues): short circut out of decoding values if there are no CVD fields

tomwayson:
This seemed like potentially hot code, esp as we look at how we are implementing it in cedar/the Hub, so I made the following changes:
- just return the original response if there are no CVD fields to decode
- don't clone the response before looping through all features/attributes

AFFECTS PACKAGES:


tomwayson:
@jgravois 

To explain why I made these changes. I started thinking about how we're going to use this in the Hub. As we discussed the other day, whenever we create charts/tables in the Hub we have direct access to the layer's metadata (fields), and I figured we'd just update the code there to _always_ call `decodeValues()` and pass those in. Then I thought that having CVD fields is more the exception than the norm, so there's no reason to clone the response and then loop through the cloned features if there are no CVD fields (the norm).

tomwayson:
Thinking specifically about the chart card in the Hub, we may want to take this a step further and `export` `extractCodedValueDomains()` and allow passing in `domains` as an alternative to `fields` to `decodeValues()`. 

The reason is that for the chart card, we _serialize_ the [card settings](https://github.com/ArcGIS/opendata-ui/blob/c0e447c9ae38cec5aac6c172ed3768a10c57797c/packages/ember-arcgis-layout-cards/addon/utils/chart/card-settings-to-definition.js
) needed to [create a definition](https://github.com/ArcGIS/opendata-ui/blob/c0e447c9ae38cec5aac6c172ed3768a10c57797c/packages/ember-arcgis-layout-cards/addon/utils/chart/card-settings-to-definition.js). So now we're going to have to add either `fields` or `domains` to those settings, and they will get stored in the page item JSON. We definitely don't want to serialize and store _all_ fields. So at the very least, the Hub would need to filter and serialize _only_ CVD fields. I say we cut out the middleman and let the Hub call `extractCodedValueDomains()` and then it can just pass and serialize the domains (if any).

Thoughts?

Issue #391, opened 2018-11-14
Possibly change feature add, update and delete params to match REST API at some point


COV-GIS:
It's personal preference but `adds`, `updates` and `deletes` params really bother me for not being the REST params, e.g. `features` and `objectIds` respectively. I never fully understood the JS API having different param names from the REST API for queries and the like. I get it, `outSpatialReference` is more descriptive than `outSR`, but the rest of the param names here are REST API driven. I just like being able to look at the REST API docs and know I can pass any param as is.

I thought about it at the time the functions were added but didn't mention it. This would be a breaking change. So may never happen. But if there ever is a v2 candidate or other opening to change the param names I'd be happy to submit changes.

jgravois:
we actually already discussed this in https://github.com/Esri/hub.js/issues/58

a PR that introduces `features` and `objectIds` and adds a [deprecation warning](https://github.com/Esri/arcgis-rest-js/blob/84ad2cdc3a3336fa2c33f7eb565aa359259c6871/packages/arcgis-rest-geocoder/src/helpers.ts#L72-L74) when parsing what we already have would be _most_ welcome.

COV-GIS:
I'm not the only one then.

Will do.

COV-GIS:
https://github.com/Esri/arcgis-rest-js/pull/396

PR #341, opened 2018-09-27
docs(bug): ensure sticky links are flagged as active in production

jgravois:
a less insane solution than #334 (to resolve #284)

PR #396, opened 2018-11-19
Fix crud param names

COV-GIS:
Per https://github.com/Esri/arcgis-rest-js/issues/391

### DIFF
--- a/packages/arcgis-rest-feature-service/src/add.ts
+++ b/packages/arcgis-rest-feature-service/src/add.ts
@@ -5,7 +5,8 @@ import { IFeature } from "@esri/arcgis-rest-common-types";
 import {
   request,
   IRequestOptions,
-  appendCustomParams
+  appendCustomParams,
+  warn
 } from "@esri/arcgis-rest-request";
 import { IEditFeaturesParams, IEditFeatureResult } from "./helpers";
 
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-feature-service/src/add.ts
```diff
@@ -26,7 +27,13 @@ export interface IAddFeaturesRequestOptions
...
    * Array of JSON features to add.
    */
-  adds: IFeature[];
+  features: IFeature[];
+  /**
+   * Deprecated. Please use `features` instead.
+   *
+   * Array of JSON features to add.
+   */
+  adds?: IFeature[];
```
jgravois:
```js
/**
* Deprecated. Please use `features` instead.
```
~~Array of JSON features to add.~~

these two will show up right next to each other in the API reference, so no need to repeat the meat of the description.

COV-GIS:
Hey @jgravois. I got a new work station. Build and test work just fine but I'm having issues running `docs:serve` to check the docs. All the tasks run w/o error. `typedoc.json` is created. The assets are all copied. But no html files are being generated. Any thoughts on why?

jgravois:
> no html files are being generated. Any thoughts on why?

yup. acetate doesn't write output files to disk when you're running the dev server. they're only in memory ✨.

its a bit of a gotcha, but until https://github.com/patrickarlt/acetate/pull/74 lands, we all have to tack `/arcgis-rest-js/` onto the url thats spawned manually.

[`http://localhost:3000/arcgis-rest-js/`](http://localhost:3000/arcgis-rest-js/)

COV-GIS:
It must have just opened `http://localhost:3000/arcgis-rest-js/` on my old machine. Thanks.

jgravois:
> It must have just opened http://localhost:3000/arcgis-rest-js/ on my old machine. Thanks.

nah, its a regression i introduced in #341.

if you could make the [breadcrumb](https://github.com/Esri/arcgis-rest-js/blob/master/CONTRIBUTING.md#building-the-documentation-site-locally) i already dropped more explicit, it would be _sincerely_ appreciated.

COV-GIS:
That looks good to me @jgravois. Had I seen that I would have known.

Issue #398, opened 2018-11-20
attempt to clean up malformed portal urls and log a warning


jgravois:
- [x] if the url contains trailing whitespace or doesn't end in a trailing slash, append it
- [ ] if the url is missing a protocol, append `https://` (and warn)
- [ ] if the url is missing `/sharing/rest`, append (and warn)

see #288 for more info

jgravois:
PRs for any of the improvements above would be welcome, but there's no need to track an open issue.

PR #400, opened 2018-11-26
add utility method to trim whitespace and trailing slashes from input urls

jgravois:
halfway toward resolving #398 

AFFECTS PACKAGES:
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-request

### DIFF
--- a/packages/arcgis-rest-geocoder/src/bulk.ts
+++ b/packages/arcgis-rest-geocoder/src/bulk.ts
@@ -1,7 +1,7 @@
 /* Copyright (c) 2017-2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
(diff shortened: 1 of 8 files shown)


tomwayson:
Obviously there are many other places we could/should use `cleanUrl()` when appending to the URL besides the places you've taken care of in this PR. Is your idea that we do the others over time, i.e. on a package by package basis and split up the work?

jgravois:
> Obviously there are ~~many~~ other places we could/should use cleanUrl() when appending to the URL 

this PR actually covers the majority (items, users, groups, sharing, feature-service-admin) because the method is called internally by `getPortalUrl()`.

that said, for sure there are still holdouts that we can/should get to in time.

Issue #403, opened 2018-11-27
update feature-service fns to use cleanUrl()


tomwayson:
I think the feature-service package is probably most significant area where `cleanUrl()` still needs to be be applied (see https://github.com/Esri/arcgis-rest-js/pull/400#issuecomment-441808360)

COV-GIS:
Hey @tomwayson. I can add `cleanUrl()` in feature service and PR if you'd like.

jgravois:
closed via #404 :tada

PR #404, opened 2018-11-27
chore: clean feature service urls

COV-GIS:
Per https://github.com/Esri/arcgis-rest-js/issues/403


COV-GIS:
I was in there any way @tomwayson. Two minute update. :)

PR #418, opened 2018-12-10
Reexported IParams and joined fetch common types

MikeTschudi:
1. Re-exported `IParams` to solve transpiler complaint from packages/arcgis-rest-request/dist/esm/utils/check-for-errors.d.ts that `IParams` was not available from request.ts

2. Combined response cases in function `request` because 1) they are handled identically, and 2) the "zip" variant doesn't work with arcgis.com (one gets a `status: 400  Format not allowed for this Item Type` error), so it seemed misleading to include it in the `switch` statement.

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -8,6 +8,8 @@ import { ArcGISRequestError } from "./utils/ArcGISRequestError";
 import { IRetryAuthError } from "./utils/retryAuthError";
 import { HTTPMethods, IParams, ITokenRequestOptions } from "./utils/params";
 
+export { IParams } from "./utils/params";
+
 /**
  * Authentication can be supplied to `request` via [`UserSession`](../../auth/UserSession/) or [`ApplicationSession`](../../auth/ApplicationSession/). Both classes extend `IAuthenticationManager`.
  * ```js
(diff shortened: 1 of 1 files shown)


jgravois:
> to solve transpiler complaint from packages/arcgis-rest-request/dist/esm/utils/check-for-errors.d.ts...

`utils/check-for-errors` was removed in https://github.com/Esri/arcgis-rest-js/commit/9b58c636f1e5fecb6336979bbd45235e68994452.
upgrading to `v1.14.3` _should_ resolve that error.

https://unpkg.com/@esri/arcgis-rest-request@1.14.3/dist/esm/utils/

MikeTschudi:
Thanks, @jgravois, I had not run `clean`. Updated request.ts to only have `switch` modification.

jgravois:
> Maybe something changed with the REST API?

i don't think so. the parameter only applies to [Package](https://developers.arcgis.com/rest/users-groups-and-items/items-and-item-types.htm) item types.

example: [Gemeentegrenzen 2014](http://www.arcgis.com/home/item.html?id=218ecc03273d4de184d99be8f22cc0a7) (layer package)

https://www.arcgis.com/sharing/rest/content/items/218ecc03273d4de184d99be8f22cc0a7/data?f=zip

PR #421, opened 2018-12-18
chore: pkg.browser > pkg.unpkg

### LINKED ISSUE #420: pkg.browser > pkg.unpkg
apparently unpkg.com moved away from relying on the `"browser`" field in a package.json file a _long time ago_.

https://github.com/unpkg/unpkg.com/issues/63

jgravois:
resolves #420 🚬 	

AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-common
@esri/arcgis-rest-feature-service-admin
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items
@esri/arcgis-rest-request
@esri/arcgis-rest-routing
@esri/arcgis-rest-sharing
@esri/arcgis-rest-users

### DIFF
--- a/packages/arcgis-rest-auth/package.json
+++ b/packages/arcgis-rest-auth/package.json
@@ -3,7 +3,7 @@
   "version": "1.14.3",
   "description": "Authentication helpers for @esri/arcgis-rest-*.",
   "main": "dist/node/index.js",
-  "browser": "dist/umd/auth.umd.js",
+  "unpkg": "dist/umd/auth.umd.js",
   "module": "dist/esm/index.js",
   "js:next": "dist/esm/index.js",
   "types": "dist/esm/index.d.ts",
--- a/packages/arcgis-rest-common/package.json
+++ b/packages/arcgis-rest-common/package.json
@@ -3,7 +3,7 @@
   "version": "1.14.3",
   "description": "Shared utility methods and types for @esri/arcgis-rest-js",
   "main": "dist/node/index.js",
-  "browser": "dist/umd/common.umd.js",
+  "unpkg": "dist/umd/common.umd.js",
   "module": "dist/esm/index.js",
   "js:next": "dist/esm/index.js",
   "types": "dist/esm/index.d.ts",
--- a/packages/arcgis-rest-feature-service-admin/package.json
+++ b/packages/arcgis-rest-feature-service-admin/package.json
@@ -3,7 +3,7 @@
   "version": "1.14.3",
   "description": "Feature service administration helpers for @esri/arcgis-rest-request",
   "main": "dist/node/index.js",
-  "browser": "dist/umd/feature-service-admin.umd.js",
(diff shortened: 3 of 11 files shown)


tomwayson:
wait, what? please 'splain. is this a thing?

jgravois:
> is this a thing?

# TL;DR

yes. if its good enough for [D3](https://github.com/d3/d3/blob/3e2bbb506b536fdf13bfdf6a106d1c16b4859666/package.json#L19), its good enough for me.

## just TL

unpkg.com can be used to serve up _any_ file in an npm package, but historically when none are specified explicitly in a url, it responded with whatever was specified in the `browser` field (or `main` as a fallback).

this led to complaints (see https://github.com/unpkg/unpkg.com/issues/25) that this is an unnecessary collision with bundlers like browserify that would expect Common JS in that slot so now `"unpkg" > "main" > "browser"` and someday `"browser"` could be dropped.

> post script. as of yesterday **no** automatic redirects are working because of a **different** bug (see https://github.com/unpkg/unpkg.com/issues/156).

> post post script. this is just trivia, but [esri-leaflet plugins](https://github.com/unpkg/unpkg.com/issues/63#issuecomment-345122238) make up the overwhelming majority of unpkg traffic still relying on the `browser` fallback.

tomwayson:
OK, I missed the ref to #420 (I've only had 1 ☕️ ) and it's ref to https://github.com/unpkg/unpkg.com/issues/63

So yeah, I think we should add `unpkg`, but not sure if we should remove `browser`, and if we don't, not sure what it should point to.

I'm having a hard time thinking of `pkg.browser` purely as a place to specify a bundler entry point that includes browser-only shims. It appears from that issue that's how browserify (R.I.P.) used it, but it's unclear to me if webpack does the same. There are other uses though. [NPM says it's a flag to indicate that your code expects to be run in a browser](https://docs.npmjs.com/files/package.json#browser), but unsurprisingly doesn't weigh in on whether it's expected to be UMD, or if dependencies are expected to be bundled.

 In cases like this, I always ask, what would Bostock do, and it looks like you may have already asked that, b/c he doesn't include a `browser` field:

https://github.com/d3/d3/blob/3e2bbb506b536fdf13bfdf6a106d1c16b4859666/package.json#L18-L20

At the end of the day, I think we should keep `browser` as is if for no other reason than it's a breaking change, and I'm pretty sure some of our own Ember code relies on it.

jgravois:
> I think we should keep browser as is if for no other reason than it's a breaking change, and I'm pretty sure some of our own Ember code relies on it.

AFAIK our ember code points straight the umds using their [explicit location](https://github.com/Esri/ember-arcgis-portal-services/blob/master/index.js#L32-L35 (edited) ). 

the benefit of _dropping_ `browser` is that it helps webpack users that _want_ to consume ES Modules avoid the footgun bostock mentions here: https://github.com/unpkg/unpkg.com/issues/63#issuecomment-326746759

Issue #174, opened 2018-04-16
Fetching token service url fails


richardhinkamp:
I need to export some data to an ArcGIS system with the REST API. I created a developer account and got the code working on my ArcGIS Online account fairly simple. When using the REST endpoint provided by our customer, things didn't work, so I tried to find out what went wrong:

Like this:

```
const URL = "https://<customer.domain>/arcgis/rest/services/<a>/<b>/FeatureServer/0"
const session = new UserSession({
    username: "<username>",
    password: "<password>",
});
let options = {
    httpMethod: "POST",
    authentication: session,
    params: {
        where: `FIELD1 = 12345`,
        outFields: "OBJECTID,FIELD1,FIELD2",
        },
    };
return request(URL + "/query", options);
```

Because of the authentication, in `request()` it calls `authentication.getToken(url)`.
I did not provide a portal url in the sesssion options, so it get's to `this.getTokenForServer(url)`. It requests "https://<customer.domain>/arcgis/rest/info". The response it gets is this:

```
{
    "currentVersion": 10.31,
    "fullVersion": "10.3.1",
    "soapUrl": "https://<customer.domain>/arcgis/services",
    "secureSoapUrl": null,
    "authInfo": {
        "isTokenBasedSecurity": true,
        "tokenServicesUrl": "https://<customer.domain>/arcgis/tokens/",
        "shortLivedTokenValidity": 60
    }
}
```

It tries to get the `owningSystemUrl` property to do a request to `owningSystemUrl + "/sharing/rest/info"` but this property is undefined. The RegExp check should probably fail, but since it's undefined it doesn't. The request fails ofcourse because `owningSystemUrl` is undefined, but it's not needed I guess because the `authInfo.tokenServicesUrl` is already available in the response above.

I tried providing a portal property with the `UserSession`. Using "https://<customer.domain>/arcgis" results in a request to "https://<customer.domain>/arcgis/generateToken" which gives a 404. Using "https://<customer.domain>/arcgis/tokens" as the portal (since portal + "/generateToken" results in the correct url) also fails, because in `getToken()` the RegExp fails and so `getTokenForServer()` is called which results in the case described above.

I'm using node v6.12.3, with ts-node v5.0.1 and typescript v2.7.2 with version 1.1.1 of @esri/arcgis-rest-auth + @esri/arcgis-rest-common-types + @esri/arcgis-rest-request + @esri/arcgis-to-geojson-utils

[...]

Issue #422, opened 2018-12-19
How should apps handle multiple credentials


tomwayson:
Unlike [IdentityManager](https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-IdentityManager.html), rest-js does not provide a means to handle multiple credentials, for example if:
- an app needs to connect to multiple servers using different credentials
- an app needs to connect to a server that is not federated w/ the current session's portal

Up until now, I think our idea is that it is up to the app to manage multiple instance of [IAuthenticationManager](https://esri.github.io/arcgis-rest-js/api/request/IAuthenticationManager/). For example to initialize multiple [UserSession](https://esri.github.io/arcgis-rest-js/api/auth/UserSession/)s and pass in a different `authentication` depending on the `url`.

However, I think there's an opportunity for us to handle this at the library level better than how consuming apps can. For example, we could create a new type of `IAuthenticationManager` called `IdentityStore` or `AuthenticationStore` that could hold multiple `UserSession`s and return the appropriate token based on the `url` passed to `getToken()`. Analogous to [IdentityManage's registerToken()](https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-IdentityManager.html#registerToken) it would have a method to register credentials ([IUserSessionOptions](https://esri.github.io/arcgis-rest-js/api/auth/IUserSessionOptions/)) for all requests that start with a given server's URL like `identityStore.add(sever: string, session: IUserSessionOptions): void`. Example: 

```ts
// create a single store that we'll pass to all requests
const identityStore = new IdentityStore();

// add credentials for working w/ our server
const ourServer = 'https://gissvr.ourorg.gov/arcgis/rest/services';
identityStore.add({
  server: ourServer,
  session: {
    // these would come from a dialog
    username: 'us',
    password: 'secret'
  }
});

// add credentials for working w/ their server
const theirServer = 'https://mapsvr.theirorg.com/arcgis/rest/services';
identityStore.add({
  server: theirServer,
  session: {
    // these would come from a dialog
    username: 'them',
    password: 'alsoSecret'
  }
});

// query our feature sever
const ourFeatureServerUrl = `{ourServer}/rivers/0`;
queryFeatures({
  url: ourFeatureServerUrl,
  // when identityStore.getToken() is called, it will return the token for our server
  authentication: identityStore
});

// query their feature sever
const theirFeatureServerUrl = `{theirServer}/streets/0`;
queryFeatures({
  url: theirFeatureServerUrl,
[...]

PR #423, opened 2018-12-20
feat: new UserSession tied to a non-federated ArcGIS Server instance

jgravois:
the proposed tweak to `UserSession` would enable developers to build up a UI of their own for users to enter credentials and access secure content from unfederated instances of ArcGIS Server.

```js
import { request } from "@esri/arcgis-rest-request";
import { UserSession } from "@esri/arcgis-rest-auth";

const authentication = new UserSession({
  username: "user1",
  password: "user1",
  server: "https://sampleserver6.arcgisonline.com/arcgis"
});

request('https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire_secure/MapServer', { authentication })
  .then(token)
```

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #174

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -1,5 +1,16 @@
-/* Copyright (c) 2017-2018 Environmental Systems Research Institute, Inc.
+/* Copyright (c) 2017-2019 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
+
+/**
+ * /generateToken returns a token that cannot be refreshed.
+ *
+ * oauth2/token can return a token *and* a refreshToken.
+ * up until the refreshToken expires, you can use it (and a clientId)
+ * to fetch fresh credentials without a username and password.
+ *
+ * the catch is that this 'authorization_code' flow is only utilized
+ * by server based OAuth 2 Node.js applications that call /authorize first.
+ */
 
 import * as http from "http";
 import {
@@ -101,6 +112,11 @@ export interface IOauth2Options {
   refreshTokenTTL?: number;
 
   /**
+   * An unfederated ArcGIS Server instance that recognizes the supplied credentials.
+   */
+  server?: string;
+
+  /**
    * The locale assumed to render the login page.
    *
    * @browserOnly
@@ -186,6 +202,11 @@ export interface IUserSessionOptions {
    * Duration (in minutes) that a refresh token will be valid.
    */
   refreshTokenTTL?: number;
+
+  /**
+   * An unfederated ArcGIS Server instance that recognizes the supplied credentials.
+   */
+  server?: string;
 }
 
 /**
@@ -519,6 +540,11 @@ export class UserSession implements IAuthenticationManager {
   public readonly refreshTokenTTL: number;
 
   /**
+   * An unfederated ArcGIS Server instance that recognizes the supplied credentials.
+   */
+  public readonly server: string;
+
+  /**
    * Hydrated by a call to [getUser()](#getUser-summary).
    */
   private _user: IUser;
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -323,8 +333,10 @@ export class UserSession implements IAuthenticationManager {
     this.tokenDuration = options.tokenDuration || 20160;
     this.redirectUri = options.redirectUri;
     this.refreshTokenTTL = options.refreshTokenTTL || 1440;
-
     this.trustedServers = {};
+    if (options.server) {
+      this.trustedServers[options.server] = null;
```
tomwayson: I _think_ I know why you're doing this, but probably a good idea to leave a comment here.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -760,7 +790,8 @@ export class UserSession implements IAuthenticationManager {
             params: {
               username: this.username,
               password: this.password,
-              expiration: this.tokenDuration
+              expiration: this.tokenDuration,
+              client: "referer"
```
tomwayson: Do we _always_ want to send `client=referer`? Do we need to add that to above call to `generateToken()` too?

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -722,25 +734,43 @@ export class UserSession implements IAuthenticationManager {
...
+           * be able to generate a token
+           */
+          if (!new RegExp(response.owningSystemUrl, "i").test(this.portal)) {
+            throw new ArcGISAuthError(
+              `${url} is not federated with ${this.portal}.`,
+              "NOT_FEDERATED"
+            );
+          } else {
+            return request(
+              `${response.owningSystemUrl}/sharing/rest/info`,
```
tomwayson: Maybe add a comment here to 'splain what this case is?

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -722,25 +734,43 @@ export class UserSession implements IAuthenticationManager {
...
+          }
+        } else if (
+          response.authInfo &&
+          this.trustedServers[root] !== undefined
         ) {
+          /**
+           * if its a stand-alone instance of ArcGIS Server that doesn't advertise
+           * federation at all and the root url is recognized, use its built in token endpoint.
+           */
+          return new Promise((resolve, reject) => {
```
tomwayson: Maybe just `return Promise.resolve({ authInfo: response.authInfo });`

tomwayson:
@jgravois

I believe we'd also need to make further changes if we wanted to support @dbouwman's [idea](https://github.com/Esri/arcgis-rest-js/issues/422#issuecomment-448778878)

> where the consuming app has the token (from a cookie, localstorage, some other auth flow managed by the app), but not the username/password?

Maybe set the token here?

https://github.com/Esri/arcgis-rest-js/pull/423#discussion-diff-243449223R338

packages/arcgis-rest-auth/src/UserSession.ts
jgravois: it was only _necessary_ in the case that i tweaked, but there's no harm in being consistent.

packages/arcgis-rest-auth/src/UserSession.ts
jgravois: 💡 	
👶

jgravois:
> where the consuming app has the token (from a cookie, localstorage, some other auth flow managed by the app), but not the username/password?

i've added another small tweak to support the use case you mentioned. this is what needs to be provided to invoke it:

```js
const authentication = new UserSession({
  server: "https://sampleserver6.arcgisonline.com/arcgis",
  token: "abc123",
  tokenExpires: new Date(1545415669763)
});
```

to do:
- [ ] get @patrickarlt's blessing
- [ ] refactor the trusted server check to compare only root urls

jgravois:
@patrickarlt given that you chimed in on #422 but not this PR i'll _assume_ you aren't **fundamentally opposed** to providing a mechanism to support making requests to non-federated, authenticated services.

i *know* you're busy with https://developers.arcgis.com tasks so if i'm wrong even a quick "hold your horses, i haven't had time to look at this yet and i'd like to", sometime this week would be welcome.

patrickarlt:
@jgravois I realized that you asked about this again ago but it all looks good to me. I figured at some point someone would ask for this, I just didn't want to do it.

jgravois:
thanks @patrickarlt!

i've rebased and pushed up some more changes to ensure test coverage and tie up the remaining loose ends. i'm happy with this if @tomwayson is.

Issue #314, opened 2018-09-11
allow developers to control how long an ApplicationSession token will be valid


jgravois:
i'm still ✈️lagged, but when i linted the code in #313 it pointed out that `duration` was declared but never used.

https://github.com/Esri/arcgis-rest-js/blob/6a17fe577b6c6f839d2667fd6d14099ab41f3bab/packages/arcgis-rest-auth/src/ApplicationSession.ts#L35-L39

PR #429, opened 2019-01-09
fix: expose option to customize expiration of ApplicationSession tokens

jgravois:
#428 put a 🔥 under my butt.

the approach i'm using is currently undocumented, but i've asked the team to rectify _that_ situation.

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #314

### DIFF
--- a/packages/arcgis-rest-auth/src/fetch-token.ts
+++ b/packages/arcgis-rest-auth/src/fetch-token.ts
@@ -39,7 +39,8 @@ export function fetchToken(
       token: response.access_token,
       username: response.username,
       expires: new Date(
-        Date.now() + (response.expires_in * 60 * 1000 - 60 * 1000)
+        // convert seconds in response to milliseconds and add the value to the current time to calculate a static expiration timestamp
+        Date.now() + (response.expires_in * 1000 - 1000)
       ),
       ssl: response.ssl === true
     };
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-auth/src/fetch-token.ts
```diff
@@ -39,7 +39,8 @@ export function fetchToken(
       token: response.access_token,
       username: response.username,
       expires: new Date(
-        Date.now() + (response.expires_in * 60 * 1000 - 60 * 1000)
+        // convert seconds in response to milliseconds and add the value to the current time to calculate a static expiration timestamp
+        Date.now() + (response.expires_in * 1000 - 1000)
```
jgravois: squashed another bug here resulting from the fact that this endpoint asks for an `expiration` request parameter in minutes and converts the information into seconds to pass back in the response.

test case:
```js
const secureUrl = `https://route.arcgis.com/arcgis/rest/services/World/Route/NAServer/Route_World/`;
  
// generate tokens that are only valid for 1 minute
const authentication = new ApplicationSession({
  clientId,
  clientSecret,
  duration: 1
})

// generate a token
authentication.getToken(secureUrl)

// make another request 61 seconds later. fetch a fresh token instead of passing along an expired token
setTimeout(() => {
  authentication.getToken(secureUrl, { authentication })
  .then(response)
}, 61000)
```

Issue #407, opened 2018-11-27
add npm script to wipe out packages/**/dist/*


jgravois:
to ensure we don't end up with another release that includes outdated files.

jgravois:
i snuck `npm run clean` into a107ae73f1cfc2f5c7c9713f0180d1c36429eb

PR #432, opened 2019-01-11
chore: remove clean script, patch vuln, simplify post install and sync peerDependencies

jgravois:
i'm having trouble releasing all the packages in this library before my npm [2FA](https://docs.npmjs.com/about-two-factor-authentication) token expires.

this PR juggles the steps a little so that we call `npm run build` for each package during `npm run release:prepare` instead of `npm run release:publish`.

when i made these changes i noticed that most of the packages were missing the `prebuild` step that removes existing content, so i fixed that too. (a superior solution than what i came up with in #407)


jgravois:
ugh.

i screwed around with this some more this morning and realized that `lerna bootstrap` **relies** on `npm prepare` being run inside each package in order to symlink local dependencies.

i don't know of an inverse equivalent to [`prepublishOnly`](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare) so i'm going to just go ahead and give up (for now). at this point this PR:
* resolves a few security vulnerabilities
* gets rid of `npm run clean`
* simplifies the release steps and travis' post install routine a little bit
* syncs `peerDependencies`

packages/arcgis-rest-common-types/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@esri/arcgis-rest-common-types",
   "version": "1.14.4",
-  "description": "Common TypeScript types for @esri/arcgis-rest-* packages.",
+  "description": "Common TypeScript types for @esri/arcgis-rest packages.",
```
tomwayson: I think `@esri/arcgis-rest-js` is best. I'll be out by the  🚲shed if you want to discuss further.

packages/arcgis-rest-request/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@esri/arcgis-rest-request",
   "version": "1.14.4",
-  "description": "Common methods and utilities for @esri/arcgis-rest-* packages.",
+  "description": "Common methods and utilities for @esri/arcgis-rest packages.",
```
tomwayson: ditto

packages/arcgis-rest-common-types/package.json
jgravois: works for me. i needed to rebase this 💩 	anyway.

jgravois:
> Coverage decreased (-0.4%) to 99.577% when pulling 2529224 on chore/speed-up-release into 22b39a7 on master.

i don't know _why_, but once again the decrease in coverage coveralls reports is the result of bumping karma to 3.x to resolve an `npm audit` nag.

its **not** reproducible locally and doesn't even show up in the coverage report that travis generates. i'm at a loss but i'm inclined to just patch the vulnerability and move on with my life.

screenshot 2019-01-14 12 58 09

PR #434, opened 2019-01-15
docs(cdn): add sri hashes

COV-GIS:
I'm throwing this out there again @jgravois.

I added it as its own `<code>` in the template and config but could replace the CDN as well.

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -5,46 +5,16 @@ const _ = require("lodash");
 
 const IS_DEV = process.env.ENV !== "prod";
 
-const pkgVersion = process.env.npm_package_version;
-const request = require("request");
-const sriToolbox = require("sri-toolbox");
-
-const packages = [
-  "auth",
-  "common",
-  "feature-service",
-  "feature-service-admin",
-  "geocoder",
-  "groups",
-  "items",
-  "request",
-  "routing",
-  "sharing",
-  "users"
-];
-
-const sriHashes = {};
-
-generateSriHashes = () => {
-  packages.forEach(pkg => {
-    const url = `http://unpkg.com/@esri/arcgis-rest-${pkg}@${pkgVersion}/dist/umd/${pkg}.umd.min.js`;
-    request(url, (err, res, body) => {
-      if (err) {
-        sriHashes[pkg] = "SRI_HASH_FAILED";
-      } else {
-        sriHashes[pkg] = sriToolbox.generate({
-          algorithms: ["sha384"]
-        }, body);
-      }
-    });
-  });
-};
-
 module.exports = function(acetate) {
   /**
-   * Generate SRI hashes for dist minified js files.
+   * Load SRI hashes.
    */
-  generateSriHashes();
+  let srihashes;
+  fs.readFile('docs/src/srihashes.json', (err, data) => {
+    srihashes = err ? {
+      packages: {}
+    } : JSON.parse(data);
+  });
 
   /**
    * Load all .html and markdown pages in the `src` folder, assigning them a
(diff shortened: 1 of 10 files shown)


docs/acetate.config.js
```diff
@@ -5,7 +5,47 @@ const _ = require("lodash");
...
+  "sharing",
+  "users"
+];
+
+const sriHashes = {};
+
+generateSriHashes = () => {
+  packages.forEach(pkg => {
+    const url = `http://unpkg.com/@esri/arcgis-rest-${pkg}@${pkgVersion}/dist/umd/${pkg}.umd.min.js`;
+    request(url, (err, res, body) => {
```
jgravois: thanks for this contribution! i'm stoked that you're still thinking about this useful security enhancement. that said, i'm _pretty_ sure that fetching the module from unpkg defeats the purpose of an integrity check entirely.

for downstream developers the whole point is to provide a hash that is **guaranteed** to be trusted to compare with the hash generated on the fly when _they_ download the resource.

to put it another way, what if someone were to compromise https://unpkg.com _before_ we pushed an update to the docs?

COV-GIS: The best way would be to use the dist files.

From what I can tell the docs build doesn't have access to the dist files.

Am I mistaken?

jgravois: > the best way would be to use the dist files.

correct. the best (and only way) is to use whats in `/packages/**/dist/`. this needs to happen only immediately prior to tagging a release.

> From what I can tell the docs build doesn't have access to the dist files.

thats not correct. to work toward solving this problem you just need to write a new standalone helper script that generates hashes from what is on disk at will and feeds that (cached) information to acetate when it builds the site.

COV-GIS: Where in the workflow does `build` that will be the release happen?

This all reminds me of the issue I originally ran into when looking at this problem; local dist files never work for hashes due to not having the same timestamp in the banner and thus different hashes. If there is no way to capture the file exactly as it will be served, and write hashes to a file for docs to use it's not going to work as part of the build process.

jgravois: > Where in the workflow does build that will be the release happen?

funny that you mention it. in #432 and discovered that the [`prepare`](https://github.com/Esri/arcgis-rest-js/blob/88bf26b6ba2f3c97b231ac3de764e0739c058630/packages/arcgis-rest-request/package.json#L18) script for each package is what triggers the "official" build.

docs/src/srihashes.json
```diff
@@ -0,0 +1,16 @@
+{
+  "version": "1.14.4",
+  "packages": {
```
COV-GIS: edit to `"packages": {}` **before** merging

COV-GIS:
This *should* work @jgravois.

1. hashes generated during release
1. `srihashes.json` committed at release for use in interrelease docs updates
1. fail safe at every step; `CDN with SRI:` simply doesn't get added if there's any errors

As noted...the `package: hash` key/values need to be removed before merging. And should have good hashes directly from `dist` files at next release.

jgravois:
nice work.

the hashes will indeed be generated during the release, but your script needs to run _after_ lerna spawns the `npm publish` command for each package (which calls `npm run prepare`, which calls `npm run build`) if we want them to be valid. 

https://github.com/Esri/arcgis-rest-js/blob/07dd8892593c697932b9e5be4036d4378ae5abc6/support/publish.sh#L33

jgravois:
> How will `srihashes.json` get committed with the release?

whoops. i forgot about that part. 😆

i've pushed up another commit to reorder the steps in which hashes are generated _and_ committed. i'll give it a go when i tag a release this afternoon.

jgravois:
@COV-GIS **thank you** for this contribution!

PR #436, opened 2019-01-16
Add a "headers" option to IRequestOptions

rgwozdz:
This PR does the following

1) adds a `headers` option to IRequestOptions. Optional headers are passed on to the target request. 

2) in Node environments, the presence of the `headers` options prevents the default `referer` header
`"@esri/arcgis-rest"` from being set on the request. If a `referer` is desired, it can be set with the header option.

I have found that (2) is critical in a particular case; ArcGIS search service sometimes rejects a request with a valid token when the `referer` is set to `"@esri/arcgis-rest"`.  (Note that in such cases, ArcGIS search returns a `498 Invalid token` response; but stripping the referer header from the request results in a success response, which indicates the problem is not with the token.)

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -6,7 +6,12 @@ import { encodeQueryString } from "./utils/encode-query-string";
 import { requiresFormData } from "./utils/process-params";
 import { ArcGISRequestError } from "./utils/ArcGISRequestError";
 import { IRetryAuthError } from "./utils/retryAuthError";
-import { HTTPMethods, IParams, ITokenRequestOptions } from "./utils/params";
+import {
+  HTTPMethods,
+  IParams,
+  ITokenRequestOptions,
+  IHeaders
+} from "./utils/params";
 
 /**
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -178,10 +188,15 @@ export function request(
...
 
-      fetchOptions.headers = {};
+      // Mixin headers from request options
+      fetchOptions.headers = { ...requestOptions.headers };
 
       /* istanbul ignore next - karma reports coverage on browser tests only */
-      if (typeof window === "undefined") {
+      if (
+        typeof window === "undefined" &&
+        requestOptions.headers === undefined
```
dmfenton: should we set the referrer when headers are passed-in but referer is still undefined?

rgwozdz: Ah, yeah, that's prob a good call

rgwozdz:
Closing this, now covered in https://github.com/Esri/arcgis-rest-js/pull/437

jgravois:
thanks for this contribution @rgwozdz 🌮:tada:!

PR #440, opened 2019-01-17
chore/codecov

### LINKED ISSUE #438: solve the flaky 99.999% coverage reports from coveralls
i _think_ i learned enough about sending coverage reports in [`react-arcgis`](https://github.com/Esri/react-arcgis/blob/master/.travis.yml) to simplify our own CI setup here and get consistent 💯% coverage reports from `karma@3`

instead of using two different reporters, we just need to figure out a way to send what gets generated in `/coverage` on disk from the report that appears in the terminal.

https://github.com/Esri/arcgis-rest-js/blob/435a6401ac946748fa0987554891b66c8e95dc73/karma.conf.js#L46-L50

jgravois:
take 2

resolves #438


jgravois:
reports are being uploaded, but i don't think codecov can post a comment on a PR until there is a base commit in `master` to compare to.

screenshot 2019-01-16 23 29 43

https://codecov.io/github/Esri/arcgis-rest-js/commit/77f55414139da651f3083aab49eda70068a5c670

PR #450, opened 2019-01-23
Added items/removeFolder function

MikeTschudi:


jgravois:
i'm getting ready to tag a release, so i pushed up 8dbbab4 with a few small tweaks.

* add a code snippet for the API reference
* ensure that we encode special characters in usernames
* define the actual structure of the response

MikeTschudi:
Thanks, @jgravois!

PR #453, opened 2019-01-25
Modified feature service creation to handle HTTP 400 errors

MikeTschudi:
If creating a feature service or adding layers &| tables to it fails with an HTTP 400 error, the createFeatureService function doesn't respond with a `resolve` or `reject`, preventing the calling program from resolving the Promise returned by the function.

### DIFF
--- a/packages/arcgis-rest-feature-service-admin/test/create.test.ts
+++ b/packages/arcgis-rest-feature-service-admin/test/create.test.ts
@@ -7,7 +7,9 @@ import { createFeatureService } from "../src/create";
 
 import {
   FeatureServiceSuccessResponse,
-  FeatureServiceFailResponse
+  FeatureServiceFailResponse,
+  FeatureServiceFailResponse400,
+  AddToFeatureServiceBadRequest
 } from "./mocks/service";
 import {
   MoveToFolderSuccessResponse,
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-feature-service-admin/test/create.test.ts
```diff
@@ -341,7 +433,56 @@ describe("create feature service", () => {
...
+      createFeatureService({
+        item: serviceDescription,
+        folderId,
+        ...MOCK_USER_REQOPTS
+      })
+        .then(
+          () => {
+            fail(); // call is supposed to fail
+          },
+          response => {
```
jgravois: when you `reject`, i'd expect it to trigger the calling `catch()`.

am i missing something?

MikeTschudi: What I was finding was that the existing approach prevented the calling routine from handling either a resolve or a reject--even a try...catch block in the calling routine wasn't enough to catch the error.  There may be a better approach than this PR, but it does trigger the calling routine's resolve or reject when a 400-level error is returned from the server.

MikeTschudi:
Handles both authentication and move 400s--thanks!

PR #457, opened 2019-02-14
feat: (@esri/arcgis-rest-request) support for raw responses

tannerjt:


tannerjt:
Reference discussion in #457

tannerjt:
Thanks @tomwayson.  Can you take a look now.  I understand what @jgravois is saying now by returning the raw response so the user can use any of the response methods.  It should achieve the desired result now.  I've also updated the assertion test and demo.

jgravois:
a most excellent contribution. thank you @tannerjt! 🙏 	

be excellent to one another

PR #465, opened 2019-02-16
WIP: add a IRequestOptionsBase for fns that don't support rawResponse

tomwayson:
TODO: how to keep rawResponse out of some interfaces that don't extend directly from IRequestOptions

See: https://github.com/Esri/arcgis-rest-js/pull/463#discussion_r257432175
and: https://github.com/Esri/arcgis-rest-js/pull/463#discussion_r257432692


tomwayson:
The other problem w/ this is that it's now _one more click_ that users have to make in the docs to get what they need.

PR #467, opened 2019-02-25
Switch polyfill.io to v3

COV-GIS:
polyfill.io v2 is depreciated. And the v2 docs and bundle builder aren't even available anymore. The v2 cdn still works but if a dev wants to add other features or just check out pollyfill.io they'll most likely switch anyway.

The `es5` feature doesn't exist in v3. I believe this what is now the `default` feature.


jgravois:
> The `es5` feature doesn't exist in v3. I believe this what is now the `default` feature.

are you sure? the [`v3` API Reference](https://polyfill.io/v3/api) still includes an example for it.

> **Example request:**
> Requesting the polyfill sets: ES5, ES6, and ES7. Wrapping each polyfill in a feature detect
> ```
> https://polyfill.io/v3/polyfill.js?features=es5,es6,es7&flags=gated
> ```

COV-GIS:
> are you sure?

Not now...

The old bundler listed es* features.

I'll look into it.

COV-GIS:
It's not well documented. `es5` will include any additional polyfills required by `fetch` and `Promise` in this case. `default` will work but it's much larger.

PR #468, opened 2019-02-26
re-export IGeocodeParams (until v2.0.0)

jgravois:
@skitterm aptly pointed out that i removed `IGeocodeParams` in `v1.17.0`.

https://github.com/Esri/arcgis-rest-js/commit/fb2ba9ac2ccaf6ed17e02ecb170009b68f638b00#diff-094efe5f622403522c0543b08434e397

just because we weren't actually utilizing the interface in any of the geocoding methods doesn't mean other developers weren't using it.

whoops!

this PR just adds a note to the CHANGELOG as breadcrumb. while i was in there i bumped the copyright year and added keywords to our package.json files to fix this 😢 	

screenshot 2019-02-25 16 21 18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,9 +31,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### @esri/arcgis-rest-geocoder
 
-* Removed
-   * the (unused) `IGeocodeParams` interface has been removed. [`fb2ba9a`](https://github.com/Esri/arcgis-rest-js/commit/fb2ba9ac2ccaf6ed17e02ecb170009b68f638b00) [#459](https://github.com/Esri/arcgis-rest-js/issues/459)
-
 * Fixes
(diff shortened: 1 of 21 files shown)


jgravois:
if @tomwayson (or any other SemVer police) tell me i need to bump to `v2.0.0`, i am **not** afraid.

patrickarlt:
@jgravois why not just add it back in until 2.0.0 THEN remove it?

tomwayson:
> i am **not** afraid



I don't _do_ semver for types, so LGTM. Types aren't real anyway.

patrickarlt:
> Types aren't real anyway.
> @tomwayson 

You monster 👹 👿 !

jgravois:
> why not just add it back in until 2.0.0 THEN remove it?

that never even occurred to me. great idea.

patrickarlt:
@jgravois should we really remove this anyway?? leaving it in doesn't bump out bundle sizes and leaving it in also allow TypeScript users to get autocomplete and stricter type checking in their IDEs.

jgravois:
> should we really remove this anyway??

it should _absolutely_ be removed. its an abandoned antipattern.

as you said in https://github.com/esri/arcgis-rest-js/issues/137#issuecomment-393362439, we want folks to extend `IRequestOptions`, _not_ `IParams`.

patrickarlt:
Derp good point.

PR #469, opened 2019-02-26
Doc Optimization

patrickarlt:
This PR proposes some of the changes I suggested in https://github.com/Esri/arcgis-rest-js/issues/137#issuecomment-393362439.

1. On pages for functions like `createItem`
2. Where there is a function param called `requestOptions`
3. Print a table under "Options" with all the properties of the type of `requestOptions`.

Then to add some more value

1. where the return type is a known declaration (i.e. something we have defined) in the library somewhere OR
2. A promise that resolves to 1
3. Print a table under "Returns" with all the properties of the type of the object.

This makes pages like `createItem` MUCH more useful since you can now reference all the options AND the return values on 1 page mostly resolving @dbouwman's issues in https://github.com/Esri/arcgis-rest-js/issues/137#issuecomment-392861393.

screen shot 2019-02-25 at 22 39 27-fullpage

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -10,10 +10,12 @@ module.exports = function(acetate) {
    * Load SRI hashes.
    */
   let srihashes;
-  fs.readFile('docs/src/srihashes.json', (err, data) => {
-    srihashes = err ? {
-      packages: {}
-    } : JSON.parse(data);
+  fs.readFile("docs/src/srihashes.json", (err, data) => {
+    srihashes = err
+      ? {
+          packages: {}
+        }
+      : JSON.parse(data);
   });
 
   /**
@@ -186,7 +188,6 @@ module.exports = function(acetate) {
   });
 
   acetate.filter("findPackage", (typedoc, name) => {
-    console.log(typedoc.packages, name);
     return typedoc.packages.find(p => p.pkg.name === name).pkg;
   });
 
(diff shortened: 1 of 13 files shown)


dbouwman:
The output looks great @patrickarlt! The build is failing in CI though...

jgravois:
the build is failing because the `postinstall` script has been removed from the package.json.

if we don't want to bootstrap automatically on `npm install` anymore another option would be to add the command to the travis.yml as a `before_script`

patrickarlt:
Fixed the build. I was having some issues related to `npm install` and `lerna bootstrap`

patrickarlt:
@jgravois whoops should be fixed now.

COV-GIS:
This is great @patrickarlt! The docs will be much more user friendly.

If you have time take a look at this... https://github.com/Esri/arcgis-rest-js/blob/master/docs/src/js/nav-toggle.js#L8-L12 matches both `@esri/arcgis-rest-feature-service` and `@esri/arcgis-rest-feature-service-admin`. So clicking `@esri/arcgis-rest-feature-service` opens both sub menus. I couldn't see a simple and elegant solution. Mildly annoying more than anything.

patrickarlt:
@noahmulfinger I added support for when we need to handle things by named reference not just by ID. This resolves almost everything in your list.

However for `arcgis-rest-js/api/users/getUser/` it accepts a union which is a BUNCH more work that I would rather delay. I might be able to get to it by dev summit, I might now.

As for the header this now says  "Available `requestOptions`"

jgravois:
thank you @patrickarlt for the **massive** DX improvement and @noahmulfinger for the shrewd feedback!

PR #474, opened 2019-02-28
fix: ensure an empty array request parameter is stringified and passed through

jgravois:
i'm working on a [web component](https://github.com/esridc/hub-components) that updates user tags and noticed this afternoon that we currently don't pass a value at all when an empty array is provided as a request parameter.

```ts
request(updateUserUrl, {
  params: { tags: [] }, 
  authentication
});
```
no tags

ArcGIS Online's (understandable) reaction to this call is to make _no_ edits to the user's current tags, when in actuality I want to remove them all.

AFFECTS PACKAGES:
@esri/arcgis-rest-request

### DIFF
--- a/packages/arcgis-rest-request/src/utils/process-params.ts
+++ b/packages/arcgis-rest-request/src/utils/process-params.ts
@@ -65,12 +65,12 @@ export function processParams(params: any): any {
     // null, undefined, function are excluded. If you want to send an empty key you need to send an empty string "".
     switch (type) {
       case "Array":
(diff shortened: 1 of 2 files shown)


jgravois:
huh, it turns out this isn't the right thing to do at all.

i should be using the optional [`clearEmptyFields`](https://developers.arcgis.com/rest/users-groups-and-items/update-user.htm) parameter instead.

jgravois:
wow. 

`clearEmptyFields` only works on string fields. i have absolutely no idea how to clear a user's current tags.

PR #475, opened 2019-02-28
fix: deprecate serializeGroup() and simplify SerializeItem()

jgravois:
it dawned on me when i submitted #474 that the custom logic we have for serializing groups and items isn't even needed.

the only thing going on in there thats worth keeping is stuffing `item.data` into a `text` parameter.

AFFECTS PACKAGES:
@esri/arcgis-rest-groups
@esri/arcgis-rest-items

### DIFF
--- a/packages/arcgis-rest-groups/src/create.ts
+++ b/packages/arcgis-rest-groups/src/create.ts
@@ -7,9 +7,9 @@ import {
   getPortalUrl
 } from "@esri/arcgis-rest-request";
 
-import { IGroupAdd } from "@esri/arcgis-rest-common-types";
+import { IGroupAdd, IGroup } from "@esri/arcgis-rest-common-types";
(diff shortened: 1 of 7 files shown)


jgravois:
@jnordling if you were seeing `createGroup()` or `updateGroup()` strip custom parameters in your requests, this PR fixes that issue and you should be good to go with [`v1.17.1`](https://github.com/Esri/arcgis-rest-js/releases/tag/v1.17.1).

if you were seeing unexpected behavior somewhere else, please let me know where.

you shouldn't _ever_ need to sidestep [UserSession](https://esri.github.io/arcgis-rest-js/api/auth/UserSession/) and pass `parameters: { token: "foo" }` in your calls. if its ever more convenient, either we need to write more doc or squash a few more 🐛s!

PR #481, opened 2019-03-06
Dev summit 2019 demos

patrickarlt:


patrickarlt:
@jgravois I've updated the read me for the sapper demo to explain the config which was git ignored. I also added a bunch of doc explaining auth. I also updated the code a bit and added some more comments.

jgravois:
awesome @patrickarlt! i just pushed up a few more typo fixes of my own.

it took me a couple minutes to figure out, but i had to remove the trailing slash from the redirect URI you were suggesting in the README in order to resolve the route that Sapper is creating on the fly from `post-sign-in.js`.

```text
REDIRECT_URI=http://localhost:3000/auth/post-sign-in
```

PR #486, opened 2019-03-13
feat: Support 3d point/location

DKaminari:
Add support for z for point and location

PR #487, opened 2019-03-13
fix(routing): Fix switched x and y

DKaminari:
Fix switched x and y for point in solveRoute.
When providing point object with x,y, solveRoute switch their order to y,x.
The test case also had the switch.
To clarify the issue, x == long and y == latitude


jgravois:
it appears that #488 is cumulative of both #486 and #487 and that this PR can be closed.

next time you'd like to open more than one pull request at the same time, just make sure `master` is the base of each of your own branches.

PR #488, opened 2019-03-13
Support routing with elevation

DKaminari:
Add support for elevation z in routing by allow a third number or z value

### DIFF
--- a/packages/arcgis-rest-common/test/location.test.ts
+++ b/packages/arcgis-rest-common/test/location.test.ts
@@ -66,24 +66,24 @@ const stopsObjectsLatitudeLongitude3: ILocation[] = [
 
 const stopsObjectsPoint: IPoint[] = [
   {
-    x: 34.056383,
-    y: -117.195677
+    x: -117.195677,
+    y: 34.056383
   },
   {
-    x: 33.812092,
-    y: -117.918976
+    x: -117.918976,
+    y: 33.812092
   }
 ];
 
 const stopsObjectsPoint3: IPoint[] = [
   {
-    x: 34.056383,
-    y: -117.195677,
+    x: -117.195677,
+    y: 34.056383,
     z: 10.11
   },
   {
-    x: 33.812092,
-    y: -117.918976,
+    x: -117.918976,
+    y: 33.812092,
     z: 8.43
   }
 ];
--- a/packages/arcgis-rest-common/test/location.test.ts
+++ b/packages/arcgis-rest-common/test/location.test.ts
@@ -11,6 +11,11 @@ const stops: Array<[number, number]> = [
   [-117.918976, 33.812092]
 ];
 
+const stops3: Array<[number, number, number]> = [
+  [-117.195677, 34.056383, 10.11],
+  [-117.918976, 33.812092, 8.43]
+];
+
 const stopsObjectsLatLong: ILocation[] = [
   {
     lat: 34.056383,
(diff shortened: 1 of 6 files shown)


packages/arcgis-rest-common/test/location.test.ts
```diff
@@ -44,30 +75,69 @@ const stopsObjectsPoint: IPoint[] = [
...
+  {
+    x: 34.056383,
+    y: -117.195677,
+    z: 10.11
+  },
+  {
+    x: 33.812092,
+    y: -117.918976,
+    z: 8.43
+  }
```
jgravois: your x's and y's are transposed here.

packages/arcgis-rest-routing/src/solveRoute.ts
```diff
@@ -80,12 +82,19 @@ export function solveRoute(
       return coords.join();
     } else if (isLocation(coords)) {
       if (coords.lat) {
-        return coords.long + "," + coords.lat;
+        return (
+          coords.long + "," + coords.lat + (coords.z ? "," + coords.z : "")
```
jgravois: can you explain the benefit of passing along elevation values to `/solve`?

i did a (very) little bit of testing and it doesn't appear to have any effect on the geometries that are returned.

https://route.arcgis.com/arcgis/rest/services/World/Route/NAServer/Route_World/solve?stops=-117.195677,34.056383,10.11;-117.918976,33.812092,8.43&f=json&&outSR=102100&directionsOutputType=esriDOTInstructionsOnly&token=[yours]

DKaminari: The benefit is to allow for 3d routing. I came across this as I was working on a project which required indoor routing over multiple floors with each floor having different elevation z. The result will be clearer if you set returnZ to true and get the result back using an elevation aware network.

packages/arcgis-rest-common/test/location.test.ts
DKaminari: Correct, I did not notice this in the location test case as I copied stopsObjectsPoint into stopsObjectsPoint3 to add the elevation z. It does not have any effect on the validity for the result so we can ignore it.

jgravois: > we can ignore it.

i'd prefer that we fix it.

packages/arcgis-rest-routing/src/solveRoute.ts
jgravois: gotcha. thanks for clarifying!

packages/arcgis-rest-common/test/location.test.ts
DKaminari: So, how can I fix it? Do I submit another pull request?

jgravois: > how can I fix it?

just push one more commit on your own `support-3d-route` branch and it will show up in _this_ pull request.

packages/arcgis-rest-routing/src/solveRoute.ts
```diff
@@ -80,12 +82,19 @@ export function solveRoute(
...
+        return (
+          coords.longitude +
+          "," +
+          coords.latitude +
+          (coords.z ? "," + coords.z : "")
+        );
       }
     } else {
-      return coords.y + "," + coords.x;
+      return coords.x + "," + coords.y + (coords.z ? "," + coords.z : "");
```
jgravois: this is a great catch.

PR #494, opened 2019-03-26
feat: new joinGroup and leaveGroup methods

jgravois:
made some tweaks to simplify protect/unprotect and make it more evident to TypeScript users that passing authentication is mandatory.

AFFECTS PACKAGES:
@esri/arcgis-rest-groups

### DIFF
--- a/packages/arcgis-rest-groups/src/join.ts
+++ b/packages/arcgis-rest-groups/src/join.ts
@@ -7,25 +7,25 @@ import { IGroupIdRequestOptions } from "./helpers";
 
 /**
  * ```js
- * import { protectGroup } from '@esri/arcgis-rest-groups';
+ * import { joinGroup } from '@esri/arcgis-rest-groups';
  * //
- * protectGroup({
+ * joinGroup({
  *   id: groupId,
  *   authentication
  * })
  *   .then(response)
  * ```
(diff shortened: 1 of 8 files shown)


packages/arcgis-rest-groups/src/join.ts
```diff
@@ -0,0 +1,56 @@
...
+ * ```js
+ * import { joinGroup } from '@esri/arcgis-rest-groups';
+ * //
+ * joinGroup({
+ *   id: groupId,
+ *   authentication
+ * })
+ *   .then(response)
+ * ```
+ * Join a Group. See the [REST Documentation](https://developers.arcgis.com/rest/users-groups-and-items/join-group.htm) for more information.
```
tomwayson: Maybe "Apply to join a group as the authenticated user" instead of "Join a Group"

packages/arcgis-rest-groups/src/join.ts
```diff
@@ -0,0 +1,56 @@
...
+  const url = `${getPortalUrl(requestOptions)}/community/groups/${
+    requestOptions.id
+  }/join`;
+
+  return request(url, requestOptions);
+}
+
+/**
+ * ```js
+ * import { joinGroup } from '@esri/arcgis-rest-groups';
```
tomwayson: this should be `leaveGroup()`

packages/arcgis-rest-groups/src/join.ts
jgravois: agreed. fixed (and rebased).

Issue #499, opened 2019-03-29
Provide option to force request to use FormData content type


ssylvia:
There are some API requests that require `FormData` as the content type. We need a fallback option to force the request to FormData when only JSON params are passed so `x-www-form-urlencoded` is used as the fallback.

jgravois:
in [`v1.19.1`](https://github.com/Esri/arcgis-rest-js/releases/v1.19.1) and above `updateItemResource()` will use FormData automatically.

if you can do me a big favor and loop back _after_ June. i'd love to get confirmation that the underlying bug has been fixed.

patrickarlt:
@jgravois even if the underlying bug IS fixed in June do we will leave this in place for people who are using older versions of Portal?

jgravois:
great point.

The Portal API doc doesn't advertise when individual parameters landed in Enterprise. Its possible that by the time you can set `access` on an existing resource, FormData will no longer be required.

That's just a guess though. until we know for sure we should probably leave our check in place.

PR #500, opened 2019-03-29
fix: force updateItemResource to use FormData

jgravois:
i hate to suggest we add this, but i have confirmed @ssylvia's report in #499 

[`/updateResources`](https://developers.arcgis.com/rest/users-groups-and-items/update-resources.htm) is indeed misconfigured and currently only honors requests constructed with FormData (even when the parameters passed don't require it).

to make things worse, the endpoint returns `{ success: true }` regardless.

```js
arcgisRest.updateItemResource({
  id: "fe8",
  name: "logo.png",
  private: true,
  authentication
})
  .then(response)
```

until the issue is resolved (hopefully in June), my best idea is ~~to inject an undocumented `IParam` that triggers the use of FormData regardless of the input parameters and then strip it out before the request is sent.~~ to insert a little more logic inside `request` that ensures FormData will be used for calls to /updateResources` regardless of the input params.

```ts
requiresFormData(params) ||
    (url && new RegExp("/items/.+/updateResources").test(url));
```
with this change `updateItemResource()` will _just work_.

AFFECTS PACKAGES:
@esri/arcgis-rest-request

ISSUES CLOSED: #499

### DIFF
--- a/packages/arcgis-rest-items/src/update.ts
+++ b/packages/arcgis-rest-items/src/update.ts
@@ -1,7 +1,11 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { request, getPortalUrl } from "@esri/arcgis-rest-request";
+import {
+  request,
+  getPortalUrl,
+  IRequestOptions
+} from "@esri/arcgis-rest-request";
 
 import { IItemUpdate } from "@esri/arcgis-rest-common-types";
 
@@ -87,9 +91,9 @@ export function updateItemResource(
   requestOptions: IItemResourceRequestOptions
 ): Promise<any> {
   const owner = determineOwner(requestOptions);
-  const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
-    requestOptions.id
-  }/updateResources`;
+  const url = `${getPortalUrl(
+    requestOptions as IRequestOptions
+  )}/content/users/${owner}/items/${requestOptions.id}/updateResources`;
 
   // mix in user supplied params
   requestOptions.params = {
@@ -99,13 +103,12 @@ export function updateItemResource(
     ...requestOptions.params
   };
 
-  // only override whatever access was specified previously if 'private' was passed explicitly
+  // only override the access specified previously if 'private' is passed explicitly
   if (typeof requestOptions.private !== "undefined") {
     requestOptions.params.access = requestOptions.private
       ? "private"
       : "inherit";
   }
-
   return request(url, requestOptions);
 }
 
--- a/packages/arcgis-rest-request/src/utils/encode-form-data.ts
+++ b/packages/arcgis-rest-request/src/utils/encode-form-data.ts
@@ -9,8 +9,12 @@ import { encodeQueryString } from "./encode-query-string";
  * @param params An object to be encoded.
  * @returns The complete [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
  */
-export function encodeFormData(params: any): FormData | string {
-  const useFormData = requiresFormData(params);
+export function encodeFormData(
+  params: any,
+  forceFormData?: boolean
+): FormData | string {
+  // see https://github.com/Esri/arcgis-rest-js/issues/499 for more info.
+  const useFormData = requiresFormData(params) || forceFormData;
   const newParams = processParams(params);
   if (useFormData) {
     const formData = new FormData();
--- a/packages/arcgis-rest-items/test/update.test.ts
+++ b/packages/arcgis-rest-items/test/update.test.ts
@@ -214,19 +214,21 @@ describe("search", () => {
         content: "jumbotron",
         ...MOCK_USER_REQOPTS
       })
-        .then(response => {
+        .then(() => {
           const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
           expect(url).toEqual(
             "https://myorg.maps.arcgis.com/sharing/rest/content/users/dbouwman/items/3ef/updateResources"
           );
           expect(options.method).toBe("POST");
-          expect(options.body).toContain("f=json");
-          expect(options.body).toContain(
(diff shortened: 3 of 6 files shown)


packages/arcgis-rest-items/src/update.ts
```diff
@@ -87,25 +91,26 @@ export function updateItemResource(
...
-  const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
-    requestOptions.id
-  }/updateResources`;
+  const url = `${getPortalUrl(
+    requestOptions as IRequestOptions
+  )}/content/users/${owner}/items/${requestOptions.id}/updateResources`;
 
   // mix in user supplied params
   requestOptions.params = {
     file: requestOptions.resource,
```
tomwayson: Is this an instance of a [File](https://developer.mozilla.org/en-US/docs/Web/API/File)? If so, shouldn't that cause `requiresFormData()` to return `true` even w/o passing `useFormData: true`?

packages/arcgis-rest-items/src/update.ts
```diff
@@ -87,25 +91,26 @@ export function updateItemResource(
...
     fileName: requestOptions.name,
     text: requestOptions.content,
```
tomwayson: If `file` isn't an instance of `File` or is optional, then maybe we can construct a new [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) out of `content`, which should cause `requiresFormData()` to return `true`, right?

ssylvia:
@tomwayson Both `file` and `params.text` are now optional so unfortunately I don’t think your suggestions will work.

We asked for a change that was added in the march release that would allow us to update the `access` param (`private` option) without having to upload and replace the current resource. The only options that are passed will be: itemId, filename, and the private option.

The portal team confirmed that FormData is currently required but will update the method to be more flexible in the 7.2 (June) release.

We already tried passing a fake, empty file, as an extra param `params.fakeFile` to force FormData but apparently the REST API still uses this to update (and replace) the current resource.

jgravois:
> We already tried passing a fake, empty file, as `params.fakeFile` to force FormData but apparently the REST API still uses this to update (and replace) the current resource.

i've confirmed this and it appears to tie our hands even further.

@patrickarlt do you have time to jump in and help us break the tie or perhaps offer an alternative suggestion?

in general, my instinct was to keep this undocumented to
1. try and keep the public API as simple as possible
2. _not_ encourage developers into thinking they typically need to worry about request internals.

i'd have liked to suggest that @ssylvia just install from this branch until the bug is patched in June, but it doesn't appear you can point npm at a package in a Lerna monorepo via a git url.

https://github.com/lerna/lerna/issues/1003#issuecomment-328597940

tomwayson:
no tie to break, I'm fine your solution @jgravois

jgravois:
@patrickarlt my first instinct was intentionally _not_ to document the change because it doesn't correspond with anything documented in the Portal API. we're simply providing internal scaffolding to work around a (temporary) bug.

i'd even be willing to go so far as making the code comments more explicit to warn that the functionality is subject to change without notice.

on the flip side, if either of you think it'd be preferable to surface this and make the change 'official', couldn't we append it directly to `IParams` instead of `IRequestOptions`?

https://github.com/Esri/arcgis-rest-js/blob/9d79fe031c8b187966995a15407b741038e286c9/packages/arcgis-rest-request/src/utils/params.ts#L26-L29

becomes:
```ts
export interface IParams { 
  f?: ResponseFormats; 
  useFormData?: boolean;
  [key: string]: any; 
} 
```

patrickarlt:
> @patrickarlt my first instinct was intentionally not to document the change because it doesn't correspond with anything documented in the Portal API. we're simply providing internal scaffolding to work around a (temporary) bug.

@jgravois I have no problem with doing this but what about the cases where someone wants to build something around `request` and needs to use this option to properly handle resources?

> couldn't we append it directly to IParams instead of IRequestOptions?

I actually feel pretty strongly about this. `IParams` is ONLY for things that are documented params on the REST API. `IRequestOptions` is specially for controlling things about the request including weird overrides such as this. The behavior for forcing the use of FormData isn't about a param on the REST API it is about modifying the behavior of the request to get a specific behavior which sounds much more like something we would put on `IRequestOptions` like `rawResponse`.

jgravois:
> what about the cases where someone wants to build something around request and needs to use this option to properly handle resources?

besides a misconfigured endpoint like the one @ssylvia discovered, i can't think of a reason a user would need to pass request parameters as FormData when the actual values don't require it.

Either way I agree with you that what i'm doing currently is basically just an abuse of `IParams`.

in light of the conversation, i came up with another idea.

i just pushed up another commit that substitutes a (simpler) RegExp test inside `request` to see if we're making a call to the endpoint that's currently misbehaving.

i feel better about this approach than my original suggestion, but i'm (still) open to feedback.

packages/arcgis-rest-request/src/utils/encode-form-data.ts
```diff
@@ -9,8 +9,14 @@ import { encodeQueryString } from "./encode-query-string";
  * @param params An object to be encoded.
  * @returns The complete [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
  */
-export function encodeFormData(params: any): FormData | string {
-  const useFormData = requiresFormData(params);
+export function encodeFormData(params: any, url?: string): FormData | string {
```
tomwayson: maybe instead of taking a `url` as the second arg, this should take `forceFormData?: boolean` and then it would then: `const useFormData = forceFormData  || requiresFormData(params);`.

That way we could have `request()` be the only place this temporary logic lives, like:

```ts
const forceFormData = new RegExp("/items/.+/updateResources").test(url);
// ...
fetchOptions.body = encodeFormData(params, forceFormData);
// ...
if (
        !requiresFormData(params) &&
        !forceFormData
      ) {}
```

jgravois: great idea. incorporated (and rebased)

PR #503, opened 2019-04-06
make it possible for retry() to strip an invalid session

jgravois:
tl:dr: _super_ edge case fix.

---
in Hub land we started encountering errors when passing a valid (www.arcgis.com) authentication session to _public_ services hosted by unfederated instances of ArcGIS Server.

because of this currently we're sending a whole lot of anonymous requests _after_ users authenticate and using retry() to tack on a session we had the whole time whenever we encounter an `ArcGISAuthError` (see https://github.com/Esri/ember-arcgis-server-services/pull/46 for more info).

it'd be a whole lot nicer to invert the flow and pass through the authentication we have consistently and `catch` the ones that are public, but unfederated.

this PR would allow us to do that, by ensuring that the original `url` and `options` are passed through.

```js
const url = `https://sampleserver6.arcgisonline.com/arcgis/rest/services/WorldTimeZones/MapServer`;
request(url, {
    authentication,
    params: {
      foo: 'bar'
    }
  })
    .catch(err => {
      if (err.name === 'ArcGISAuthError') {
        // instead of supplying auth, strip it 
        err.retry(() => Promise.resolve(undefined))
          .then(response)
      }
    })
```

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -151,13 +151,18 @@ export function request(
 
   const fetchOptions: RequestInit = {
     method: httpMethod,
-    // ensures behavior mimics XMLHttpRequest. needed to support sending IWA cookies
+    /* ensures behavior mimics XMLHttpRequest.
+    needed to support sending IWA cookies */
     credentials: "same-origin"
   };
 
   return (authentication
-    ? authentication.getToken(url, {
-        fetch: options.fetch
+    ? authentication.getToken(url, { fetch: options.fetch }).catch(err => {
+        /* if necessary, append original request url and
+         requestOptions to the error thrown by getToken() */
+        err.url = url;
+        err.options = options;
+        throw err;
       })
     : Promise.resolve("")
   )
--- a/demos/vue/src/components/App.vue
+++ b/demos/vue/src/components/App.vue
@@ -222,9 +222,7 @@ export default {
             // session.
             this.$store.dispatch("updateSession", session);
           })
-          .catch(error => {
-            console.error(error);
-          });
+          .catch(error);
       }
     },
     // The signup with an inline redirect workflow. In this case the user is just redirected to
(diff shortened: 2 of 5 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -156,9 +160,13 @@ export function request(
...
   return (authentication
-    ? authentication.getToken(url, {
-        fetch: options.fetch
-      })
+    ? authentication.getToken(
+        url,
+        {
+          fetch: options.fetch
+        },
+        options
```
tomwayson: First, doesn't this need to be `originalRequestOptions: options`?

Also, the third arg is now being passed regardless of whether or not `authentication` is a `UserSession` or any other type of `IAuthenticationManager`, but we haven't added `originalRequestOptions` to `IAuthenticationManager`'s `getToken()` signature. I _think_ that's technically OK, but feels a little wonky.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -793,7 +802,10 @@ export class UserSession implements IAuthenticationManager {
           if (!new RegExp(response.owningSystemUrl, "i").test(this.portal)) {
             throw new ArcGISAuthError(
               `${url} is not federated with ${this.portal}.`,
-              "NOT_FEDERATED"
+              "NOT_FEDERATED",
+              response,
+              url,
+              originalRequestOptions
```
tomwayson: can we just pass `requestOptions` here instead of needing a new argument?

jgravois:
@tomwayson was right. we can just `catch` and release the AuthError from request().

🎣🐠🐟

patrickarlt:
@jgravois I would like to review this too. I should have it done before the end of the day.

jgravois:
sounds good @patrickarlt. take your time.

jgravois:
rebased (and merged). thank you both!

Issue #504, opened 2019-04-08
getItemData() throws JSON parsing error if data is empty


tomwayson:
The error is:

```
Fetch item data error. SyntaxError: Unexpected end of JSON input
    at eval (VM12042 request.js:134)
```

If the data is empty, this function could return `null` instead of forcing consuming apps to test for the above error.

jgravois:
@qlqllu this is resolved in [`v1.19.2`](https://github.com/Esri/arcgis-rest-js/releases/v1.19.2)

qlqllu:
Thanks for your quick fix, it works now.

pfcstyle:
If in Safari, the error is
`SyntaxError: The string did not match the expected pattern.`

So it still has problem in safari. @jgravois

jgravois:
I won't be able to look into this anytime soon. i'd recommend opening a new issue with a code snippet that runs against an existing public item.

a PR would work too if you know how to fix the problem.

pfcstyle:
Thanks @jgravois !  I submitted a PR: https://github.com/Esri/arcgis-rest-js/pull/677

Issue #495, opened 2019-03-27
Params with a "null" value are not included in request.


ssylvia:
I'm using the `updateItem` method to set the `itemDetails` and `itemData` on a Story Map. If a user clicks "Unpublish", we need to remove this `itemData`. This should be able to be accomplished with the following:

```ts
updateItem({
  authentication: this.userSession,
  item: this.itemDetails,
  params: {
    // Clear item data
    text: null,
  },
})
```

However, the `text: null` is never included as a param in the request.

Currently, I have to make a second request if I want to update the itemDetails and remove the itemData at the same time:

```ts
addItemJsonData({
  authentication: this.userSession,
  id: this.itemId,
  data: null,
});
```

jgravois:
right now to send a parameter with an empty key you need to pass `""`.

https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/src/utils/process-params.ts#L62-L65

would that work for you in this scenario?

jgravois:
i did a little testing and i was _not_ able to reproduce the behavior you described.

for me, this syntax induces the error below.
```js
addItemJsonData({
  authentication,
  id,
  data: null,
});
```

> {"error":{"code":400,"messageCode":"CONT_0057","message":"Invalid JSON in 'text' parameter.","details":[]}}

i just opened #496 with a tweak that resolves the error, but does _not_ get to the root of your problem.

when those changes are incorporated:

1. `data: null` causes the parameter to be dropped entirely (which is consistent with the rest of the library).
2. `data: ""` causes an empty param _is_ passed, which i assume is what you want. Unfortunately, the Portal API appears to ignore it (even if you send the additional parameter below).

```js
addItemJsonData({
  authentication,
  id,
  data: "",
  params: { clearEmptyFields: "text,data" } 
});
```

because of the hair pulling i did in #474, i can safely say that `clearEmptyFields` is only honored by string fields. you can pass a `","` to clear `tags` but i don't know of another way to clear item `data` (other than passing `{}`).

ssylvia:
Thanks @jgravois. I was originally testing by looking at the output in AGO Assistant. With further investigation on my end, it looks like passing `null` is not what I want as you suspected. I reached out to few members of the Portal API team and they were unaware of a method to actually clear the itemData once it's been added. 

The empty text string is probably what we want. I'll submit an issue with the Portal API team to get this added. I'll close this issue for now.

jgravois:
[...]

Issue #384, opened 2018-11-08
Param builders for building complex parameter objects


patrickarlt:
Given that this library now has the capacity to do routing https://github.com/Esri/arcgis-rest-js/pull/382 and [create feature services](https://esri.github.io/arcgis-rest-js/api/feature-service-admin/createFeatureService/) I want to put forward the idea of including some reasonable APIs for building the parameters objects like what I specced for https://github.com/Esri/esri-leaflet-routing/blob/master/Route.md and https://github.com/Esri/esri-leaflet-routing/blob/master/TravelArea.md. THis would make the job of building out the params or options objects much more reasonable for things like `IAddToServiceDefinitionRequestOptions` to wrap common use cases. For example:

```typescript
import { addToServiceDefinition, layerBuilder } from '@esri/arcgis-rest-feature-service-admin';

const layer = layerBuilder()
  .field({/*...*/})
  .field({/*...*/})
  .field({/*...*/})
  .spatialReference(4326)
  .enableEditing();

addToServiceDefinition(serviceurl, {
  authentication: userSession,
  layers: [layer]
});
```

Opening this up for comment, @noahmulfinger @araedavis @tomwayson @dbouwman @jgravois.

dbouwman:
Having fluent builders would certainly help us bury more platform idiosyncrasies, so that alone is a win.

```js
import { search, queryBuilder } from '@esri/arcgis-rest-items';

// ability to construct a template that can later be refined... 
// a sort of partial-application on an object graph
const texasWaterWebApps = queryBuilder()
   .withAuth(myAuthFromSessionOrWherever)
   .hasType('Web Mapping Application')
   .withoutType('Web Map')
   .hasTag('water')
   .hasTag('texas')
   .culture('en-us')
   .hasKeyword('hubSolution')
   .withoutKeyword('hubSolutionTemplate');

// later... given some user input...
const query = queryBuilder(texasWaterWebApps)
    .search('waco');

return search(query).then(...)
```

jgravois:
i think this is a _great_ idea.

patrickarlt:
@tomwayson @dbouwman I'm really interested in writing at least the first one of these for building a search `query` parameter after running through the horror of https://github.com/Esri/arcgis-rest-js/blob/master/demos/node-cli-item-management/index.js#L101-L132.

[...]

Issue #339, opened 2018-09-26
facilitate composition of our low level functional API to reduce end user friction


tomwayson:
One area of friction with our functional approach is that sometimes multiple requests that are sequential or at least made w/in the same context will share parameters that will need passed into each fn call. For example, when working w/ a feature service, you may want to call [queryFeatures](https://esri.github.io/arcgis-rest-js/api/feature-service/queryFeatures/), [getFeature](https://esri.github.io/arcgis-rest-js/api/feature-service/queryFeatures/), and the [getAttachments](https://esri.github.io/arcgis-rest-js/api/feature-service/getAttachments/) in succession, all w/ the same `url` (and maybe the same `authentication`).

An OO solution would be to add `new FeatureService({ url })` that would expose methods that wrap the stateless functions passing in `this.url` to each. I'm pretty sure none of us (least of all @dbouwman)  want to get into that business due to the risk of memory leaks etc associated w/`new`ing up instances in browsers.

Instead, I propose we provide at least an example somewhere of how to use functional techniques like higer order functions to compose our low level functional API and reduce the above friction. For example: 

```js
import { queryFeatures, getFeature } from '@esri/arcgis-rest-feature-service';

// create functions to query the trees service
const treesUrl = 'https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0';
const queryTrees = withUrl(treesUrl, queryFeatures);
const getTree = withUrl(treesUrl, getFeature);
```

Here's a working JSBin w/ what I have in mind: https://jsbin.com/geravah/edit?html,js,console

I think the question for me is where to include this information? A guide topic w/ a linked JSBin? Or would we go so far to include common helpers like `withUrl()` and/or `withAuthentication()` in the libraries? If so, in what package?

tomwayson:
Or do we only provide a `withOptions(commonRequestOptions, fn)` that would work at least on any fn that has only a single requestOptions argument?

dbouwman:
100% agree with this sentiment and approach. In `hub.js` I've put functional helpers in the `hub-common` module, but have not added higher order fn's like this yet. 

Perhaps put the `withAuthentication` in the `-authentication` package...
And I think the `withUrl` seems to fit w/ the feature-service package as that's really where we'd be manipulating urls

tomwayson:
Thanks @dbouwman.

[...]

PR #505, opened 2019-04-08
fix: catch err internally when getItemData doesnt find any

jgravois:
AFFECTS PACKAGES:
@esri/arcgis-rest-items

ISSUES CLOSED: #504

### DIFF
--- a/packages/arcgis-rest-items/src/get.ts
+++ b/packages/arcgis-rest-items/src/get.ts
@@ -75,7 +75,14 @@ export function getItemData(
     options.params.f = null;
   }
 
-  return request(url, options);
+  return request(url, options).catch(err => {
+    /*
+     * if the item doesn't include data, the response
+     * will be empty and the internal call to
+     * response.json() will fail with the error below
+     */
+    if (err.message === "Unexpected end of JSON input") return;
+  });
 }
 
 export interface IGetRelatedItemsResponse {
--- a/packages/arcgis-rest-items/src/get.ts
+++ b/packages/arcgis-rest-items/src/get.ts
@@ -76,13 +76,13 @@ export function getItemData(
   }
 
   return request(url, options).catch(err => {
+    /* if the item doesn't include data, the response will be empty
+       and the internal call to response.json() will fail */
+    const emptyResponseErr = RegExp(
+      /Unexpected end of (JSON input|data at line 1 column 1)/i
+    );
     /* istanbul ignore else */
-    if (err.message.indexOf("Unexpected end of JSON input") > -1) {
-      /*
-       * if the item doesn't include data, the response
-       * will be empty and the internal call to
-       * response.json() will fail with the error below
-       */
+    if (emptyResponseErr.test(err.message)) {
       return null;
     } else throw err;
   });
--- a/packages/arcgis-rest-items/src/get.ts
+++ b/packages/arcgis-rest-items/src/get.ts
@@ -83,7 +83,7 @@ export function getItemData(
     );
     /* istanbul ignore else */
     if (emptyResponseErr.test(err.message)) {
-      return null;
+      return;
     } else throw err;
   });
 }
--- a/packages/arcgis-rest-items/src/get.ts
+++ b/packages/arcgis-rest-items/src/get.ts
@@ -55,7 +55,7 @@ export function getItem(
  * getItemData("ae7", { authentication })
  *   .then(response)
  * ```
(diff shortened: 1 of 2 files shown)


patrickarlt:
@jgravois I'm assigning myself as a reviewer on this one. I actually write some code that watches for this error so I want to see how it behaves.

patrickarlt:
@jgravois ok here is the code I wrote in node:

```js
  try {
    config = await getItemData(options.itemId, {
      authentication: session
    });
  } catch (e) {
    if (e.message === "500: Item does not have a file.") {
      config = {};
    } else if (e.type === "invalid-json") {
      config = {};
    } else {
      log.error(logSymbols.error, "error getting config:", e.message);
      log.trace(e.stack);
      process.exit(1);
    }
  }
```

@jgravois this is actually a breaking change because I want to set `config` to {} if there is an error. The problem here is what should we assume the value should be when there is no data or invalid data? You have it defined as `null` which is technically valid JSON. So  if my item has a data value is actually `null` that is indistinguishable from it not having a value at all which semantically might be more like `undefined` which isn't a valid JSON value. Furthermore this will still throw a syntax error on invalid JSON.

I also complicated things by checking `e.type === "invalid-json"` which isn't actually standard code it is something the `isomorphic-fetch` adds so I'm not really doing it right either.

I'm kind of on the fence as to what I would like to do so I'm going to call in @dbouwman and @tomwayson on this one.

1. Keep your code as it return `null` even though `null` could actually be the value of the data
2. Return `undefined` since nothing was defined in the JSON
3. Don't merge this and keep throwing a syntax error and write some doc for this edge case so people don't run into this.
4. Define a new error type for invalid/no JSON and throw that instead so we can better document this.

tomwayson:
I'm in favor of 

> 3. Return undefined since nothing was defined in the JSON

I was already leaning that way yesterday as @jgravois and I were discussing this.

jgravois:
> I'm in favor of `undefined` since nothing was defined in the JSON

i'm just a caveman geographer, but i assumed the goal was to:

1. avoid forcing the caller to `catch` this error themselves
2. provide an opportunity to check that the response actually contains something

```js
getItemData("fe8")
  .then(response => {
    if (response) {
      // do something with actual JSON
    } 
  })
```
`undefined` and `null` are indistinguishable in that respect.

> return `null` even though `null` could actually be the value of the data

its just trivia, but when @ssylvia and i discussed this in https://github.com/Esri/arcgis-rest-js/issues/495#issuecomment-477306765 we came to the conclusion that there is no way to nullify the `/data` for an item after something has been injected. even if this _were_ possible, i can't think of a scenario where the caller would benefit from us not conflating the two.

i'm just playing devil's advocate though. i don't have a horse in the race, so i've pushed a commit to just `return` and check for `undefined` in the unit test.

caveman lawyer

patrickarlt:
@jgravois I'm fine with `undefined` as the return value here. But 2 things:

1. This technically might be a REALLY minor breaking change.
2. @jgravois will you update the docs on `getItemData()` to reflect that this will be `undefined` is the item has no data.

tomwayson:
To clarify - I was wishy-washy between `null` and `undefined` yesterday, but _leaning towards_ `undefined`. I see pros and cons of each.

The original ask from @qlqllu was for `null`.

jgravois:
> will you update the docs on getItemData()

done.

> This technically might be a REALLY minor breaking change.

you both know i've been itching for an excuse to rip off the band-aid, clean up all the loose ends in #137 and tag `v2.0.0` anyway. if you want to give me the green light to go through and tackle all the rest, just say the word.

@patrickarlt had expressed some interest in taking a crack at #339 and #384 to make sure no _other_ breaking changes would be necessary too though. is this a good time for that?

patrickarlt:
> @patrickarlt had expressed some interest in taking a crack at #339 and #384 to make sure no other breaking changes would be necessary too though. is this a good time for that?

@jgravois I might try to prototype them this week if I have a chance.

tomwayson:
I want to keep that bandaid on forever.

To me this is not a breaking change. I would not say it was intentional for this library to `throw` when there was no data in an item. I think it would be a mistake to have this be the straw that breaks John's beard.

jgravois:
> To me this is not a breaking change.

fwiw, i agree with @tomwayson. i was only capitulating for the sake of 'ulterior motives'.

patrickarlt:
Ok lets merge this and :shipit:

PR #506, opened 2019-04-09
fix: loosen peerDependencies to document the oldest compatible package

jgravois:
### tl:dr

it has been an error on my part to sync the `peerDependencies` manually with the `devDependencies` that lerna bumps automatically. we should be advertising the _lowest_ compatible version, not the most recent.

---

@jPurush shared a thread with me the other day that made me realize:

* lerna is making a conscious choice not to bump the `peerDependencies` for each package 
* if it _is_ necessary to bump a `peerDependency` to exclude a previously compatible version, its likely that a breaking change is being introduced.

for example, we should have tagged `v2.0.0` when we landed #400 because we introduced a dependency on a new `cleanUrl()` method inside the request package in existing functions in _other_ packages.

folks even reported uncaught errors to me when `cleanUrl()` wasn't found in their own projects that were resolved by upgrading `request`. i knew enough about what was going on to recommend the fix to people but not enough to realize that we'd broken our contract.

getting in here and trying to figure out exactly what the minimum requirements were for our packages actually made the relationships _more_ clear to me.

besides that, going forward we'll benefit from the lack of rote manual version bumping.

reference:
* https://github.com/lerna/lerna/issues/1575#issuecomment-420380754
* https://github.com/lerna/lerna/issues/1018
* example in [babel](https://github.com/babel/babel/blob/282129ea66dcdbfc9093eb8ff2278bed1127c861/packages/babel-plugin-proposal-do-expressions/package.json)

AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-feature-service-admin
@esri/arcgis-rest-geocoder
@esri/arcgis-rest-groups
@esri/arcgis-rest-items
@esri/arcgis-rest-routing
@esri/arcgis-rest-sharing
@esri/arcgis-rest-users

### DIFF
--- a/packages/arcgis-rest-auth/package.json
+++ b/packages/arcgis-rest-auth/package.json
@@ -20,8 +20,8 @@
     "@esri/arcgis-rest-request": "^1.19.2"
   },
   "peerDependencies": {
-    "@esri/arcgis-rest-common-types": "^1.19.2",
-    "@esri/arcgis-rest-request": "^1.19.2"
+    "@esri/arcgis-rest-common-types": "^1.1.2",
+    "@esri/arcgis-rest-request": "^1.0.0"
   },
   "scripts": {
     "prepare": "npm run build",
--- a/packages/arcgis-rest-feature-service-admin/package.json
+++ b/packages/arcgis-rest-feature-service-admin/package.json
@@ -22,10 +22,10 @@
     "@esri/arcgis-rest-request": "^1.19.2"
   },
   "peerDependencies": {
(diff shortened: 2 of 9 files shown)


noahmulfinger:
@jgravois Just thinking out loud, but I wonder if there's a way we can test packages with both the latest and oldest compatible version of their dependent packages, so stuff like `cleanUrl` doesn't slip through the cracks.

jgravois:
i'm still glad i did the research, but this PR is irrelevant now.

PR #507, opened 2019-04-10
fix(add resource): make resource an optional parameter

noahmulfinger:
Resources can be added with just text content, so the resource parameter should be optional.

AFFECTS PACKAGES:
@esri/arcgis-rest-items

@jgravois Making `resource` optional means that `IItemResourceAddRequestOptions` is not necessary, so I marked it as deprecated. Let me know if that makes sense.

### DIFF
--- a/packages/arcgis-rest-items/src/add.ts
+++ b/packages/arcgis-rest-items/src/add.ts
@@ -23,6 +23,9 @@ export interface IItemDataAddRequestOptions extends IItemIdRequestOptions {
   data: any;
 }
 
+/**
+ * Deprecated. Please use `IItemResourceRequestOptions` instead.
+ */
 export interface IItemResourceAddRequestOptions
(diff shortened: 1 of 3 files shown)


jgravois:
also, you're okay with waiting to release this until after we finish hemming and hawing in #506 right?

noahmulfinger:
@jgravois How's that? And yeah I'm fine with waiting a bit for a release. My current workaround of passing `resource: undefined` works for now.

jgravois:
> i used to say my favorite PRs had more red than green.

wow. i didn't say that. @tomwayson did: https://twitter.com/tomwayson/status/665013115310878720

so i rip off his code _and_ his philosophy. 🏄‍♀️🎧

tomwayson:
At this point, I can't tell where Tom ends and John begins.

PR #509, opened 2019-04-11
Search param builder

patrickarlt:
After some initial research in https://github.com/Esri/arcgis-rest-js/issues/384#issuecomment-482220691 I decided to implement a builder for search queries. Here is some basics of the API:

```typescript
new SearchQueryBuilder()
  .match("test")
  .in("tags"); // "tags: test"

new SearchQueryBuilder()
  .startGroup()
    .match("test")
    .in("tags")
  .endGroup()
  .not()
  .match("seed")
  .in("tags"); // "(tags: test) NOT tags: seed"

const NOW = Date.now();
const FIVE_DAYS_AGO = new Date(Date.now() - (24*60*60*1000) * 5);

new SearchQueryBuilder()
.from(FIVE_DAYS_AGO)
.to(NOW)
.in("created")
.and()
.match("test")
.in("tags"); // "created: [1554570502723 TO 1555002502723] AND tags: test"

new SearchQueryBuilder()
.match("patrickarlt7104")
.in("owner")
.and()
.startGroup()
  .match("Web Mapping Application")
  .in("type")
  .or()
  .match("Mobile Application")
  .in("type")
  .or()
  .match("Application")
  .in("type")
.endGroup()
.and()
.match("VTS")
.in("*"); //owner: patrickarlt7104 AND (type: Web Mapping Application OR type: Mobile Application OR type: Application) AND fire

const query = new SearchQueryBuilder()
.match("patrickarlt7104")
.in("owner")
.and()
.startGroup()
  .match("Web Mapping Application")
  .in("type")
  .or()
  .match("Mobile Application")
  .in("type")
  .or()
  .match("Application")
  .in("type")
.endGroup(); // "owner: patrickarlt7104 AND (type: Web Mapping Application OR type: Mobile Application OR type: Application)"

query.clone()
  .and()
  .match("fire")
  .in("*"); // "owner: patrickarlt7104 AND (type: Web Mapping Application OR type: Mobile Application OR type: Application) AND fire"
```

Instead of trying to implement all sorts of crazy helpers like `andHasTags` and `orHasTags` ect... I sent for a simpler approach that matches what the REST API search query thing is. This simply tries to build the string as best as possible and should allow for almost ANY permutation of what can be passed to the search API.

In the end there is no way to type option 2 from https://github.com/Esri/arcgis-rest-js/issues/384#issuecomment-482220691 without a TON of extra work (it might not even be possible at all) so I went with a class based approach.

I think that eventually we will want both `IParamBuilder` for building a single param and `IParamsBuilder` for returning an object with multiple params. This PR limits itself to `IParamBuilder` for now and leaves `IParamsBuilder` open for things like routing or feature layer/webmap creation later. I think that for building complex JSON objects we could use something like https://medium.com/@bensammons/building-a-fluent-interface-with-typescript-using-generics-in-typescript-3-4d206f00dba5 would be :100: but I'll play around with it later.

This PR is still in draft while:

1. We discuss the general API around `IParamBuilder`
1. We discuss the specific API of `SearchQueryBuilder`
1. I harden the API of `SearchQueryBuilder` to be more forgiving of stupid things like `new SearchQueryBuilder().endGroup().and().or().from("foo").and().not("bar")`
1. I write tests for `SearchQueryBuilder`
1. I write tests for `IParamBuilder`
1. Write more documentation around
   1. Adding more builders for other params
   2. A guide topic to using the `SearchQueryBuilder`
1. Replace the hellscape of https://github.com/Esri/arcgis-rest-js/blob/master/demos/node-cli-item-management/index.js#L102-L132 with the new builder


jgravois:
cc/ @pranavkulkarni

pranavkulkarni:
This is awesome. I am currently porting some of the existing query building logic opendata-ui to hub.js/search. I could use this SearchQueryBuilder which is very intuitive and will help develop hub.js better. Any idea when this will roll out tentatively?

jgravois:
our goal is to include this in a `v2.0.0` release fairly quickly.

my _hope_ would be to ship by late next week.

if you could symlink pat's branch in the meantime and provide more detailed feedback, that would be _really_ cool.

patrickarlt:
@pranavkulkarni I would really want to land this in 2.0.0 next week.

@jgravois I'm probably going to be in Redlands sometime next week but I might only have 1-2 meetings do you and maybe @tomwayson want to spend a day on releasing 2.0.0?

jgravois:
> do you and maybe @tomwayson want to spend a day on releasing 2.0.0?

for. sure.

packages/arcgis-rest-request/src/utils/process-params.ts
```diff
@@ -46,7 +50,12 @@ export function processParams(params: any): any {
   const newParams: any = {};
 
   Object.keys(params).forEach(key => {
-    const param = params[key];
+    let param = params[key];
+
+    if (param.toParams) {
```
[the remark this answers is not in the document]
jgravois: i agree. naming the method `toParam()` makes more sense.

patrickarlt:
Since most people seem ok with the API I went ahead a wrote a pretty full test suite and added a bunch of warnings when people do stupid things like:

```js
const query = new SearchQueryBuilder()
  .match("test")
  .not()
  .and()
  .toParam();
```

I would still like to write at least API doc for this. I'll do written doc some time after this merges. I also might replace `IParamBuilder` with something more like `IOptionsBuilder` and have it build out `IRequestOptions`. Then make a custom `SearchOptions` builder class just to demonstrate the other levels of builders. I'm just not 100% on board with making people leverage `params`.

patrickarlt:
Since this will go into 2.0.0 I'm also going to handle https://github.com/Esri/arcgis-rest-js/issues/137#issuecomment-482238840.

patrickarlt:
@jgravois @tomwayson OK I have done some major refactoring of `searchItems`. This is technically for 2.0.0 but I'm making the PR against `master` and @jgravois and I can figure it out from there.

Does the following:

1. Add support for a `IParamBuilder` which will allow for custom builder classes to provide the value of a single param in the API.
2. Concrete implementation of `IParamBuilder` in `SearchQueryBuilder` for the `q` param in search items.
3. Rewrite `appendCustomParams` and move it to common. `appendCustomParams` now accepts a custom `IWhateverOptions` object and a whitelist of keys to copy to params and returns the finished `IRequestOptions` object to use. You can optionally pass in a default set of options which will be merge before the keys are copied over.

This has 100% test coverage and should be ready to merge. 

@jgravois feel free to refactor this and change the base to the 2.0.0 branch.

packages/arcgis-rest-request/src/utils/IParamsBuilder.ts
```diff
@@ -0,0 +1,5 @@
+import { IParams } from "./params";
+
+export interface IParamsBuilder {
+  toParams(additionalParmas?: IParams): IParams;
```
tomwayson: `additionalParams`

packages/arcgis-rest-common/package.json
```diff
@@ -13,7 +13,8 @@
     "dist/**"
   ],
   "dependencies": {
-    "tslib": "^1.9.3"
+    "tslib": "^1.9.3",
+    "@esri/arcgis-rest-request": "^1.19.2"
```
tomwayson: should this be a `devDependency`?

packages/arcgis-rest-items/src/search.ts
```diff
@@ -29,6 +27,7 @@ export interface ISearchResult {
   num: number;
   nextStart: number;
   results: IItem[];
+  nextPage?: () => Promise<ISearchResult>;
```
tomwayson: This is clever. Should we also include a `previousPage()`?

jgravois:
i just pushed up two more commits with a handful of tweaks to get tests passing on CI and use @patrickarlt's new appendCustomParams method in `feature-service` and `geocoder`.

doing this helped me catch (and fix) a couple more edge cases in the refactored utility method.

if @tomwayson is happy, i can rebase this branch on top of v2.0.0 and resolve the conflicts.

packages/arcgis-rest-common/package.json
jgravois: i'd say yes. fixed in 99300a8

packages/arcgis-rest-items/src/search.ts
jgravois: gotta save something for 2.1!

patrickarlt:
@jgravois your commits LGTM. I'll let you refactor this into 2.0.0.

PR #510, opened 2019-04-11
remove deprecated crud params for v2

COV-GIS:
https://github.com/Esri/arcgis-rest-js/issues/137


COV-GIS:
I've never changed the base before. If it causes problems let me know and I'll pull v2.0.0 down and branch it for a new PR.

COV-GIS:
thanks for the homonym fix @jgravois. it didn't look right but ran with it and did the ol' copy and paste routine.

PR #521, opened 2019-04-18
Refactor feature-layer/ I*Params to I*Options

tomwayson:
more consistent naming for interfaces in the feature-layer package

AFFECTS PACKAGES:

BREAKING CHANGE:
renamed IEditFeaturesParams to ISharedEditOptions & ISharedQueryParams to ISharedQueryOptions;
removed IDeleteFeaturesParams


tomwayson:
addresses https://github.com/Esri/arcgis-rest-js/pull/520#pullrequestreview-227958016

tomwayson:
while doing this I noticed that `IDeleteFeaturesResult` is inconsistent w/ the other `I*Response` interfaces

image

but I assume we're going to get rid of of all of those as per #137, so I didn't address that in this PR.

Speaking of, do you want me to do ^^^ for the feature-layer package, @jgravois? I assume it would be a matter of just removing the interface and inlining the shape of the response into the function declaration's `Promise<>`, like:

```ts
export function deleteFeatures(
  requestOptions: IDeleteFeaturesRequestOptions
): Promise<{ deleteResults?: IEditFeatureResult[]; }> {
// instead of Promise<IDeleteFeaturesResult>
```

right?

jgravois:
> right?

right.

COV-GIS:
@jgravois  @tomwayson the reason add, update and delete had different result interfaces is ags returns different for each. i haven't been following the rework closely so maybe it doesn't matter. but if there is going to be some sort of shared result interface just be aware the result is different.

tomwayson:
> but if there is going to be some sort of shared result interface just be aware the result is different.

No shared interface, just each fn will declare inline what it returns.

COV-GIS:
Good deal.

On hold starting a couple new projects to get up to speed on v2. I appreciate all the debate and hard work for my benefit @tomwayson @jgravois @patrickarlt. :)

tomwayson:
sounds like we've got our first beta tester!

tomwayson:
response interfaces removed in #523

packages/arcgis-rest-feature-layer/src/helpers.ts
```diff
@@ -19,7 +19,7 @@ export interface ILayerRequestOptions extends IRequestOptions {
   url: string;
 }
 
-export interface ISharedQueryParams {
+export interface ISharedQueryOptions {
```
jgravois: i was thinking this:

```ts
export interface ISharedQueryOptions extends ILayerRequestOptions {
```

so that you could do

```ts
export interface IAddFeaturesRequestOptions
  extends ISharedEditOptions {
```

tomwayson:
Yeah, I'd love to pull "Request" from all the options. I don't think the work would be bad, it would just be the fact that we would be moving _all_ the 🧀 , which shouldn't affect too many users. I would guess that even TS users pass POJOs to the fns most of the time.

jgravois:
> I'd love to pull "Request" from all the options.

its a wonder how _either of us_ gets any real work done.

tomwayson:
I don't think that we do.

patrickarlt:
@jgravois  I'm going to make a sticker of this:

> my OCD rules everything around me.

Issue #518, opened 2019-04-17
v2.0.0 doc improvements


patrickarlt:
Given the volume of changes in 2.0.0 there are a lot of doc related tasks. We have also gotten a lot better at writing TypeScript and thus the custom typedoc rendering can't handle our fancy new typings.

#### Guide

* [x] Written guide for new 2.x features (@jgravois or @tomwayson)
* [ ] Write a new v1.x > 2.x migration guide (@jgravois or @tomwayson)
* [x] Guide for composition using `withOptions()`
* [x] Guide for defaults using `setDefaultRequestOptions()`
* [x] Update the package overview page for the new packages  (@jgravois or @tomwayson)
* [x] Guide for searching for items using `SearchQueryBuilder`.
* [x] Consolidate the WebPack/Rollup guides into a single guide (possibly also add parcel?)  (@jgravois or @tomwayson)
* [ ] Add a Typescript specific guide to talk about the common types
* [x] Add a guide for tree shaking (specifically webpack) (@jgravois)
* [x] Remove the browserify topic  (@jgravois or @patrickarlt)
* [ ] Add guide for contributors to wrapping an API endpoint (@patrickarlt)
* [ ] Add guide for writing a new param builder (@patrickarlt)

#### API Ref

* [x] Link the REST API page for each method (@jgravois)
* [x] API ref for `SearchQueryBuilder` (@patrickarlt)
* [x] API ref for `withOptions()` (@patrickarlt)
* [x] API ref for `setDefaultRequestOptions()` (@patrickarlt)
* [x] Fix formatting for return type of `updateGroup()` also https://github.com/Esri/arcgis-rest-js/pull/523). (@patrickarlt)
* [x] Handle type params for `ISearchResponse` (@patrickarlt)
* [x] Render `results` in `ISearchResponse` properly (@patrickarlt)
* [x] Render values properly for things like `NODEJS_DEFAULT_REFERER_HEADER`. (@patrickarlt)
* [x] Sort parameter tables by (@patrickarlt)
  1. Required
  1. Required + Inherited
  1. Optional
  1. Optional + Inherited
* [x] Sort the lists of exported items in the API reference for clarity (@patrickarlt)
   1. Classes
   2. Functions
   3. Constants
   4. Enums
   5.  Interfaces

jgravois:
we're going to need _at least_ one demo to show off `SearchQueryBuilder`.

it'd be nice to be able to share a live link, instead of point folks at our /demos folder.

[...]

PR #523, opened 2019-04-18
refactor(feature responses): removes unneccessay feature layer respon…

tomwayson:
…se interfaces

AFFECTS PACKAGES:

BREAKING CHANGE:
removes IAddFeaturesResult, IAddAttachmentResponse, IDeleteFeaturesResult,
IDeleteAttachmentsResponse, IGetAttachmentsResponse, IUpdateFeaturesResult, &
IUpdateAttachmentResponse

### DIFF
--- a/packages/arcgis-rest-feature-layer/src/add.ts
+++ b/packages/arcgis-rest-feature-layer/src/add.ts
@@ -28,16 +28,6 @@ export interface IAddFeaturesRequestOptions
 }
 
 /**
- * Add features results.
- */
-export interface IAddFeaturesResult {
-  /**
-   * Array of JSON response Object(s) for each feature added.
-   */
-  addResults?: IEditFeatureResult[];
-}
-
-/**
  * ```js
  * import { addFeatures } from '@esri/arcgis-rest-feature-layer';
  * //
@@ -57,7 +47,7 @@ export interface IAddFeaturesResult {
  */
 export function addFeatures(
   requestOptions: IAddFeaturesRequestOptions
-): Promise<IAddFeaturesResult> {
+): Promise<{ addResults?: IEditFeatureResult[] }> {
   const url = `${cleanUrl(requestOptions.url)}/addFeatures`;
 
   // edit operations are POST only
--- a/packages/arcgis-rest-feature-layer/src/addAttachment.ts
+++ b/packages/arcgis-rest-feature-layer/src/addAttachment.ts
@@ -20,16 +20,6 @@ export interface IAddAttachmentOptions extends ILayerRequestOptions {
 }
 
 /**
- * `addAttachment()` request response.
- */
-export interface IAddAttachmentResponse {
-  /**
-   * Standard AGS add/update/edit result Object for the attachment.
-   */
-  addAttachmentResult: IEditFeatureResult;
-}
-
-/**
  * ```js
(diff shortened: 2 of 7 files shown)


tomwayson:
I didn't touch the query response interfaces b/c if we ever decide that we're going to try an accurately capture this [maze of possible outcomes](https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm#GUID-EE6C21C1-03E8-4379-9EF0-44F38AA98779) in types, we're going to need a few interfaces. I'm sure some poor soul on the JSAPI team had to do that, so maybe there's something in [@types/arcgis-js-api](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/arcgis-js-api) that we can use if we every try to tease apart that rats nest.

patrickarlt:
@tomwayson as a heads up the docs don't render this particularly well. It is much easier to render these if we have seperate interface.

tomwayson:
Good point @patrickarlt 

I guess the question is, which is harder, fixing this:

image

Or adding a way to filter the left nav to not show interfaces?

patrickarlt:
@tomwayson I would also have to fix this in the returns section.

Screenshot 2019-04-18 10 21 45

what about my suggestion in #518 for sorting things in order of classes/methods/interfaces?

jgravois:
> sorting things in order of classes/methods/interfaces?

that'd be a **big** improvement.

COV-GIS:
I don't want to monkey wrench here...but why not just just go with `applyEdits` endpoint, have a single `editFeatures` fnc, and get rid of `addFeatures`, `updateFeatures`, and `deleteFeatures`? One method. One response.

tomwayson:
that would be a big improvement, and I don't mind throwing away this PR, but I'd feel better about it if we knew it wouldn't be too big a lift to add filtering after sorting.

tomwayson:
@COV-GIS yeah, we probably _should_ have started w/ that, but I don't think it makes sense at this point to remove the existing fns.

jgravois:
> why not just go with applyEdits endpoint

Because of what you told us in https://github.com/Esri/arcgis-rest-js/pull/185#discussion_r186539051 :smile:

COV-GIS:
I was green.

patrickarlt:
From an API design perspective I REALLY dislike `applyEdits` because it isn't clear and it looks like it only covers editing as opposed to having seperate operations for `addFeatures`, `editFeatures` and `deleteFeatures`

COV-GIS:
I guess I don't think about the big picture too much being primarily a user @patrickarlt. Good perspective and I'll try to keep it in mind in the future. I suggested `applyEdits` for more than just this issue. For bulk transactions particularly to clean up a dirty store backing a tabular grid or using node to access, process and update services from 3rd party interwebs data `Promise.all().then().catch()` is rabbit hole. `Promise.all().always()` works but takes a ton of logic to ensure success and handle failures. Using `applyEdits` with `request()` is easier and works just fine w/o a dedicated method. Plus bulk attachments. I'll let the dust settle on v2 and maybe propose a method to hit `applyEdits` down the road.

tomwayson:
FWIW - I'll gladly merge an `applyEdits()` PR... _after_ 2.0.0 is released. 

🤐  We don't have to tell @patrickarlt.

PR #527, opened 2019-04-18
V2.0.0

jgravois:
more red than green. more red than green!


jgravois:
i'd typically rebase a branch like this to get rid of all the intermediate commits and merge commits, but given how many people worked on it, i'm also happy to preserve the real history.

tomwayson:
:shpit:.. as is.

PR #529, opened 2019-04-19
refactor(interfaces): removes "Request" from option interface names

tomwayson:
AFFECTS PACKAGES:

BREAKING CHANGE:
Options interfaces in @esri/arcgis-rest-feature-layer have been renamed


tomwayson:
Yep. Right after `routing`.

PR #531, opened 2019-04-19
2.0.0 doc improvements

patrickarlt:
Resolves the following in https://github.com/Esri/arcgis-rest-js/issues/518

* [x] Add support for linking the the REST API page for each method. YOu can now add `@restlink https://developers.arcgis.com/rest/network/api-reference/route-synchronous-service.htm` to add a link to the REST API on teh developers site to any function. See `solveRoute()` for an example.
* [x] Fix formatting for return type of `updateGroup()` also https://github.com/Esri/arcgis-rest-js/pull/523)
* [x] Sort tables
* [x] Sort the lists of exported items in the API reference for clarity
* [x] Render values properly for things like `NODEJS_DEFAULT_REFERER_HEADER`.
* [x] Handle type params for `ISearchResponse`

This also fixes a lot of other small case:

* Tables with long strings now simply scroll left to right (see `UserSession`)
* Each package now includes a sample ES 6 import statement. By default this is all classes and functions but can be customized with a custom `esri.keyExports` field in `package.json` of each package (See `-request`, `-auth` and `-types`)
* Add a blacklist for blocking exports like `genericSearch` and `DEFAULT_ARCGIS_REQUEST_OPTIONS`.
* Update calcite web to latest
* General improvements to the information density of each doc page.

@jgravois this wraps up everything in the API ref that ISN'T actually writing doc. I'll do a separate PR for that later to day if you want to review this.

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -217,22 +217,28 @@ module.exports = function(acetate) {
 
   // <code> friendly script tag string
   acetate.helper("scriptTag", function(context, package) {
-    return `&lt;script src="https://unpkg.com/${
-      package.name
-    }@${package.version}/dist/umd/${package.name.replace("@esri/arcgis-rest-", "")}.umd.min.js"&gt;&lt;/script&gt;`;
+    return acetate.nunjucks.renderString(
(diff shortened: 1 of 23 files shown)


jgravois:
ps. you changes look **awesome**!

PR #533, opened 2019-04-19
Fix/handle binary data consistently and deprecate addItemJsonData

jgravois:
i couldn't resist the urge to deprecate `addItemJsonData` and just ensure that `addItemData` handles _both_ JSON object literals and binary files.

when i dug into this, @tomwayson and i realized that there was a bug in `createItem` and `updateItem` that kept _them_ both from passing through binary files appropriately.

this PR:
1. fixes the bug in `createItem` and `updateItem`
2. deprecates `addItemJsonData` and ensures that remaining method (`addItemData`) handles both. keeping things DRY by calling `updateItem` internally.

i promise i'm done now :smile:

### DIFF
--- a/packages/arcgis-rest-portal/src/items/add.ts
+++ b/packages/arcgis-rest-portal/src/items/add.ts
@@ -12,48 +12,13 @@ import {
   determineOwner,
   IManageItemRelationshipOptions
 } from "./helpers";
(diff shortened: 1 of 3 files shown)


jgravois:
at this rate `v2.0.0` is going to be smaller than our last release. 

even with the new features and guide!

PR #541, opened 2019-04-23
Update generate-srihashes.js

COV-GIS:


COV-GIS:
@jgravois not all the packages are getting added to https://github.com/Esri/arcgis-rest-js/blob/master/docs/src/srihashes.json. when I run locally all are added. not a big issue but just something to check on next time you build docs.

jgravois:
> when I run locally all are added.

now that this PR has landed, i'm seeing the same thing.

PR #543, opened 2019-04-23
add aggregations to ISearchResult

pranavkulkarni:


packages/arcgis-rest-portal/src/util/search.ts
```diff
@@ -23,4 +23,5 @@ export interface ISearchResult<T extends IItem | IGroup> {
   nextStart: number;
   results: T[];
   nextPage?: () => Promise<ISearchResult<T>>;
+  aggregations?: { counts: any };
```
jgravois: aggregations aren't very self-explanatory. lets add the doc below (and denote that counts will be an array).

```ts
/**
 * Aggregations will only be present if [`fieldCounts`](https://developers.arcgis.com/rest/users-groups-and-items/search.htm#GUID-1C625F8A-4A12-45BB-B537-74C82315759A) are requested.
 */
aggregations?: { counts: any[] };
```

pranavkulkarni: cool thanks

patrickarlt: @jgravois do you think we should type `counts` more strongly?

jgravois: at first glance i thought the payload was too complicated to bother, but it's actually not.

packages/arcgis-rest-portal/src/util/search.ts
```diff
@@ -23,4 +23,5 @@ export interface ISearchResult<T extends IItem | IGroup> {
   nextStart: number;
   results: T[];
   nextPage?: () => Promise<ISearchResult<T>>;
+  aggregations?: { counts: any[] };
```
jgravois:
```suggestion
  aggregations?: { counts: {
    fieldName: string;
    fieldValues: {
      value: any;
      count: number;
    }[]
  } };
```

patrickarlt: Looks good but I don't think this pass lint.

pranavkulkarni: Thanks guys

PR #556, opened 2019-04-25
ask travis to always run on stable/LTS node.js

jgravois:


jgravois:
i guess its a little early for this.

Issue #559, opened 2019-04-30
Cannot `request` a public layer


jpeterson:
I'm on v2.0.0

I'm running into an [`ArcGISAuthError`](https://esri.github.io/arcgis-rest-js/api/request/ArcGISAuthError/) when trying to request [a public MapServer layer](https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2).

```js
request('https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2',
  {
    httpMethod: 'GET',
    authentication
  }
);
```

My [UserSession's `portal` property](https://esri.github.io/arcgis-rest-js/api/auth/UserSession/#portal): `https://qaext.arcgis.com/sharing/rest`.

Result:

![image](https://user-images.githubusercontent.com/3399064/56981557-64829500-6b3c-11e9-9f6d-49cafa385f39.png)

```json
{
  "name": "ArcGISAuthError",
  "message": "NOT_FEDERATED: https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2 is not federated with any portal and is not explicitly trusted.",
  "originalMessage": "https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2 is not federated with any portal and is not explicitly trusted.",
  "code": "NOT_FEDERATED",
  "url": "https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2",
  "options": {
    "httpMethod": "GET",
    "params": {
      "f": "json"
    },
    "authentication": {
      "clientId": "KU6w5ji2TfnAx5gl",
      "username": "jpeterson_iot",
      "token": "U_HrpG0BBRwPxk17JENdPIvIvooLNGyJez-vYEPiw-w2cpedH9EvAtk1cFbZF-zdp1j2KrVxF23KmGvztgpDtPYuXvvRA1uXmHcIDGaHe9yT8uZqnsEYE6Q9GlCPwN9X5EAkkPZLc8mGv9iH8hiV96Z1863QPqX4JCMbQy9I8Q6m4GF8VH0mPXMxNKO37PzR1wRxCu-cc72Fbg8-g3KzFZ2QejrAFhNja6TfUheG_jk.",
      "tokenExpires": "2019-05-14T16:30:40.047Z",
      "portal": "https://qaext.arcgis.com/sharing/rest",
      "ssl": true,
      "tokenDuration": 20160,
[...]

PR #560, opened 2019-05-01
fix(try): try an anonymous request before throwing an auth error

jgravois:
AFFECTS PACKAGES:
@esri/arcgis-rest-auth
@esri/arcgis-rest-request

ISSUES CLOSED: #559

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -242,7 +242,29 @@ export function request(
     })
     .then(data => {
       if ((params.f === "json" || params.f === "geojson") && !rawResponse) {
-        return checkForErrors(data, url, params, options, originalAuthError);
+        const response = checkForErrors(
+          data,
+          url,
+          params,
+          options,
+          originalAuthError
+        );
+        if (originalAuthError) {
+          /* if the request was made to an unfederated service that 
+          didnt require authentication, add the base url and a dummy token
+          to the list of trusted servers to avoid another federation check
+          in the event of a repeat request */
+          const truncatedUrl: string = url
+            .toLowerCase()
+            .split(/\/rest(\/admin)?\/services\//)[0];
+          (options.authentication as any).trustedServers[truncatedUrl] = {
+            token: [],
+            // default to 24 hours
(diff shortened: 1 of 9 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -139,11 +142,20 @@ export function request(
...
         err.url = url;
         err.options = options;
-        throw err;
+        /**
+         * if an attempt is made to talk to an unfederated server
+         * first try the request anonymously. if a 'token required'
+         * error is thrown, throw the UNFEDERATED error then.
+         */
+        originalAuthError = err;
+        return Promise.resolve("");
```
jgravois: initially i thought we'd need something like this:
```ts
if (err.code === "NOT_FEDERATED") {
  originalAuthError = err;
  return Promise.resolve("");
}
throw err
```
but in updating the tests i couldn't isolate a scenario in which `getToken()` throws a different error.

packages/arcgis-rest-request/src/utils/check-for-errors.ts
```diff
@@ -35,7 +36,11 @@ export function checkForErrors(
       messageCode === "GWM_0003" ||
       (code === 400 && message === "Unable to generate token.")
     ) {
-      throw new ArcGISAuthError(message, errorCode, response, url, options);
+      if (originalAuthError) {
+        throw originalAuthError;
```
tomwayson: ah, great idea. this was my concern about re-trying, that we'd show the wrong error message.

jgravois:
to do:

- [ ] cache the url as a 'trusted' server so that repeat requests won't trigger a federation check.

jgravois:
in c0dad417c6f034ff4583958eba643427612ba8e9 i implemented logic to cache the root server url in the event of a successful anonymous request in order to avoid repeat federation checks.

ref: https://github.com/Esri/arcgis-rest-js/pull/423#issuecomment-449473077

patrickarlt:
@jgravois new changes LGTM.

jgravois:
thanks @patrickarlt!

after i merge this i'm going to cut a `v2.0.1` release.

PR #567, opened 2019-05-14
sapper app fix deps

COV-GIS:


COV-GIS:
you bet @jgravois. it seems to have some other issues with the web map stuffs. but i was mainly interested in the auth component. those fixes will allow it to compile, run and auth at least.

Issue #411, opened 2018-11-29
Do not force node users to require fetch


dmfenton:
It's awkward to have code like below. It's also an anti-pattern to use globals in Node.
```js
require('isomorphic-fetch')
const { getItem } = require('@esri/arcgis-rest-items')
```

Instead node-fetch could be made a dependency and only loaded if a node environment is detected. And rest-request could do something like 
```
if (process) {
  const fetch = require('node-fetch')
}
```

tomwayson:
I'm not sure there's a good way to to make node-fetch an "optional" dependency.

In discussing w/ @dmfenton he suggested to add it to `pacakge.dependencies` but not `import` it, and I presume use the above call to `require()`. We'd have to verify that rollup would tree shake that out, but I suspect it won't.

dmfenton:
Looks like there's this handy thing:
https://www.npmjs.com/package/rollup-plugin-ignore
```js
const emptyFile = 'export default {}';
const emptyFileName = '\0empty_module';

function ignore(list) {
	return {
		resolveId(importee) {
			return list.indexOf(importee) > -1 ? emptyFileName : null;
		},
		load(id) {
			return id === emptyFileName ? emptyFile : null;
		},
	};
}

module.exports = ignore;
ignore.__esModule = true;
ignore.default = ignore;
```

dmfenton:
There is also the built in option `external`
https://rollupjs.org/guide/en#core-functionality
```
external -e/--external
Either a Function that takes an id and returns true (external) or false (not external), or an Array of module IDs that should remain external to the bundle. The IDs should be either:

the name of an external dependency, exactly the way it is written in the import statement. I.e. to mark import "dependency.js" as external, use "dependency.js" while to mark import "dependency" as external, use "dependency"
a resolved ID (like an absolute path to a file)
// rollup.config.js
import path from 'path';

export default {
  ...,
  external: [
    'some-externally-required-library',
    path.resolve( __dirname, 'src/some-local-file-that-should-not-be-bundled.js' )
  ]
};
When given as a command line argument, it should be a comma-separated list of IDs:

rollup -i src/main.js ... -e foo,bar,baz
When providing a function, it is actually called with three parameters (id, parent, isResolved) that can give you more fine-grained control:

id is the id of the module in question
parent is the id of the module doing the import
isResolved signals whether the id has been resolved by e.g. plugins
When creating an iife or umd bundle, you will need to provide global variable names to replace your external imports via the output.globals option.
```

PR #573, opened 2019-05-20
Replace isomorphic-fetch with cross-fetch in node.js guide

pr3tori4n:
**Justification:**
cross-fetch has more up to date dependencies. isomorphic-fetch seems to be unmaintained right now.
https://github.com/lquixada/cross-fetch/blob/master/package.json#L49
https://github.com/matthew-andrews/isomorphic-fetch/blob/master/package.json#L21

**Callouts**
There's several other places in the repo, both in guides and in code that are using isomorphic-fetch. I recommend you update those as well. This is a small start.
Related PR where I'm using cross-fetch succesffuly: https://devtopia.esri.com/WebGIS/arcgis-js-api/pull/20702


jgravois:
thx for this @pr3tori4n!

i'm going to tack on a couple more commits to make sure we're pointing folks toward `cross-fetch` consistently and see what all we'll need to do to tackle #411 at the same time.

jgravois:
Given that neither pollute the global, i _think_ it makes more sense for `node-fetch` to be our general purpose recommendation than `cross-fetch`.

i just pushed up another commit which:

* shows a code snippet that uses `setDefaultRequestOptions` to take advantage of the ponyfill 🎠.
```js
const fetch = require("node-fetch");
const { request, setDefaultRequestOptions } = require("@esri/arcgis-rest-request");

// use node-fetch for each request instead of relying on a global
setDefaultRequestOptions({ fetch })
```
* updates the Node.js demos to use `"node-fetch"`
* updates the test suite to use `"node-fetch"`
* mentions `"cross-fetch"` in the context of Universal/Isomorphic apps

this addresses @dmfenton's concerns in #411

patrickarlt:
From an ease of use perspective I REALLY prefer just using global fetch even in node.

`cross-fetch` allow you to pull in `node-fetch`:

```js
import fetch from 'cross-fetch'; // this is really just node-fetch

request(someUrl, {fetch}).then(r => {
  console.log(r)
});
```

or as a global:

```js
import 'cross-fetch/polyfill';

request(someUrl).then(r => {
  console.log(r)
});
```

 I think we can promote both approaches in the guide here.

jgravois:
> I think we can promote both approaches in the guide here.

thats fine with me.

something in cross-fetch's Response implementation must be _slightly_ different than what fetch-mock returns because several tests with [`rawResponse`](https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/#rawResponse) started failing when i swapped in the new dependency. 

```js
expect(response instanceof Response).toBe(true);
```
nothing blocking though, i just replaced them with something more verbose (and arguably a more useful)

```js
expect(response.status).toBe(200);
expect(response.ok).toBe(true);
expect(response.body.Readable).not.toBe(null);
response.json().then((raw: any) => {
  expect(raw).toEqual(cannedResponse);
  done();
});
```

jgravois:
i'd rebase and fix the conflict, but i don't have push access to the repo anymore.

💋

PR #585, opened 2019-06-17
 feat(portal): new addGroupUsers() function

### LINKED ISSUE #584: Add functions to add users/admins to a group
The group [addUsers](https://developers.arcgis.com/rest/users-groups-and-items/add-users-to-group.htm) operation is not yet covered in this library and it is an important feature to add.

The ArcGIS REST API has one single endpoint to add both users and admins. But for simplicity, we can add two functions `addGroupAdmins(groupId, admins)` and `addGroupUsers(groupId, users)`, which take an array of usernames as the second parameter.

A limitation of the original endpoint is that it only accepts up to 25 usernames per request and throws an error if the developer passes a long array. In practice, the developer needs to send multiple requests if there are many group members to add. The logic is like this:

``` javascript
// pass a groupId and a long array of usernames
function addManyGroupUsers (groupId, usernames) {
  // break the long array into smaller ones, each has up to 25 elements
  const chunks = _.chunk(usernames, 25);

  // send multiple requests at the same time (or maybe sequentially)
  const promises = chunks.map((users) => sendAddUserRequest(groupId, users));

  return Promise.all(promises);
}
```

This batch request logic can be included the new functions and the user can forget the endpoint limitation.

The problem of batching requests is from the request error. Fortunately, the `addUsers` operation is idempotent and the developer can retry many times until succeed, without worrying about duplicate members. So I think it can simply fail the whole function if one request fails.

Otherwise, the request error can be caught as

```  javascript
const promises = chunks.map(
  (users) => sendAddUserRequest(groupId, users)
    .catch(err => {
      return {
        // these usernames are not added because of the request failure
        notAdded: chunks,
        // it needs a flag to indicate whether the "notAdded" is from a request
        // failure or a user issue (already added, group member limit reached,
        // etc)
        requestFailure: true
      }
    })
);
```

The original issue is posted at https://github.com/Esri/ember-arcgis-portal-services/issues/157

@tomwayson for review

haoliangyu:
resolve #584


packages/arcgis-rest-portal/src/groups/add-users.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+function _sendSafeRequest(
+  url: string,
+  requestOptions: IAddGroupUsersOptions
+): Promise<IAddGroupUsersResult> {
+  return request(url, requestOptions).catch(error => {
+    return {
+      // the request should either add users or admins, not both
+      notAdded: requestOptions.users || requestOptions.admins,
```
tomwayson: I don't think we should populate `notAdded` when there are errors. We don't want them to get mixed up w/ any that were not added b/c they were already  in the group.

haoliangyu: I have updated the code

Issue #588, opened 2019-06-19
f=geojson not returning geojson for queryFeatures


apfister:
I saw this https://github.com/Esri/arcgis-rest-js/commit/79bc776ca91ddca4685eef2cf4c8d1ed2f4a4117, but i'm not able to return geojson using the example in the docs:

```
require("isomorphic-fetch");
require("isomorphic-form-data");

const {queryFeatures} = require('@esri/arcgis-rest-feature-layer');

queryFeatures({
  url: "http://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/3",
  where: "STATE_NAME = 'Alaska'",
  f: 'geojson',
}).then(result => {
    // result is still in arcgis json
    console.log(result);
});
```

apfister:
if i add `f` to the `queryOptions` list here, https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-feature-layer/src/query.ts#L189, geojson is returned. is this something that's an easy fix?

COV-GIS:
`f` is a request option https://esri.github.io/arcgis-rest-js/api/request/IParams/ not a `queryFeatures` option.

```javascript
queryFeatures({
  url: "http://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/3",
  where: "STATE_NAME = 'Alaska'",
  params: {
    f: 'geojson'
  }
})
```

apfister:
@COV-GIS hmm, thanks and I can see that does work. It is a bit confusing though because the query parameters for a feature service aren't usually nested like this when making a REST call to the service. I assumed that params like `url` `where` _and_ `f` would be on the same "level", if that makes sense.

COV-GIS:
It's a pretty rare case that `f` isn't `json`. So it's baked in as the only real default as well as `f` being the only required param for all REST calls. The helper methods like `queryFeatures` simply expose common params for that particular REST call as top level params. With the exception of `url`, they get mixed into `params` for the request itself. `params` allows you to add any additional params which are not exposed or custom ones. You could add `where` to `params` and it will work just fine. As a stylistic choice and because `params` often times gets constructed thru several functions, I put all options in `params` and simply use the helper methods to construct all the proper endpoint urls beyond the feature service url.

```javascript
const params = {
  outFields: '*',
  time: `${start}, ${end}`,
[...]

PR #589, opened 2019-06-22
add f to query options whitelist

apfister:
addresses #588 to help support returning geojson from queries


jgravois:
travis is complaining because you need to add the `f` parameter to `IQueryFeaturesOptions` too.

```ts
export interface IQueryFeaturesOptions extends ISharedQueryOptions {
  // ...
  f: "json" | "geojson" | "html" | "pbf";
}
```

tomwayson:
@apfister - the fix for this was released in v2.1.1

PR #592, opened 2019-06-25
arcgis-rest-feature-layer - geometry not getting passed in to queryFeatures

gavinr:
if you pass in any of the following properties they do not get passed through the query:

- `geometry`
- `geometryType`
- `spatialRel`

Adding a unit test to illustrate the issue


tomwayson:
Actually, nvm, I'll do that after merging and fixing #589

tomwayson:
@gavinr - this was released in v2.1.1

PR #609, opened 2019-08-08
fix(portal): setting item access to public shares correctly

dmfenton:
Currently whe sharing a item w/ access public, the library requests `everyone=true&org=true`. Unfortunately, these appear to cancel each other out and the item is not shared at all. The ArcGIS Onlie Home app sets account=true and everyone=true when sharing publicly. This PR follows suit with the Home App.


dmfenton:
Thanks @tomwayson, how do we release this thang?

tomwayson:
I am cutting a release w/ the other recently merged PRs

Issue #611, opened 2019-08-15
support async and multipart upload for new item


haoliangyu:
The [addItem](https://developers.arcgis.com/rest/users-groups-and-items/add-item.htm) API is able to upload item data asynchronously. Implementing this feature in the library includes three parts.

## Update createItem() and createItemInFolder()

The `file`, `filename`, `multipart`, `dataUrl`, and `async` parameters are to trigger async upload and they need to be added to the function. There are two ways to do this:
* add to the `ICreateItemOptions` interface
* add to the `IItemAdd` interface

Since these parameters are not in the item metadata (once created), I think it makes more sense to add to the `ICreateItemOptions`.

## Add getItemStatus()

This is for the [status](https://developers.arcgis.com/rest/users-groups-and-items/status.htm) API to retrieve the processing status.

## Add addItemPart()

This is for the [addItemPart](https://developers.arcgis.com/rest/users-groups-and-items/add-item-part.htm) API to send file parts for a given item.

## Add getItemParts()

This is for the [getItemParts](https://developers.arcgis.com/rest/users-groups-and-items/parts.htm) API to get a list of uploaded file parts for a given item.

haoliangyu:
@tomwayson @mjuniper If you have any thought on this, please let me know.

PR #613, opened 2019-08-19
F/support async and multipart upload during item creation

haoliangyu:
address #611

### DIFF
--- a/packages/arcgis-rest-portal/src/items/create.ts
+++ b/packages/arcgis-rest-portal/src/items/create.ts
@@ -63,11 +63,10 @@ export function createFolder(
  * ```js
  * import { createItemInFolder } from "@esri/arcgis-rest-portal";
  * //
- * createItem({
+ * createItemInFolder({
  *   item: {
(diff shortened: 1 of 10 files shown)


packages/arcgis-rest-portal/src/items/create.ts
```diff
@@ -63,11 +63,10 @@ export function createFolder(
  * ```js
  * import { createItemInFolder } from "@esri/arcgis-rest-portal";
  * //
- * createItem({
+ * createItemInFolder({
```
tomwayson: nice catch

packages/arcgis-rest-portal/src/items/create.ts
```diff
@@ -91,6 +101,12 @@ export function createItemInFolder(
 
   // serialize the item into something Portal will accept
   requestOptions.params = {
+    file: requestOptions.file,
```
tomwayson: You should use `appendCustomParams()` for this.

haoliangyu:
@tomwayson I have updated the code.

PR #617, opened 2019-09-04
docs(queryFeatures): update IQueryFeaturesOptions docs to say we do n…

tomwayson:
…ot support pbf

AFFECTS PACKAGES:

ISSUES CLOSED: 612


jgravois:
is there any reason why we wouldn't want to allow folks to use this library to fetch .pbfs and something like [tippecanoe](https://github.com/mapbox/tippecanoe#tippecanoe-decode) to decode them?

tomwayson:
I'm not saying we _shouldn't_ support querying PBFs, I'm just saying we currently do **not** support it.

tomwayson:
I mean, I guess, to that end, I'm not sure what happens if you specify `"html"` right now.

jgravois:
my hope is that it'd be as simple as treating the response as a `blob` and passing it along. I'll try and find time in the next week or two to confirm.

tomwayson:
I believe in the case of pbf we would use [arrayBuffer()](https://developer.mozilla.org/en-US/docs/Web/API/Body/arrayBuffer).

I know we _can_ handle these kind of responses, and even do the right thing automagically (return an [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) for `"pbf"`, a string for `"html"`), but:
- no one has asked for these things
- I don't have time
- you _should_ *not* have time 😉 

So until someone PRs first class support, I just want the docs to reflect the current reality.

jgravois:
* I agree wholeheartedly that its not worth Esri development time to do anything 'automagic'.
* I'm also fine with you just doccing the current state of affairs and calling it a day.
* I also concede your point that I should have better things to do. 🤓 

That said, given the back door we put in earlier, this _should_ already just work, and to be honest, it's a pretty cool integration.

```js
queryFeatures({
  url,
  where,
  rawResponse: true
})
  .then(response => response.arrayBuffer())
  .then(pbf => { // have something like tippecanoe do the rest) })
```
I only say this because in lieu of 'first class support' I think a code snippet in the doc would be more useful to folks than just words of caution. ❤️

packages/arcgis-rest-feature-layer/src/query.ts
```diff
@@ -80,7 +80,10 @@ export interface IQueryFeaturesOptions extends ISharedQueryOptions {
   returnTrueCurves?: false;
   sqlFormat?: "none" | "standard" | "native";
   returnExceededLimitFeatures?: boolean;
-  f?: "json" | "geojson" | "html" | "pbf";
+  /**
+   * Response format. Protocol buffer (pbf) is not currently supported.
+   */
+  f?: "json" | "geojson" | "html";
```
pspraveenkr: Doesn't seem like there is need/use-case to support `html` format for clients. Do we need to support it?

COV-GIS:
It shouldn't be supported. `html` is only useful to AGS on a query page endpoint for testing or developing query params. No one ever has purposefully made a `f=html` call outside of the AGS query page context.

tomwayson:
Thanks for the feedback everyone. I'm happy w/ how it ended up:

image

Also, I doc'd the other default values we set (`where` and `outFields`) while I was in there.

PR #622, opened 2019-09-11
Add missing param to interface

MikeTschudi:
The `updateItem` function in `arcgis-rest-portal` has a single argument of type `IUpdateItemOptions`. This type does not include the optional property `params`; however, that property is accepted by the function
```js
export function updateItem(
  requestOptions: IUpdateItemOptions
): Promise<IUpdateItemResponse> {
  const owner = determineOwner(requestOptions);
  const url = `${getPortalUrl(requestOptions)}/content/users/${owner}/items/${
    requestOptions.item.id
  }/update`;

  // serialize the item into something Portal will accept
  requestOptions.params = {
    ...requestOptions.params,
    ...serializeItem(requestOptions.item)
  };

  return request(url, requestOptions);
}
```
and is used in the package's tests
```js
it("should update an item with custom params", done => {
    :           :           :
  updateItem({
    item: fakeItem,
    authentication: MOCK_USER_SESSION,
    params: {
      clearEmptyFields: true
    }
  })
    .then(response => {
      :           :           :
    });
});
```

The absence of this property in `IUpdateItemOptions` can cause a type error in callers that use the property.

### DIFF
--- a/packages/arcgis-rest-portal/src/items/update.ts
+++ b/packages/arcgis-rest-portal/src/items/update.ts
@@ -16,7 +16,15 @@ import {
 } from "./helpers";
 
 export interface IUpdateItemOptions extends ICreateUpdateItemOptions {
+  /**
+   * Portions of item to update; contents are serialized before being sent
+   */
   item: IItemUpdate;
+  /**
+   * Additional content to merge into parameters; in particular, useful for content that you don't
+   * want to serialize (e.g., blobs such as an item's thumbnail or metadata)
+   */
+  params?: any;
 }
 
 export interface IMoveItemOptions extends ICreateUpdateItemOptions {


jgravois:
`IUpdateItemOptions` extends
`ICreateUpdateItemOptions` which extends 
`IUserRequestOptions`  which extends
[`IRequestOptions`](https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/) 

so I'm not sure why you're encountering a type error when you try to use it. Maybe we've exceeded even TypeScript's intellisense ability when we ask it to traverse the inheritance of five different interfaces? 😬

MikeTschudi:
Intellisense works and local tests pass; it's Travis-ci that's complaining:
image

jgravois:
I've seen rogue errors like that before. maybe try running `npm run clean`?

MikeTschudi:
The error is from the [Travis-CI system](https://travis-ci.org/Esri/solution.js/builds/583877614?utm_source=github_status&utm_medium=notification); doesn't that always start with a clean space?

MikeTschudi:
Switched to a different computer and now see the errors regularly. Tried adding `params` property to object after creating it; Intellisense completed it; build rejected it.
Simplest path appears to be to change the options structure to `any`, alas.

PR #631, opened 2019-10-21
feat(arcgis-rest-portal): add searchUsers function

mjuniper:
Add searchUsers function. Hits /portals/self/users/search.

AFFECTS PACKAGES:
@esri/arcgis-rest-portal


packages/arcgis-rest-portal/src/users/search-users.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SearchQueryBuilder } from "../util/SearchQueryBuilder";
+import { ISearchOptions, ISearchResult } from "../util/search";
+import { genericSearch } from "../util/generic-search";
+
+// export interface IUserSearchOptions extends IAuthenticatedRequestOptions, IPagingParams {
+//   q: string | SearchQueryBuilder;
+//   sortField?: string;
+//   sortOrder?: string;
+//   [key: string]: any;
+// }
```
tomwayson: Did you mean to check this in?

packages/arcgis-rest-portal/src/util/generic-search.ts
```diff
@@ -35,9 +35,20 @@ export function genericSearch<T extends IItem | IGroup>(
...
+  let path = searchType === "item" ? "/search" : "/community/groups";
+  switch (searchType) {
+    case "item":
+      path = "/search";
+      break;
+    case "group":
+      path = "/community/groups";
+      break;
+    default:
+      // "users"
```
tomwayson: This is an implementation detail, so feel free to ignore, but sorta feels like to me that `item` should be the default type.

Issue #630, opened 2019-10-21
BulkGeocode - adding params causes error 


gavinr:
## Summary
When I add a `params` property to my `bulkGeocode` function call, I get an error.

## Expected
I should be able to add a `params` property to the passed object to pass arbitrary parameters to the REST call, per [the doc](https://esri.github.io/arcgis-rest-js/api/geocoding/bulkGeocode/#params).

## Actual
When I pass just `addresses` property to `bulkGeocode()`, [it works fine](https://codepen.io/gavinr/pen/vYYgLWQ):

```js
arcgisRest.bulkGeocode({ 
    endpoint: '...',
    addresses: [...]
  })
  .then((response) => {
    console.log('res', response);
  });
```

When I add a `params` property (below is what I'm using, but I think it happens with any contents of the params object):

```js
arcgisRest.bulkGeocode({ 
    endpoint: '...',
    addresses: [...],
    params: {
      outSR: '4326'
    }
  })
  .then((response) => {
    console.log('res', response);
  });
```

... I get this error: ([code showing issue](https://codepen.io/gavinr/pen/MWWJyNZ))

```
Uncaught TypeError: Cannot read property 'records' of undefined
    at bulk.ts:78
```

gavinr:
@jgravois I was comparing the batch code to the geocoder code - would it make sense to utilize the [appendCustomParams](https://github.com/Esri/arcgis-rest-js/commit/99300a8c1a73fa9c9a77a592b672e3cc53b4ff65#diff-3e2ebf3255ae6b0604225f67f678f5dcR78) module in bulkGeocoder as well ([here](https://github.com/Esri/arcgis-rest-js/blob/cba6e5647a99d7154780cf9bf5f99c2ac90924f4/packages/arcgis-rest-geocoding/src/bulk.ts#L67-L74))?

jgravois:
that would allow you to define an interface that exposed more properties specific to the operation as 'first class citizens'. if I could do it all over again I'd only do that for required parameters though.

I think [intimidating interfaces](https://esri.github.io/arcgis-rest-js/api/feature-layer/IQueryFeaturesOptions/) overwhelm the benefit of an 'uglier' signature. For everything optional, the `params` backdoor works just fine.

PR #633, opened 2019-10-22
Bulk Geocode - Properly pass thorugh `params`

gavinr:
#630


gavinr:
@jgravois would love your feedback on this code change if you have time.

packages/arcgis-rest-geocoding/src/bulk.ts
```diff
@@ -66,13 +66,18 @@ export function bulkGeocode(
...
-      forStorage: true,
-      addresses: { records: [] }
-    },
+    params: {},
     ...requestOptions
   };
 
+  if (!options.params.forStorage) {
+    options.params.forStorage = true;
+  }
```
jgravois: I know its against the TOU and I _think_ that failing to set `forStorage: true` will cause the call to fail. I'd recommend removing the if statement entirely.

other than that, this should be good to go. 😎

gavinr: @jgravois after looking into this, I found the documentation that supports your point:

> It is assumed that results from this operation are always stored for later use, so forStorage is an invalid parameter and shouldn't be added to a request.

image
https://developers.arcgis.com/rest/geocode/api-reference/geocoding-free-vs-paid.htm#ESRI_SECTION1_B9D659BA26584F26A65CBA24E59AFE6E

Thanks for the feedback!

jgravois:
I'm glad you didn't tell him to get clever and do:
```ts
options.params.addresses = {
  records: options.addresses.map(attributes => { attributes })
};
```
#forEach()4lyfe

tomwayson:
Actually @gavinr do you mind using `map` as @jgravois suggests above and then doing `npm run c` to generate a commit message that will make it easier for me to generate the changelog? then I will merge and deploy.

PR #638, opened 2019-11-14
feat(add applyEdits function): adds \`applyEdits()\` to feature-layer…

COV-GIS:
… for bulk crud transactions.

### DIFF
--- a/packages/arcgis-rest-feature-layer/src/helpers.ts
+++ b/packages/arcgis-rest-feature-layer/src/helpers.ts
@@ -50,9 +50,9 @@ export interface IApplyEditsAttachmentResult extends IEditFeatureResult {
  * Apply edits result Object.
  */
 export interface IApplyEditsResult {
-  addResults?: IEditFeatureResult[];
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-feature-layer/src/helpers.ts
```diff
@@ -39,6 +39,27 @@ export interface IEditFeatureResult {
...
+  globalId: string;
+}
+
+/**
+ * Apply edits result Object.
+ */
+export interface IApplyEditsResult {
+  addResults?: IEditFeatureResult[];
+  updateResults?: IEditFeatureResult[];
+  deleteResults?: IEditFeatureResult[];
```
sandromartis: Doesn't the REST API always return `addResults`, `updateResults` and `deleteResults`?
I don't think these have to be optional.

tomwayson:
This is in [v2.7.0]( https://github.com/Esri/arcgis-rest-js/releases/tag/v2.7.0)

PR #642, opened 2019-12-05
Add Portal getSubscriptionInfo fn

btpschroeder:
Added a Portal `getSubscriptionInfo` method -- essentially the same thing as `getPortal` or `getSelf` but with `/subscriptionInfo` appended to the URL. Needed to access certain properties only available from this endpoint.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5205,7 +5205,8 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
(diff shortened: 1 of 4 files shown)


tomwayson:
This is in [v2.7.0]( https://github.com/Esri/arcgis-rest-js/releases/tag/v2.7.0).

jgibson02:
@tomwayson I think there might be an issue with the TS declarations not including this method, here's a repro sandbox, do you see the same thing?
https://codesandbox.io/s/arcgis-rest-portal-getsubscriptioninfo-c8tik

Also I noticed on the latest docs it mentions the import is from `@esri/arcgis-rest-request` but it should be from `@esri/arcgis-rest-portal` right?
image

PR #648, opened 2020-01-24
Added new item relationship types

MikeTschudi:
Added new types listed in https://developers.arcgis.com/rest/users-groups-and-items/relationship-types.htm

Updated tslib to 1.10.0 (current) in order to get `npm install` to get past `src/utils/with-options.ts(46,21): error TS2343: This syntax requires an imported helper named '__spreadArrays' which does not exist in 'tslib'. Consider upgrading your version of 'tslib'.` error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -565,7 +565,7 @@
     "asn1.js": {
       "version": "4.10.1",
       "resolved": "https://registry.npmjs.org/asn1.js/-/asn1.js-4.10.1.tgz",
-      "integrity": "sha1-ucK/WAXx5kqt7tbfOiv6+1pz9aA=",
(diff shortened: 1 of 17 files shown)


tomwayson:
Thx @MikeTschudi

What context are you running `npm install` in when you see that message? The root of this monorepo? Your own app/library?

MikeTschudi:
@tomwayson, root

tomwayson:
I'm wary of bumping tslib (and in fact have been wary about even including it as a dependency in the first place since I'm uncertain of how much redundant code it's injecting into the built code). Can you PR _just_ the interface change and I'll merge/release that?

MikeTschudi:
Sounds good to me

Issue #580, opened 2019-06-06
UserSession.completeOAuth2 error in production React app


r-watkins:
I get the following error when attempting to complete an authentication using UserSession.completeOAuth2:

```
TypeError: o.opener.parent[("__ESRI_REST_AUTH_HANDLER_" + z)] is not a function
    at r (UserSession.js:181)
    at Function.e.completeOAuth2 (UserSession.js:222)
    at auth.js:41
    at c (runtime.js:45)
    at Generator._invoke (runtime.js:264)
    at Generator.e.<computed> [as next] (runtime.js:98)
    at t (asyncToGenerator.js:3)
    at z (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
```

The app is built off of React (CRA) and is using the latest arcgis-rest packages (2.0.3).
It works with no errors in local but fails with this error when deployed to the production site. I am passing a URL parameter through the redirect which I'm not sure is affecting things (popup is set to false).

Even though it fails here and halts the app I can copy/paste the redirect URL that it produces into a new tab and the authentication goes through with no errors.

I'm stumped.

tomwayson:
if you open the console of your built app does the `window` define a global variabl that starts w/ `__ESRI_REST_AUTH_HANDLER_`?

What version of react-scripts are you using?

tomwayson:
FYI - this CRA (react-scripts@3.0.0) app uses arcgis-rest-js v2.0.3 and the log in works in prod mode:

https://create-arcgis-app.surge.sh/

Here's the source:

https://github.com/tomwayson/create-arcgis-app/

r-watkins:
Thanks for the reply @tomwayson 

I don't see `__ESRI_REST_AUTH_HANDLER_` in the `window` object of either the production build that fails or the local build that works.

The project is currently on react-scripts 3.0.1.

In reference to the create-arcgis-app, it looks like it only needs to use the `UserSession.beginOAuth2` method because it has the popup enabled, is that right? I am trying to avoid the popup so I have to return from the redirect with `UserSession.completeOAuth2` where the error is occurring.

tomwayson:
@patrickarlt 

I met w/ @r-watkins and I'm pretty sure this is a bug.

[...]

PR #652, opened 2020-01-28
Fix completeOAuth2 for edge case where window is opened from another window.open call

patrickarlt:
In cases like the ArcGIS Online app switcher the windows of each respective app are opened with `window.open` which sets `window.opener` and `window.opener.parent` to ArcGIS Online. When used with `{popup: false}` this means that when you call `completeOAuth2` after auth happens `window.opener` is still set and the auth assumes you are in a popup workflow.

This PR adds some extra checks:

1. Check that `popup` is true in the options.
2. Check that the handler exists before trying to call it.

This MIGHT be considered a breaking change because up until now we didn't really do anything if you didn't pass `popup` to `completeOAuth2`. it all still worked because didn't really check popup, which is the error here so this could also be considered a bug. 

I think that by adding a default of `popup:true` this will prevent a breaking change. Users in the popup:true` workflow already has to pass popup:false` to both oAuth function to get this to work which is still the case.

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -348,7 +348,7 @@ export class UserSession implements IAuthenticationManager {
   /* istanbul ignore next */
   public static completeOAuth2(options: IOAuth2Options, win: any = window) {
     const { portal, clientId }: IOAuth2Options = {
-      ...{ portal: "https://www.arcgis.com/sharing/rest" },
+      ...{ portal: "https://www.arcgis.com/sharing/rest", popup: true },
       ...options
     };
 
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -353,7 +353,7 @@ export class UserSession implements IAuthenticationManager {
     };
 
     function completeSignIn(error: any, oauthInfo?: IFetchTokenResponse) {
-      if (win.opener && win.opener.parent && win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`] && popup) {
+      if (popup && win.opener && win.opener.parent && win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`]) {
         const handlerFn = win.opener[`__ESRI_REST_AUTH_HANDLER_${clientId}`];
         if (handlerFn) {
           handlerFn(
@@ -365,7 +365,7 @@ export class UserSession implements IAuthenticationManager {
         return undefined;
       }
 
-      if (win !== win.parent && win.parent && win.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`] && popup) {
+      if (popup && win !== win.parent && win.parent && win.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`]) {
         const handlerFn = win.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`];
         if (handlerFn) {
           handlerFn(
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -353,7 +353,7 @@ export class UserSession implements IAuthenticationManager {
     };
 
     function completeSignIn(error: any, oauthInfo?: IFetchTokenResponse) {
-      if (win.opener && win.opener.parent) {
+      if (win.opener && win.opener.parent && win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`] && popup) {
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -365,7 +365,7 @@ export class UserSession implements IAuthenticationManager {
         return undefined;
       }
 
-      if (win !== win.parent) {
+      if (popup && win !== win.parent && win.parent && win.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`]) {
```
patrickarlt: I think this might need to be wrapped in a `try` because trying to access `win.opener.parent` results in an error.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -348,12 +348,12 @@ export class UserSession implements IAuthenticationManager {
   /* istanbul ignore next */
   public static completeOAuth2(options: IOAuth2Options, win: any = window) {
     const { portal, clientId }: IOAuth2Options = {
```
noahmulfinger: I think `popup` needs to be declared here.

patrickarlt:
@noahmulfinger @dbouwman  I updated the doc and tests a little bit.

packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -1011,6 +1011,36 @@ describe("UserSession", () => {
...
+
+      const MockWindow = {
+        location: {
+          href:
+            "https://example-app.com/redirect-uri#error=Invalid_Signin&error_description=Invalid_Signin"
+        },
+        get parent() {
+          return MockParent;
+        }
+      };
```
noahmulfinger: I think this mock window is backwards. We reference `window.opener.parent` in the function, but this mock window creates `window.parent.opener`.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -339,42 +339,59 @@ export class UserSession implements IAuthenticationManager {
...
+            );
+          }
+          win.close();
+          return undefined;
         }
-        win.close();
-        return undefined;
+      } catch (e) {
+        throw new ArcGISAuthError(
+          `Unable to complete authentication. You specified popup based oAuth2 but no handler from "beginOAuth2()" present. This generally happens because the "popup" option differs between "beginOAuth2()" and "completeOAuth2()".`
```
noahmulfinger: Very minor, but we might want to change the wording to something like `It's possible you specified...` because this try catch happens outside of the `if` statements, so its possible that some different error occurred to trigger the catch.

tomwayson:
This seems like a very similar issue to #580 

My proposed fix for that was to just put a guard around the existence of the handler fn before trying to invoke it, see: https://github.com/Esri/arcgis-rest-js/issues/580#issuecomment-500618952

I haven't wrapped my head around the code and discussion in PR yet, but I want to make sure that whatever we end up doing here will also fix that issue.

patrickarlt:
@tomwayson this is pretty much #580 but with 2 differences:

1. The auth isn't popup based so the window (and `window.opener`) property stay the same during the entirety of the oAuth process. The fixes from #580 should work but...
2. The `window.opener` is on a different domain so trying to access `window.opener.parent` throws a cross domain error.

This PR also adds a check to make sure that `popup` is true before checking for handlers at all.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -339,42 +339,60 @@ export class UserSession implements IAuthenticationManager {
...
-          );
+      try {
+        if (
+          popup &&
+          win.opener &&
+          win.opener.parent &&
+          win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`]
+        ) {
+          const handlerFn =
+            win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`];
```
patrickarlt: This used to be 

```
win.opener[`__ESRI_REST_AUTH_HANDLER_${clientId}\`]
```

Which I think is incorrect since we check for `win.opener.parent`. This is also how it works in the Esri Leaflet sample and all test pass so I think this is ok.

patrickarlt:
@dbouwman @noahmulfinger @tomwayson I think this ready now. If you want to give it another look. Especially https://github.com/Esri/arcgis-rest-js/pull/652/files#r372560436.

packages/arcgis-rest-auth/src/UserSession.ts
tomwayson: I believe you, but I don't think tests passing means much since I assume the called APIs are mocked. I will pull this and run the OAuth demo. If that works, I think we're GTG.

Issue #654, opened 2020-02-05
type: enum that can represent LineSymbolStyles


kumarGayu:
can we have this types. so that it is easy for readability and user can use enum to define the type.

```
export const enum LineSymbolStyles {
  dash = "esriSLSDash",
  dashDot = "esriSLSDashDot",
  dashdotdot = "esriSLSDashDotDot",
  dot = "esriSLSDot",
  solid = "esriSLSSolid"
}
```

maneetgoyal:
Adding some quick details:

1. Currently, `arcgis-rest-types` exports:
```ts
export type SimpleLineSymbolStyle =
  | "esriSLSDash"
  | "esriSLSDashDot"
  | "esriSLSDashDotDot"
  | "esriSLSDot"
  | "esriSLSNull"
  | "esriSLSSolid";

export interface ISimpleLineSymbol extends ISymbol {
  type: "esriSLS";
  style?: SimpleLineSymbolStyle;
  color?: Color;
  width?: number;
}

// etc.
```

2. As Kumar pointed out, if the following is done instead, it is serving more use cases.

```ts
export const enum SimpleLineSymbolStyle {
  Dash = "esriSLSDash",
  DashDot = "esriSLSDashDot",
  DashDotDot = "esriSLSDashDotDot",
  Dot = "esriSLSDot",
  Solid = "esriSLSSolid",
  Null = "esriSLSNull"
}

export interface ISimpleLineSymbol extends ISymbol {
  type: "esriSLS";
  style?: SimpleLineSymbolStyle;
  color?: Color;
  width?: number;
}

// etc.
```

3. As enums can be used as types too, the older usage isn't changing much. Further, we have cases wherein a particular `SimpleLineSymbolStyle` value needs to be assigned to a variable. With enums, we wouldn't need to maintain our own list of strings and be readily in sync with `arcgis-rest-types`.

jgravois:
sounds reasonable enough to me.

please open a pull request with the proposed change and we can continue the discussion there.

maneetgoyal:
@subgan82 @pemberdom

PR #656, opened 2020-02-07
Replace string union types with const enums

maneetgoyal:
- Fixes https://github.com/Esri/arcgis-rest-js/issues/654
- `enum` names can be changed. Currently went with what seemed more intuitive.
- ~Some breaking changes are possible. For instance:~ See https://github.com/Esri/arcgis-rest-js/pull/656#issuecomment-591798396.

Earlier, we had:
```ts
export interface ISimpleFillSymbol extends ISymbol {
  type: "esriSFS";
  style?: SimpleFillSymbolStyle;
  color?: Color;
  outline?: ISimpleLineSymbol; // if outline has been specified
}

// The following is OK for TSC
const sample: ISimpleFillSymbol = {
  type: "esriSFS"
}; 
```

But now we have:
 ```ts
export const enum SymbolType {
  SLS = "esriSLS",
  SMS = "esriSMS",
  SFS = "esriSFS",
  PMS = "esriPMS",
  PFS = "esriPFS",
  TS = "esriTS"
}

export interface ISimpleFillSymbol extends ISymbol {
  type: SymbolType.SFS;
  style?: SimpleFillSymbolStyle;
  color?: Color;
  outline?: ISimpleLineSymbol; // if outline has been specified
}

// The following is WRONG for TSC
const sample: ISimpleFillSymbol = {
  type: "esriSFS"
};

// The following is OK for TSC
const sample2: ISimpleFillSymbol = {
  type: SymbolType.SFS
};
```

### DIFF
--- a/packages/arcgis-rest-types/src/statisticDefinition.ts
+++ b/packages/arcgis-rest-types/src/statisticDefinition.ts
@@ -2,7 +2,7 @@
  * Apache-2.0 */
 
 /**
- * Enum of all different statisticsType
+ * Enum of all different statistics operations
  */
 export const enum StatisticType {
   Count = "count",
@@ -16,6 +16,14 @@ export const enum StatisticType {
   DiscretePercentile = "percentile_disc"
 }
 
+/**
+ * Enum of sorting orders
+ */
+export const enum SortingOrder {
+  Ascending = "asc",
+  Descending = "desc"
+}
+
 export interface IStatisticDefinition {
   /**
    * Statistical operation to perform (count, sum, min, max, avg, stddev, var, percentile_cont, percentile_disc).
@@ -26,7 +34,7 @@ export interface IStatisticDefinition {
    */
   statisticParameter?: {
     value: number;
-    orderBy?: "asc" | "desc";
+    orderBy?: SortingOrder;
   };
   /**
    * Field on which to perform the statistical operation.
--- a/packages/arcgis-rest-types/src/statisticDefinition.ts
+++ b/packages/arcgis-rest-types/src/statisticDefinition.ts
@@ -28,13 +28,23 @@ export interface IStatisticDefinition {
   /**
    * Statistical operation to perform (count, sum, min, max, avg, stddev, var, percentile_cont, percentile_disc).
    */
-  statisticType: StatisticType;
+  statisticType:
+    | "count"
+    | "sum"
+    | "min"
+    | "max"
+    | "avg"
+    | "stddev"
+    | "var"
+    | "percentile_cont"
+    | "percentile_disc"
+    | StatisticType;
   /**
    * Parameter to be used along with statisticType. Currently, only applicable for percentile_cont (continuous percentile) and percentile_disc (discrete percentile).
    */
   statisticParameter?: {
     value: number;
-    orderBy?: SortingOrder;
+    orderBy?: "asc" | "desc" | SortingOrder;
   };
   /**
    * Field on which to perform the statistical operation.
--- a/packages/arcgis-rest-types/src/statisticDefinition.ts
+++ b/packages/arcgis-rest-types/src/statisticDefinition.ts
@@ -1,20 +1,26 @@
 /* Copyright (c) 2017-2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
+/**
+ * Enum of all different statisticsType
+ */
+export const enum StatisticType {
+  Count = "count",
+  Sum = "sum",
+  Minimum = "min",
+  Maximum = "max",
+  Average = "avg",
+  StandardDeviation = "stddev",
+  Variance = "var",
+  ContinuousPercentile = "percentile_cont",
+  DiscretePercentile = "percentile_disc"
+}
+
 export interface IStatisticDefinition {
   /**
    * Statistical operation to perform (count, sum, min, max, avg, stddev, var, percentile_cont, percentile_disc).
    */
-  statisticType:
-    | "count"
-    | "sum"
-    | "min"
-    | "max"
-    | "avg"
-    | "stddev"
-    | "var"
-    | "percentile_cont"
-    | "percentile_disc";
+  statisticType: StatisticType;
   /**
    * Parameter to be used along with statisticType. Currently, only applicable for percentile_cont (continuous percentile) and percentile_disc (discrete percentile).
    */
--- a/packages/arcgis-rest-types/src/symbol.ts
+++ b/packages/arcgis-rest-types/src/symbol.ts
@@ -225,7 +225,7 @@ export const enum HorizontalAlignment {
  *
  */
 export interface ITextSymbol extends IMarkerSymbol {
-  type: SymbolType.TS;
+  type: "esriTS" | SymbolType.TS;
   color?: Color;
   backgroundColor?: Color;
   borderLineSize?: number; // <size>;
--- a/packages/arcgis-rest-types/src/symbol.ts
+++ b/packages/arcgis-rest-types/src/symbol.ts
@@ -185,6 +185,23 @@ export interface ISimpleMarkerSymbol extends IMarkerSymbol {
 /**
  *
  */
+export const enum VerticalAlignment {
+  Baseline = "baseline",
+  Top = "top",
+  Middle = "middle",
+  Bottom = "bottom"
+}
+
+export const enum HorizontalAlignment {
+  Left = "left",
+  Right = "right",
+  Center = "center",
+  Justify = "justify"
+}
+
+/**
+ *
+ */
 export interface ITextSymbol extends IMarkerSymbol {
   type: SymbolType.TS;
   color?: Color;
@@ -193,8 +210,8 @@ export interface ITextSymbol extends IMarkerSymbol {
   borderLineColor?: Color;
   haloSize?: number; // <size>;
   haloColor?: Color;
-  verticalAlignment?: "baseline" | "top" | "middle" | "bottom";
-  horizontalAlignment?: "left" | "right" | "center" | "justify";
+  verticalAlignment?: VerticalAlignment;
+  horizontalAlignment?: HorizontalAlignment;
   rightToLeft?: boolean;
   kerning?: boolean;
   font?: IFont;
--- a/packages/arcgis-rest-types/src/symbol.ts
+++ b/packages/arcgis-rest-types/src/symbol.ts
@@ -4,18 +4,6 @@
 /**
  *
  */
-export const enum SymbolType {
-  SLS = "esriSLS",
-  SMS = "esriSMS",
-  SFS = "esriSFS",
-  PMS = "esriPMS",
-  PFS = "esriPFS",
-  TS = "esriTS"
-}
-
-/**
- *
- */
 export type Color = [number, number, number, number];
 
 /**
@@ -69,6 +57,18 @@ export interface IPictureSourced {
   angle?: number;
   xoffset?: number;
   yoffset?: number;
+}
+
+/**
+ *
+ */
+export const enum SymbolType {
+  SLS = "esriSLS",
+  SMS = "esriSMS",
+  SFS = "esriSFS",
+  PMS = "esriPMS",
+  PFS = "esriPFS",
+  TS = "esriTS"
 }
 
 /**
--- a/packages/arcgis-rest-types/src/symbol.ts
+++ b/packages/arcgis-rest-types/src/symbol.ts
@@ -4,7 +4,47 @@
 /**
  *
  */
+export const enum SymbolType {
+  SLS = "esriSLS",
+  SMS = "esriSMS",
+  SFS = "esriSFS",
+  PMS = "esriPMS",
+  PFS = "esriPFS",
+  TS = "esriTS"
+}
+
+/**
+ *
+ */
 export type Color = [number, number, number, number];
+
+/**
+ *
+ */
+export const enum FontStyle {
+  Italic = "italic",
+  Normal = "normal",
+  Oblique = "oblique"
+}
+
+/**
+ *
+ */
+export const enum FontWeight {
+  Bold = "bold",
+  Bolder = "bolder",
+  Lighter = "lighter",
+  Normal = "normal"
+}
+
+/**
+ *
+ */
+export const enum FontDecoration {
+  LineThrough = "line-through",
+  Underline = "underline",
+  None = "none"
+}
 
 /**
  *
@@ -12,9 +52,9 @@ export type Color = [number, number, number, number];
 export interface IFont {
   family?: string; // "<fontFamily>";
   size?: number; // <fontSize>;
-  style?: "italic" | "normal" | "oblique";
-  weight?: "bold" | "bolder" | "lighter" | "normal";
-  decoration?: "line-through" | "underline" | "none";
+  style?: FontStyle;
+  weight?: FontWeight;
+  decoration?: FontDecoration;
 }
 
 /**
@@ -60,7 +100,7 @@ export interface IMarkerSymbol extends ISymbol {
  *
  */
 export interface IPictureFillSymbol extends ISymbol, IPictureSourced {
-  type: "esriPFS";
+  type: SymbolType.PFS;
   outline?: ISimpleLineSymbol; // if outline has been specified
   xscale?: number;
   yscale?: number;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by maneetgoyal** [comment]
@jgravois

**Comment by maneetgoyal** [comment]
Hi @tomwayson, @jgravois, realized that `const string` enums were nowhere used in the package. Were there any issues seen with that approach?

**Comment by maneetgoyal** [comment]
Hi @tomwayson, @jgravois, can you please review this PR? We are just looking for a better string management solution. Enum is not the only way, so would be good to discuss other alternates too.

**Comment by jgravois** [comment]
this one is above my paygrade (doesn't take much at $0/hour).

@patrickarlt what do you think about this change?

**Comment by jf990** [comment]
I'm positive on this change as it utilizes the TypeScript language better to do static type checking on these enumerations, but this is a breaking change. Any existing code using the current string definitions will break with this. The fix is simple, a 1-for-1 change of the string to the enum, but code will break nonetheless.

**Comment by jgravois** [comment]
is there a way to ship enumerations alongside what we already have?

having something developers would opt into at their leisure would definitely be an easier sell than shipping a `v3.0.0`.

**Comment by maneetgoyal** [comment]
> is there a way to ship enumerations alongside what we already have?

@jgravois Thanks, that's a great thought.

1. Earlier, we had:
```ts
export interface ISimpleFillSymbol extends ISymbol {
  type: "esriSFS";
  style?: SimpleFillSymbolStyle;
  color?: Color;
  outline?: ISimpleLineSymbol;
}

```

2. Instead of the current solution, if we do the following, then we can have `const string` enums **without** breaking changes:

```ts
export const enum SymbolType {
 SLS = "esriSLS",
 SMS = "esriSMS",
 SFS = "esriSFS",
 PMS = "esriPMS",
 PFS = "esriPFS",
 TS = "esriTS"
}

export interface ISimpleFillSymbol extends ISymbol {
 type: "esriSLS" | SymbolType.SFS;
 style?: SimpleFillSymbolStyle;
 color?: Color;
 outline?: ISimpleLineSymbol;
}
```

3. Once it is done, the following is achieved:

```ts
// The following is OK for TSC; no breaking change
const sample: ISimpleFillSymbol = {
 type: "esriSFS"
};

// The following is also OK for TSC; new pattern
const sample2: ISimpleFillSymbol = {
 type: SymbolType.SFS
};
```

4. Basically, instead of doing `type: "esriSLS"`, the new solution does `type: "esriSLS" | SymbolType.SFS;`.

5. The same effect can be achieved by just creating `const string` enums and not doing `type: "esriSLS" | SymbolType.SFS;` also, but then the intelisense won't show that `SymbolType` enum can also be used instead of plain `"esriSLS"`, and new developers won't able to readily start using the new pattern.

maneetgoyal:
[Here's](https://github.com/Esri/arcgis-rest-js/compare/master...maneetgoyal:string-union-plus-enum) the new change. Can push those changes to this PR if it looks more backward compatible.

jgravois:
> Can push those changes to this PR if it looks more backward compatible.

please do. the additional union types definitely look like something we could ship without the rigamarole of publicizing a breaking change.

tomwayson:
Thank you @maneetgoyal.

Our apps don't use the types, so I only have a partial understanding of the benefits.

I want to enable a better TS experience, but I want to make sure we do that via a comprehensive plan, _especially_ if we're considering breaking changes. 

By that I mean that our types should be consistent across all the rest-js packages and also probably consistent w/ the types of the JSAPI itself.

Furthermore, since I don't use the types (rest-js nor JSAPI), I cannot be the person that comes up w/ that plan, nor even evaluate whether or not this or future PRs are consistent w/ that plan.

COV-GIS:
I have no opinion on this conceptually. But an easy fix for the readability/editor info is to add proper commenting to the types.

```javascript
/**
 * Symbol types. See the [REST Documentation](https://developers.arcgis.com/documentation/common-data-types/symbol-objects.htm) for more information.
 * 
 * `esriSLS` Simple Line Symbol
 * 
 * `esriSMS` Simple Marker Symbol
 * 
 * etc...
 */
export type SymbolType =
  | "esriSLS"
  | "esriSMS"
  | "esriSFS"
  | "esriPMS"
  | "esriPFS"
  | "esriTS";
```

You don't get the ability to arrow thru to values or other `enum` benefits but it's better than just seeing a list of strings that aren't necessarily clear or descriptive. And it would improve the docs too.

maneetgoyal:
- @jgravois The new changes are pushed.
- @tomwayson The change is non-breaking after the new commit. But yes, the consistency concern still remains. Will get back after discussing with our team.

kumarGayu:
I feel `export enum` has advantage over the `export const enum` So i feel having enum is better than union type as union types does not show available values in js and union does not have that advantage `enum` over `const enum` has iterating capability. I feel it is better to have enums.

maneetgoyal:
Hi @dbouwman, we are currently testing another solution using `as const`. It'll be great to wait just a little till we finish our testing. Might prevent drawing a second minor release.

maneetgoyal:
- @COV-GIS Adding descriptive comments is certainly useful, we'll try to address it incrementally, in another upcoming PRs. (cc @subgan82). 
- However, comments alone doesn't meet all our requirements. So currently evaluating the `enum` v/s `const enum` v/s `as const` (object literal) approach. 
- Will update with the findings.

subgan82:
We are testing this approach an alternative to enum and can also support the union string enums
`as const` works in newer TS versions introduced in 3.4 

```javascript
const FontStyle = {
 Italic: "italic",
 Normal: "normal",
 Oblique: "oblique"
} as const;
export type FontStyle = typeof FontStyle[keyof typeof FontStyle];


export interface IFont {
 family?: string; // "<fontFamily>";
 size?: number; // <fontSize>;
 style?: FontStyle;
 weight?: "bold" | "bolder" | "lighter" | "normal" | FontWeight;
 decoration?: "line-through" | "underline" | "none" | FontDecoration;
}

const a: IFont = {
   style: FontStyle.Normal
 }  
 const a: IFont = {
   style: "normal"
 } 

```

COV-GIS:
No worries @maneetgoyal. I realize you're goals are more than extra info in the editor and this lib's docs. I had never looked at the types before this PR. I was actually surprised to see empty comments when I did. The rest of the lib is quite robust in that regard. I wouldn't worry about describing the key/value pairs so much. A simple description and markdown link to the appropriate rest api/spec docs page is good enough. That link to the authoritative spec is the best thing to treat all your ills. I click those all the time.

maneetgoyal:
So, upon comparing:

`as const`:
- Unlike `const enum`, the `as const` approach leads to an extra object in the compiled JS, so bigger bundle sizes. How large the impact is depends on how big and how many objects are used.
- The intelisense shows the possible string values, and doesn't indicate that an object can be used instead. So it will be hard to promote usage like `FontStyle.Italic` over the conventional `"italic"`.
Screen Shot 2020-02-27 at 4 28 37 PM

- `Object.keys(FontStyle)` will produce an array of all possible string values, which is a desired functionality for our use cases.

`const enum`:
- The compiled JS, doesn't have any trace of the enum declaration, so smaller bundles. Wherever the enum property is used as a value, the corresponding is replaced inline in compiled JS. 
- `const enum` lacks the support to iterate over all of its possible values.

`enum`:
- In some ways it is similar to `as const`, as it allows producing an array of all possible values of the object and also leads to larger bundles.
- However, [an article](https://medium.com/@maxheiber/alternatives-to-typescript-enums-50e4c16600b1) suggests that the compilation target can impair debuggability. (I personally have used mostly `const enums` only and not `enums`, so don't have any experience to support this claim.)

---

It is boiling down to how badly we need the capability to produce an array of all possible string values. If it is needed, `as const` is looking more plausible. @subgan82 @kumarGayu

tomwayson:
As someone who doesn't use the types, I don't want anything additional added to the bundle.

Does this merged PR add to the bundle?

maneetgoyal:
Out of the 3 approaches discussed above, only the merged PR one doesn't add any extra code to the consumer's bundles. Here are the `TS ==> JS` behaviors for all these 3 approaches:

1. `as const`:
```ts
// TS File
const FontStyle = {
 Italic: "italic",
 Normal: "normal",
 Oblique: "oblique"
} as const;
export type FontStyle = typeof FontStyle[keyof typeof FontStyle];
```

```js
// Compiled JS
const FontStyle = {
    Italic: "italic",
    Normal: "normal",
    Oblique: "oblique"
};
```

---

2. `const enum`

```ts
// TS File
const enum FontStyle {
 Italic = "italic",
 Normal = "normal",
 Oblique = "oblique"
}
```

```js
// Compiled JS
// No extra code
```

---

3. `enum`

```ts
// TS File
enum FontStyle {
 Italic = "italic",
 Normal = "normal",
 Oblique = "oblique"
}
```

```js
// Compiled JS
var FontStyle;
(function (FontStyle) {
    FontStyle["Italic"] = "italic";
    FontStyle["Normal"] = "normal";
    FontStyle["Oblique"] = "oblique";
})(FontStyle || (FontStyle = {}));
```

---

- Important to note that with `const enums` we are losing the capability to iterate over possible values (strings) of the concerned type.
- But at least, `const enum` still seems to cater to more use cases than using `string union` type.

subgan82:
> It is boiling down to how badly we need the capability to produce an array of all possible string values. If it is needed, as const is looking more plausible.

Ok , another benefit was that we dont have both string union types and enums listed in the types as we have with merged PR when using the Object Literals.

With Object Literal `as const` approach

```javascript

const FontStyle = {
  Italic: "italic",
  Normal: "normal",
  Oblique: "oblique"
} as const;

export type FontStyle = (typeof FontStyle)[keyof typeof FontStyle];

export interface IFont {
  family?: string; // "<fontFamily>";
  size?: number; // <fontSize>;
  style?: FontStyle;
  weight?: FontWeight;
  decoration?: FontDecoration;
}
```

Merged PR using `const enum`

```javascript
export interface IFont {
  family?: string; // "<fontFamily>";
  size?: number; // <fontSize>;
  style?: "italic" | "normal" | "oblique" | FontStyle;
  weight?: "bold" | "bolder" | "lighter" | "normal" | FontWeight;
  decoration?: "line-through" | "underline" | "none" | FontDecoration;
}
```

Can we check the doc on this impact ? It preserves the backwards compatability and provides type for uses who would like to use `FontStyle.Normal`.


On the question of if we choose `enum` vs `const enum` const enum will not add extra lookup table and it is better for smaller bundle sizes.

On using `const enum vs Object Literal`, the bundle size will increase with `Object as Const` approach as we  are creating an Object. `Const Enum` gets resolved at compile time itself.

Overall the merged PR does not have a breaking change , the issue would be we have both string union types and enum, this may confuse the user. Other than it is good solution.

Since JS does not have native enum support, there is a [stage 0 proposal](https://github.com/rbuckton/proposal-enum)l which may deviate from TS enums.

So there is group of users who suggest to avoid Enums and use alternatives like `Object Literal as const` approaches. The Pro Enum group advocate string based enums are good.

I believe it boils down to on the API if we want to use enums or alternatives to enums to provide Types instead of string union types.

If we want to decide purely on bundle size, string based [const enums ](https://2ality.com/2020/01/typescript-enums.html#const-enums) are good though we miss capability to iterate over possible values. 

If we want the iteration over possible values `Object based approach as const` is better alternative than [producing lookup table](https://netbasal.com/a-smaller-bundle-with-const-enums-in-typescript-b7f786227e5e) with using `enum`.

Comparisons of various [approaches](https://2ality.com/2020/02/enum-alternatives-typescript.html#summary-of-enums-and-enum-alternatives) 

cc: @dasa

tomwayson:
I appreciate the research.

> only the merged PR one doesn't add any extra code to the consumer's bundles

That's good.

> If we want to decide purely on bundle size

I'm not opposed to adding to the bundle _as long as it provides run-time value_ to non-TS users, and is not just an artifact of build-time static type checking. So when you say:

> we loose iterating over the values

You are talking about at run-time, right? Sounds like you have a need to be able to iterate over the values _at runtime_ and produce something a drop down list. I can see how that would be valuable  to others and am not opposed to adding something like that. I assume that could also be introduced in a non-breaking way, right?

However, it's not like any of these ways add run-time type checking in JS, right? I'm not even sure I'd want that, I'm just trying to fully understand the run-time benefits of adding to the bundle.

> the issue would be that we will have both string union types and enum, this might confuse the user. 

If we end up introducing any of these patterns alongside string unions we could decide to deprecate the string unions and remove them at the next major release. So I wouldn't worry about the confusion long term.

I'd also like to hear from @dasa to see what the JS API's longer term plan is. I'd like to be consistent w/ that if possible.

dasa:
We do use const enum, but we can't use `as const` yet since we're still on TS 3.3, but that does seem to be the preferred way going forward unless you need to performance of const enum and don't need the runtime inspection.

subgan82:
> If we end up introducing any of these patterns alongside string unions we could decide to deprecate the string unions and remove them at the next major release. So I wouldn't worry about the confusion long term.

That is a good plan , for newer types we can go with enums or enum alternatives based on what we decide here instead of union string types.

subgan82:
Here TS version is 3.0 
https://github.com/Esri/arcgis-rest-js/blob/master/package.json#L60 
While running the `build` it worked without any issues when using `as const`.

subgan82:
>We do use const enum, but we can't use as const yet since we're still on TS 3.3, but that does seem to be the preferred way going forward unless you need to performance of const enum and don't need the runtime inspection.

@dasa once JSAPI updates TS version , Will it use `as const` most likely  over `const enum` ?

dasa:
const assertions were added at 3.4: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

It looks like rest-js is using TS 3.5.3: https://github.com/Esri/arcgis-rest-js/blob/master/package-lock.json#L13819

> once JSAPI updates TS version , Will it use `as const` most likely over `const enum` ?

I'm not sure. I think the 3D code likes the performance benefit of const enum.

subgan82:
> I'm not sure. I think the 3D code likes the performance benefit of const enum.

If we go with this current merged PR using `const enum` it sound like it will be fine, and then when JSAPI chooses a different approach we can sync up also.

Here is the [branch](https://github.com/Esri/arcgis-rest-js/compare/ganesh/enum-alt-as-const?expand=1) that uses `as const` approach

subgan82:
> I'm not opposed to adding to the bundle as long as it provides run-time value to non-TS users, and is not just an artifact of build-time static type checking. So when you say:
You are talking about at run-time, right? 

Yes using `const enum` it gets resolve at compile time, At runtime those are just the values.

If we want to provide run-time value to non-TS (JS) users then `as const` using `Object Literal` approach would be the alternative.

maneetgoyal:
> Can we check the impact of doc in both cases?

- `const enum`:

Screen Shot 2020-02-28 at 12 34 18 PM

Upon clicking `StatisticType`:

Screen Shot 2020-02-28 at 12 33 46 PM

---

- `as const`:

I had to use `export const StatisticType = {...` instead of `const StatisticType = {` in `src/statisticDefinition.ts`. After that `StatisticsType` link became clickable:

Screen Shot 2020-02-28 at 12 45 26 PM

Upon clicking `StatisticType`:

Screen Shot 2020-02-28 at 12 47 51 PM

subgan82:
> I had to use export const StatisticType = {... instead of const StatisticType = { in src/statisticDefinition.ts. After that StatisticsType link became clickable:

When i was running the typedoc was using `3.2.4` where `as const` was not supported , updating the TS version may help 

```
Using TypeScript 3.2.4 from /Users/gane5688/gitrepos/arcgis-rest-js/node_modules/typedoc/node_modules/typescript/lib
Error: /Users/gane5688/gitrepos/arcgis-rest-js/packages/arcgis-rest-types/src/statisticDefinition.ts(16)
```

subgan82:
After updating the TypeDoc version and also fixing TS version to 3.5.2 as used in [package-lock.json ](https://github.com/Esri/arcgis-rest-js/pull/656#issuecomment-592626789)

Running Build and Docs on the Object as Const Enum Alternative [branch](https://github.com/Esri/arcgis-rest-js/compare/ganesh/enum-alt-as-const), without any changes to that branch the docs for StatisicType 


It has the StatisticType Link and which looks as follows.

Screen Shot 2020-02-28 at 4 27 49 PM

Screen Shot 2020-02-28 at 4 27 58 PM

maneetgoyal:
Now since docs are also responding well, `as const` is looking quite a plausible option. I tried importing `StatisiticsType` in a **different file** to use it as a value. It threw the following error:

Screen Shot 2020-02-28 at 4 49 21 PM

Again, upon using `export const StatisticType = {...` instead of `const StatisticType = {`, it got fixed:

Screen Shot 2020-02-28 at 4 50 54 PM

PR #661, opened 2020-02-24
fix(getItemResources): do not mutate requestOptions in getItemResources

dbouwman:
Perviously we mutated the requestOptions, but this causes issues if the same requestOptions
instance is used in multiple calls - as in the case of long-promise chains orchestrating complex
interactions with the platform.

AFFECTS PACKAGES:
@esri/arcgis-rest-portal

### DIFF
--- a/packages/arcgis-rest-portal/src/items/get.ts
+++ b/packages/arcgis-rest-portal/src/items/get.ts
@@ -148,13 +148,16 @@ export function getItemResources(
 ): Promise<any> {
   const url = `${getPortalUrl(requestOptions)}/content/items/${id}/resources`;
 
-  // mix in user supplied params
-  requestOptions.params = {
-    ...requestOptions.params,
-    num: 1000
+  // Mix in num:1000 with any user supplied params
+  // Key thing - we don't want to mutate the passed in requestOptions
+  // as that may be used in other (subsequent) calls in the course
+  // of a long promise chains
+  const options: IRequestOptions = {
+    ...requestOptions
   };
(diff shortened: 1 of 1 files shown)


packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -148,13 +148,16 @@ export function getItemResources(
...
-    ...requestOptions.params,
-    num: 1000
+  // Mix in num:1000 with any user supplied params
+  // Key thing - we don't want to mutate the passed in requestOptions
+  // as that may be used in other (subsequent) calls in the course
+  // of a long promise chains
+  const options: IRequestOptions = {
+    ...requestOptions
   };
+  options.params = { ...options.params, ...{ num: 1000 } };
```
jgravois: I think you could skip one destructuring assignment and allow folks to set their own `num` if you did something like this instead.
```suggestion
  options.params = { num: 1000, ...options.params };
```

dbouwman: good call... I was just bustin' out the 🔨 having spent an hour+ pulling my hair out tracking down what was janking the requestOptions in the middle of a long promise chain... causing fetchMock to not match a subsequent search api call... but - commit inbound

jgravois: the inverted order preceded your hammering.

nice sleuthing! 🚧🕵👷🚧

dbouwman: updated & thanks! hope all is well @jgravois :)

Issue #658, opened 2020-02-13
support group item search?


asizer:
I'm looking for a function that will search a group's contents, and couldn't find one here. Specifically, I need to hit this endpoint:
```
https://www.arcgis.com/sharing/rest/content/groups/<groupid>/search
```

I see I can get the entirety of the group's contents using `getGroupContent`, but I would actually like to _search_ the contents, with a `q` query or a `categories` param or both.

I know what you're thinking. Why not just do a `searchItems` with a groupid? Well, it turns out these two queries have different results:

```
https://www.arcgis.com/sharing/rest/content/groups/<groupid>/search?q=<my query here>
```
```
https://www.arcgis.com/sharing/rest/search?q=group:<groupid> <my query here> 
```

Here's an example. The first url, `groups/<groupid>/search` reports 460 total items, while the second, just `/search`, only has 29. 

https://www.arcgis.com/sharing/rest/content/groups/47dd57c9a59d458c86d3d6b978560088/search?categories=["/categories/infrastructure"]&q=type:"Web%20Map"%20-type:"Web%20Mapping%20Application"&sortField=modified&sortOrder=desc&start=1&num=100&f=json

https://www.arcgis.com/sharing/rest/search?categories=["/categories/infrastructure"]&q=group:47dd57c9a59d458c86d3d6b978560088%20type:"Web%20Map"%20-type:"Web%20Mapping%20Application"&sortField=modified&sortOrder=desc&start=1&num=100&f=json

This may be an agol issue, since logically these two searches should return the same results, but since they don't... I'm hoping a `searchGroupContent` function could be added to this repo. 🙏

jgravois:
AFAIK, you'd be the most likely candidate to add support for a new method like that.

its not the simplest first time contribution someone could make, but its definitely doable. your best bet would probably be adding another wrapper and making some tweaks within genericSearch to account for the additional route.
```ts
export function searchGroupContent {
  search: string | ISearchOptions | SearchQueryBuilder
): Promise<ISearchResult<IItem>> {
  return genericSearch<IItem>(search, "item");
}
```
https://github.com/Esri/arcgis-rest-js/blob/25778e3e6b16d73cc22a79a7b9bc1c97f4f90ac4/packages/arcgis-rest-portal/src/util/generic-search.ts#L39-L51

dbouwman:
@asizer fwiw I believe we do this in Hub using the main `/search` endpoint, but we pass in 
[...]

PR #675, opened 2020-03-18
Added support for group contents search

MikeTschudi:
#658

### DIFF
--- a/packages/arcgis-rest-portal/src/util/generic-search.ts
+++ b/packages/arcgis-rest-portal/src/util/generic-search.ts
@@ -35,13 +35,20 @@ export function genericSearch<T extends IItem | IGroup | IUser>(
     );
   }
 
-  let path = searchType === "item" ? "/search" : "/community/groups";
+  let path;
   switch (searchType) {
     case "item":
       path = "/search";
       break;
     case "group":
       path = "/community/groups";
+      if (
+        typeof search !== "string" &&
+        !(search instanceof SearchQueryBuilder) &&
+        search.groupId
+      ) {
+        path = `/content/groups/${search.groupId}/search`;
+      }
       break;
     default:
       // "users"
--- a/packages/arcgis-rest-portal/src/util/generic-search.ts
+++ b/packages/arcgis-rest-portal/src/util/generic-search.ts
@@ -61,14 +61,7 @@ export function genericSearch<T extends IItem | IGroup | IUser>(
       ) {
         path = `/content/groups/${search.groupId}/search`;
       } else {
-        return Promise.reject({
-          query: options.params.q,
-          total: 0,
-          start: 1,
-          num: 1,
-          nextStart: -1,
-          results: [] as IItem[]
-        } as ISearchResult<IItem>);
+        return Promise.reject(new Error("you must pass a `groupId` option to `searchGroupContent`"));
       }
       break;
     default:
--- a/packages/arcgis-rest-portal/src/util/generic-search.ts
+++ b/packages/arcgis-rest-portal/src/util/generic-search.ts
@@ -10,11 +10,19 @@ import { IItem, IGroup, IUser } from "@esri/arcgis-rest-types";
 
 import { SearchQueryBuilder } from "./SearchQueryBuilder";
 import { getPortalUrl } from "../util/get-portal-url";
-import { ISearchOptions, ISearchResult } from "../util/search";
+import {
+  ISearchOptions,
+  ISearchGroupContentOptions,
+  ISearchResult
+} from "../util/search";
 
 export function genericSearch<T extends IItem | IGroup | IUser>(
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-portal/src/util/generic-search.ts
```diff
@@ -35,13 +35,20 @@ export function genericSearch<T extends IItem | IGroup | IUser>(
...
+  let path;
   switch (searchType) {
     case "item":
       path = "/search";
       break;
     case "group":
       path = "/community/groups";
+      if (
+        typeof search !== "string" &&
+        !(search instanceof SearchQueryBuilder) &&
```
tomwayson: so is this something that should be added to `SearchQueryBuilder`?

patrickarlt: @MikeTschudi is there a particular reason that we can't search group content with a string or `SearchQueryBuilder`?

MikeTschudi: Useful, but perhaps as a new issue, an enhancement. In the meantime, this change addresses what Alison and we were seeking with #658.

patrickarlt: Couldn't we just add a new search type like `groupContent` and then add a new `searchGroupContent()` method to https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-portal/src/groups/search.ts? 
Implementation would probably be something like:

```ts
export interface ISearchGroupContentOptions extends ISearchOptions {
  q: string | SearchQueryBuilder;
  groupId: string;
}


export function searchGroupContent(options: ISearchGroupContentOptions): Promise<ISearchResult<IItem>> {
  return genericSearch<IItem>(options, "groupContent");
}
```

Then you could use `SearchQueryBuilder` as follows:

```ts
const query = new SearchQueryBuilder()
  .match("Web Mapping Application")
  .in("type")
  .match("Demo App");

searchGroupItems({
  groupId: "ABC...",
  q: query
}).then((results) => {
  console.log(request);
});
```

tomwayson:
I agree w/ @patrickarlt, would you be able to make those changes @MikeTschudi.

MikeTschudi:
We can't search group content via `genericSearch()` with a `string` or `SearchQueryBuilder` query or `ISearchOptions` options directly because none has a group id, so these input types have to be screened out. But all can be handled via the new `searchGroupContent()` and `ISearchGroupContentOptions` as Patrick's example above shows.

MikeTschudi:
@patrickarlt, @tomwayson, would you please re-review?
codecov appears to have [passed](https://codecov.io/gh/Esri/arcgis-rest-js/pull/675/commits) but gotten stuck again

packages/arcgis-rest-portal/src/util/generic-search.ts
```diff
@@ -35,14 +43,34 @@ export function genericSearch<T extends IItem | IGroup | IUser>(
...
+        path = `/content/groups/${search.groupId}/search`;
+      } else {
+        return Promise.reject({
+          query: options.params.q,
+          total: 0,
+          start: 1,
+          num: 1,
+          nextStart: -1,
+          results: [] as IItem[]
+        } as ISearchResult<IItem>);
```
patrickarlt, line 71:
```suggestion
        return Promise.reject(new Error("you must pass a `groupId` option to `searchGroupContent`"));
```

packages/arcgis-rest-portal/src/util/generic-search.ts
```diff
@@ -35,14 +43,34 @@ export function genericSearch<T extends IItem | IGroup | IUser>(
...
+      // Need to have groupId property to do group contents search,
+      // cso filter out all but ISearchGroupContentOptions
+      if (
+        typeof search !== "string" &&
+        !(search instanceof SearchQueryBuilder) &&
+        search.groupId
+      ) {
+        path = `/content/groups/${search.groupId}/search`;
+      } else {
+        return Promise.reject({
```
patrickarlt, line 64: @MikeTschudi should this reject with an error?

packages/arcgis-rest-portal/test/groups/search.test.ts
```diff
@@ -51,6 +55,42 @@ describe("groups", () => {
...
+    it("should catch search for group contents without group id", done => {
+      genericSearch(
+        {
+          q: "water"
+        },
+        "groupContent"
+      ).then(
+        () => fail(),
+        err => {
+          expect(err).toEqual(EmptyGroupSearchResponse);
```
patrickarlt, line 89: @MikeTschudi since this throws an error now should this test change?

MikeTschudi: @patrickarlt, thank you

tomwayson:
I will merge and release this Monday.

PR #677, opened 2020-03-26
getItemData() throws JSON error if data is empty

pfcstyle:
If item data is empty, the error message is `SyntaxError: The string did not match the expected pattern` in safari.
Please test the issue by the public item:  `38503de66a594dd7a656c72a34c04183`.


patrickarlt:
@tomwayson why don't you merge this so we can get it in the next release.

Issue #671, opened 2020-03-16
determineOwner should lookup username from `UserSession`


patrickarlt:
In https://github.com/Esri/arcgis-rest-js/issues/667#issuecomment-599636933 we are going to allow for "unmanaged" `UserSession`s with just a `token`. This impacts `determineOwner` which should now be async and call something like `UserSession.fetchUsername()` which will get the `username` from the portal and cache it for subsequent requests.

Thankfully since `determineOwner` is a private utility this isn't a breaking change.

Issue #667, opened 2020-03-05
 Allow for "unmanaged" UserSessions where tokens are not checked for expiration


sandromartis:
The `authentication` prop is required in both `IAddItemDataOptions` and `IUpdateItemOptions` which means that a `UserSession` has to be passed in.

This is not ideal in a server-side environment where only a token is available but not a `UserSession`.
Would it be possible to change those two types so that `authentication` is optional and instead a token can be passed in through `params`?

In `applyEdits` for example, `authentication` is also not required and a token can be passed through directly via:

```
params: {
  token: 'token'
}
```

There it seems to be handled just fine. Maybe we could change `addItemData` and `updateItem` to be similar?

Let me know if I should make a PR.

jgravois:
hey @sandromartis :wave:

> not ideal in a server-side environment where only a token is available but not a `UserSession`.

very little beyond a token is need to instantiate a `UserSession`.

```ts
const session = new UserSession({
  clientId, // comes from registering an application for OAuth 2.0
  token,
  tokenExpires
});
// ref: /packages/arcgis-rest-auth/test/UserSession.test.ts#L51-L55
```

the demo application below is an example of how server-side code can use OAuth 2.0 to help folks sign in and create a UserSession that is shared by both the client and server.

[/demos/webmap-checker-sapper](https://github.com/Esri/arcgis-rest-js/tree/master/demos/webmap-checker-sapper)

sandromartis:
hey @jgravois 😉 

I actually tried creating the `UserSession` on-the-fly but I ran into `ArcGISAuthError: Unable to refresh token.` errors.

In your example, what should `tokenExpires` be?

Our use case is a little bit different than a traditional client/server application.
We are building a serverless (and thus completely stateless) GraphQL API. The goal is that our API users pass in an ArcgisOnline token when making requests and that token gets directly forwarded to the AGO Rest API (through arcgis-rest-js).
If a token is expired the Rest API would throw an error which then gets returned to our GraphQL API user. So it's the responsibility of the user to have an updated token, same as if they would use the Rest API directly. Does that make sense?

This works nicely with just using
```
params: {
  token: 'token'
}
```

but gets a bit more tricky with a `UserSession` because it tries to refresh the token if it's expired.
It looks like I can just set `tokenExpires` to some date in the future, but it seems like a bit of a hacky workaround.

jgravois:
> Does that make sense?

[...]

PR #678, opened 2020-03-27
Add support for unmanaged sessions, async determineOwner

patrickarlt:
This resolves https://github.com/Esri/arcgis-rest-js/issues/671 and https://github.com/Esri/arcgis-rest-js/issues/667. Doing the following:

1. `UserSession` now can just accept the `token` option and nothing else when working against ArcGIS Online. `token` and `portal` are required for ArcGIS Enterprise. Tokens are assumed to be valid if `tokenExpires` is not passed. Previously this was an error so it is not a breaking change.
2. Since we might not know the `username` for a given `token`, utility methods like `determineOwner()` now user `UserSession.getUsername()`/`UserSession.getUser()` which use `/community/self` to fetch the `username` for the current token. `UserSession.getUser()` has no external changes and `UserSession.getUsername()` is new.
3. `determineOwner()` is now async since it might call into `UserSession.getUsername()` which also might be async. This function is an undocumented private util so no breaking change.

Even though tests have changed there shouldn't be any breaking changes this release since most changes were from `community/{USERNAME}` to `community/self`. Some tests for `reassignItem` changed to create a new session for every test since reusing the same session doesn't work.

### DIFF
--- a/packages/arcgis-rest-portal/src/items/helpers.ts
+++ b/packages/arcgis-rest-portal/src/items/helpers.ts
@@ -1,7 +1,7 @@
 /* Copyright (c) 2017-2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IRequestOptions, ArcGISRequestError } from "@esri/arcgis-rest-request";
 import { IItemAdd, IItemUpdate, IItem } from "@esri/arcgis-rest-types";
 import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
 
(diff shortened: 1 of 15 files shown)


packages/arcgis-rest-portal/src/items/helpers.ts
```diff
@@ -238,15 +238,20 @@ export function serializeItem(item: IItemAdd | IItemUpdate | IItem): any {
...
+export function determineOwner(requestOptions: any): Promise<string> {
   if (requestOptions.owner) {
-    return requestOptions.owner;
-  }
-  if (requestOptions.item && requestOptions.item.owner) {
-    return requestOptions.item.owner;
+    return Promise.resolve(requestOptions.owner);
+  } else if (requestOptions.item && requestOptions.item.owner) {
+    return Promise.resolve(requestOptions.item.owner);
+  } else if (requestOptions.authentication) {
```
tomwayson: This will work for `UserSession`, but not for `ApplicationSession`, right?
```suggestion
  } else if (requestOptions.authentication && requestOptions.authentication.getUsername) {
```

patrickarlt:
@tomwayson I merged your suggestion. I don't think ti makes much different though since everything that uses `determineOwner` specifies a `UserSession`  for `authentication`.

patrickarlt:
I'm merging this I think CodeCov is';t talking to GitHub

Issue #685, opened 2020-04-06
Categories parameter not handled correctly for item and group searches


MikeTschudi:
The categories parameter may be used up to 8 times in a query.  Each category listed with a single one of these parameters is ORed together; the set of categories parameters is ANDed together.

arcgis-rest-js merges the list of strings provided to categories into a single parameter, thus ORing everything.

cf. [item search](https://developers.arcgis.com/rest/users-groups-and-items/search.htm) and [group content search](https://developers.arcgis.com/rest/users-groups-and-items/group-content-search.htm)

patrickarlt:
@MikeTschudi can you provide code examples?

MikeTschudi:
```
<!doctype html>
<html lang="en" dir="ltr">
<body>
  <script src="lib/require.js"></script>
  <script>
    require.config({
      paths: {
        "@esri/arcgis-rest-auth": "https://unpkg.com/@esri/arcgis-rest-auth@2.10.0/dist/umd/auth.umd.min",
        "@esri/arcgis-rest-portal": "https://unpkg.com/@esri/arcgis-rest-portal@2.10.0/dist/umd/portal.umd.min",
        "@esri/arcgis-rest-request": "https://unpkg.com/@esri/arcgis-rest-request@2.10.0/dist/umd/request.umd.min"
      }
    });

    require(["@esri/arcgis-rest-auth", "@esri/arcgis-rest-portal"], function (auth, portal) {

      // From https://developers.arcgis.com/rest/users-groups-and-items/search.htm
      // "Example: search for items with water or forest group category in the US"
      // (/Categories/Water OR /Categories/Forest) AND /Region/United States
      portal.searchItems({
        q: "*",
        params: {
          categories: ["/Categories/Water,/Categories/Forest", "/Region/United States"]
        }
      }).then(
        results => console.log(JSON.stringify(results, null, 2)),
        error => console.log(JSON.stringify(error, null, 2))
      );

    });
  </script>
</body>
</html>
```

Actual request URL: `https://www.arcgis.com/sharing/rest/search?f=json&categories=%2FCategories%2FWater%2C%2FCategories%2FForest%2C%2FRegion%2FUnited%20States&q=*`
* f: json
* categories: /Categories/Water,/Categories/Forest,/Region/United States
* q: *

[...]

PR #686, opened 2020-04-06
Implements search filter by categories using AND

MikeTschudi:
Current use of categories ORs all categories terms; this PR implements format used by REST API to distinguish between OR and AND.

see #685 

note: built upon, but does not require, approved PR #675


MikeTschudi:
@patrickarlt, can we revisit this? We'd like to implement AND in category searching.

MikeTschudi:
In TravisCI, passes "Node.js: lts/*" and "Node.js: 10"; fails "Node.js: node" with a lerna problem

MikeTschudi:
Replaced by [Support array of arrays when encoding query strings #779](https://github.com/Esri/arcgis-rest-js/pull/779)

PR #689, opened 2020-04-11
Supply folder parameter during creation so that a subsequent move is not needed

MikeTschudi:
The REST API's [createService](https://developers.arcgis.com/rest/users-groups-and-items/create-service.htm) allows a folder to be specified, so instead of creating a service and then moving it to the desired folder, we simply create it in the desired folder.

Originally didn't think that this was possible ([Added create feature service #307](https://github.com/Esri/arcgis-rest-js/pull/307)), but @chris-fox suggested saving the extra round trip to the server.


__reinstallNodeModules_rest.bat
```diff
@@ -0,0 +1,15 @@
...
+call rmdir/q/s packages\arcgis-rest-geocoding\node_modules
+call rmdir/q/s packages\arcgis-rest-portal\node_modules
+call rmdir/q/s packages\arcgis-rest-request\node_modules
+call rmdir/q/s packages\arcgis-rest-routing\node_modules
+call rmdir/q/s packages\arcgis-rest-service-admin\node_modules
+
+call npm install
+
+call npm run test:all
+
```
dbouwman, line 13: I think you could achieve this in a cross-platform manner using `rimraf` in a npm script, similar to how the `docs:serve` script does it. Alternatively you could also run [windows bash](https://www.howtogeek.com/249966/how-to-install-and-use-the-linux-bash-shell-on-windows-10/)

__createTag.bat
```diff
@@ -0,0 +1,4 @@
+for %%F in ("%cd%") do set "folder=%%~nxF"
```
dbouwman, line 1: I'm not really clear what this script does - when we create releases and/or tags, normal git commands seem to work... do those not work for you? or is this for some other purpose?

MikeTschudi: Thanks, @dbouwman; these are errors--I have a different .gitignore file that excludes tool files prefixed with "__".  I'll get rid of them.

dbouwman: sounds good! thanks!

tomwayson:
Thanks for reviewing this @dbouwman 

Let's hold on merging this until I get https://github.com/Esri/arcgis-rest-js/pull/703 merged and released

tomwayson:
Merge and release at will https://github.com/Esri/arcgis-rest-js/pull/703#issuecomment-627346705

PR #695, opened 2020-04-27
feat(arcgis-rest-portal): add getPortalSettings function

mjuniper:
AFFECTS PACKAGES:
@esri/arcgis-rest-portal


mjuniper:
I'm not really sure because a) my TS-fu is weak and b) I'm not certain that I know the entirety of what that response could look like. What I was trying to express was that I'm pretty sure it will have those 4 props but it will likely have others....

tomwayson:
> I was trying to express was that I'm pretty sure it will have those 4 props but it will likely have others....

You did the right thing

Issue #557, opened 2019-04-29
use an "X-Esri-Authorization" header to pass along tokens (not a query parameter)


jgravois:
see https://github.com/Esri/arcgis-rest-js/issues/290#issuecomment-414375805 for more info.

jgravois:
i did some testing with this and currently hosted feature services honor tokens pass via an `X-Esri-Authorization` header, but endpoints like [`/search`](https://developers.arcgis.com/rest/users-groups-and-items/search.htm) do not. 

probably best to put this enhancement "on ice" for now.

ssylvia:
Did some basic testing with Postman and it appears that `/search` endpoint does work with this header. After communication with others, it appears that support was added for this header a few years back for AGO. It's also supported in portal as of `10.5.1`.

I'll take a stab at adding support for this in a PR. This is my plan if you have any feedback before I start:

1. Add new options to IRequestOptions
	- `secureToken: boolean` (hides the token in POST body or GET header)
	- `mode: FetchMode` (standard fetch mode: cors, no-cores, etcs). Looks like custom headers are only supported over fetch when cors mode.
2. If `secureToken` is true, we'll convert all GET requests to POST, unless the mode is also cors, in which case we'll append the token in the `X-ESRI-AUTH..` header.
3. I'll leave both of these off by default so it can be merged in a patch release. Then we can add them to the default setting at a major release if we choose to.

jgravois:
> it appears that /search endpoint does work with this header.

i'm sure i was just doing something wrong. it's been awhile, but i don't remember `Bearer ` being in the examples last year.

> If secureToken is true, we'll convert all GET requests to POST

just curious, what is the benefit of that?

ssylvia:
> If secureToken is true, we'll convert all GET requests to POST
just curious, what is the benefit of that?

`X-ESRI-AUTHORIZATION` header is not allowed in the request headers when fetch is using `no-cors`, which is default. To prevent the token being leaked in query param, we will convert to `POST` so it is placed in body.

PR #697, opened 2020-04-27
Secure token in POST body or auth header

ssylvia:
Adds two option to the IRequestOption:
- <s>`fetchMode`: specify the mode to fetch with (cors, no-cors, etc)</s>
- `secureToken`: hides the token of `GET` request in `X-Esri-Authorization` header.

Both options are disabled by default. This could become the default at a major release.

Fixes: #557

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -287,8 +287,8 @@ export function request(
       } = {};
       
       if (fetchOptions.method === "GET") {
-        // Prevents token from being passed in query params when secureToken option is used.
-        if (params.token && options.secureToken) {
+        // Prevents token from being passed in query params when hideToken option is used.
+        if (params.token && options.hideToken) {
           requestHeaders["X-Esri-Authorization"] = `Bearer ${params.token}`
           delete params.token;
         }
@@ -307,7 +307,7 @@ export function request(
           fetchOptions.method = "POST";
 
           // Add token back to body with other params instead of header
-          if (token.length && options.secureToken) {
+          if (token.length && options.hideToken) {
             params.token = token;
             delete requestHeaders["X-Esri-Authorization"];
           }
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -241,7 +241,7 @@ export function request(
     );
   }
 
-  const { httpMethod, authentication, rawResponse } = options;
+  const { httpMethod, fetchMode, authentication, rawResponse } = options;
 
   const params: IParams = {
     ...{ f: "json" },
@@ -252,6 +252,7 @@ export function request(
 
   const fetchOptions: RequestInit = {
     method: httpMethod,
+    mode: fetchMode,
     /* ensures behavior mimics XMLHttpRequest.
     needed to support sending IWA cookies */
     credentials: "same-origin"
@@ -281,7 +282,17 @@ export function request(
         params.token = token;
       }
 
+      // Mixin headers from request options
+      fetchOptions.headers = {
+        ...options.headers
+      };
+      
       if (fetchOptions.method === "GET") {
+        // Fetch `mode: "no-cors"` only allows a limited set of headers in the request.
+        if (params.token && options.secureToken && typeof fetchOptions.mode !== 'undefined' && fetchOptions.mode !== 'no-cors') {
+          fetchOptions.headers["X-Esri-Authorization"] = `Bearer ${params.token}`
+          delete params.token;
+        }
         // encode the parameters into the query string
         const queryParams = encodeQueryString(params);
         // dont append a '?' unless parameters are actually present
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -281,16 +282,27 @@ export function request(
...
 
         if (
-          options.maxUrlLength &&
-          urlWithQueryString.length > options.maxUrlLength
+          (options.maxUrlLength &&
+            urlWithQueryString.length > options.maxUrlLength) || 
+            (params.token && options.secureToken)
         ) {
           // the consumer specified a maximum length for URLs
           // and this would exceed it, so use post instead
```
tomwayson, line 308: Let's add to this comment like
```js
          // this would exceed the maximum length for URLs specified by the consumer
          // or the user wants to secure the token in no-cors mode
          // so use post instead
```

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -14,6 +15,10 @@ export interface IRequestOptions {
    * The HTTP method to send the request with.
    */
   httpMethod?: HTTPMethods;
+  /**
+   * The fetch mode to send the request with.
```
tomwayson, line 19: Let's add link to some docs on this, maybe https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Syntax

packages/arcgis-rest-request/src/utils/FetchMode.ts
```diff
@@ -0,0 +1,6 @@
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+/**
+ * Fetch mode used by the ArcGIS REST API.
+ */
+export type FetchMode = "cors" | "no-cors" | "same-origin";
```
tomwayson, line 6: I don't know if we need a type for this. Can we just put the string union inline in `IRequestOptions` for now?

packages/arcgis-rest-request/src/request.ts
```diff
@@ -281,16 +282,27 @@ export function request(
         params.token = token;
       }
 
+      // Mixin headers from request options
+      fetchOptions.headers = {
+        ...options.headers
+      };
+
```
tomwayson, line 289: I'm not sure we should move this above the blocks below. I know why you did, but I think if someone passes in an `X-Esri-Authorization` as well as `secureToken`, then that should trump what we generate.

packages/arcgis-rest-request/test/request.test.ts
```diff
@@ -171,6 +171,68 @@ describe("request()", () => {
...
+        httpMethod: 'GET',
+        secureToken: true
+      }
+    )
+      .then(response => {
+        const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
+        expect(url).toEqual(
+          "https://www.arcgis.com/sharing/rest/content/items/43a8e51789044d9480a20089a84129ad/data"
+        );
+        expect(options.body).toContain("token=token");
```
tomwayson, line 198: should we also expect the method to be `POST`?

packages/arcgis-rest-request/src/request.ts
```diff
@@ -295,6 +305,12 @@ export function request(
           // the consumer specified a maximum length for URLs
           // and this would exceed it, so use post instead
           fetchOptions.method = "POST";
+
+          // Add token back to body with other params instead of header
+          if (token.length && options.secureToken) {
+            params.token = token;
+            delete requestHeaders["X-Esri-Authorization"];
+          }
```
tomwayson, line 313: I think you are right to move the token out of the header and into the body if we have to use POST.

I only wonder if we should check for the existence of the token in the header rather than the re-check the criteria for putting it there.

```suggestion
          const headerToken = requestHeaders["X-Esri-Authorization"];
          if (headerToken) {
            // Add token back to body with other params instead of header
            params.token = headerToken;
            delete requestHeaders["X-Esri-Authorization"];
          }
```

ssylvia: @tomwayson We can do that but it will get more complicated. When passed as a header, the value is different. It's actually `Bearer ${token}`. Thought it would be better to use our original which we know is not modified elsewhere.

tomwayson: let's leave it as is

tomwayson:
Thanks @patrickarlt 

Some name ideas:
- `noTokenInUrl`: short, to the point. A little "No Country for Old Men" for my taste
- `useAuthHeaderForGet`: this is a horrible name, but it does what it says
- `useAuthHeader`: less horrible, but in this case it might be a little confusing when we don't for POST requests. This could be a string like `'get' | 'always'`, but we'd have to change the current logic to honor that.

mpayson:
hideToken?

tomwayson:
Thanks for this contribution and your patience, @ssylvia, it's been released as [v2.13.0](https://github.com/Esri/arcgis-rest-js/releases/tag/v2.13.0)

aert-nearmap:
I'm curious, why was the choice made to only put the token into the header for a GET and not a POST too (for consistency if anything)?

PR #701, opened 2020-05-05
Added function getItemResource

MikeTschudi:
To get individual resources by name rather than all of them


packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -136,9 +136,35 @@ export function getRelatedItems(
...
+  id: string,
+  resourceName: string,
+  requestOptions?: IRequestOptions
+): Promise<any> {
+  const url = `${getPortalUrl(requestOptions)}/content/items/${id}/resources/${resourceName}`;
+
+  const options: IRequestOptions = {
+    ...requestOptions,
+    params: {
+      f: null
```
dbouwman, line 157: Have you checked how this works w/ json or text resources? I know StoryMaps and ExB's use json resources (which for reasons need to be sent up as Blobs)

dbouwman: ah - I recall... this gets interesting b/c the `request` function tries to do all the magical parsing. So I think you need to leverage the `rawResponse` option, and then in this fn, based on the file extension, decide how to parse the responose - response.blob() for image types, response.json() for .json, response.text() for text-y things

MikeTschudi: This works with json; it's just a Blob of json instead of a raw response of json.

I'm avoiding interpretation because the file extension is not sufficient info: the resource could be a partial file (files > 50MB have to be broken up), or it could be a file with an extension not supported as a resource camouflaged as a ZIP file.  With this implementation, everything is returned as a Blob, corresponding to the arcgis-rest-request request handling of an unknown `f` param type.

tomwayson: > This works with json; it's just a Blob of json instead of a raw response of json.

Once someone's gotten that Blob, how do they know it's JSON inside, and how do they get it out?

MikeTschudi:
In the case where this function is planned to be used, the calling function doesn't know the type of the resource either; that information is stored elsewhere and matched to the resource later.

Because of the need to deceive AGO at times, it seems best to not try to guess what the calling function is looking for. Just get and return the resource.

packages/arcgis-rest-portal/src/items/get.ts
MikeTschudi: Metadata stored elsewhere holds the type of the Blob

tomwayson:
FYI - see the discussion in #758 for how we handled the above file type issue for item info files. I suspect we could generalize the pattern we came up w/ there to apply to resources too.

tomwayson:
I'm working on refactoring the code behind `getItemInfo()` to work for resources too. I'll ping you on the PR @MikeTschudi

tomwayson:
See #760

gavinr:
Closing this PR for now. We can follow-up with a separate PR to use Tom's work if applicable.

PR #703, opened 2020-05-11
Patch hideToken for browser CORS support

ssylvia:
- Turns out the sharing API does not support `X-Esri-Authorization` header as part of the `OPTIONS` request which is required by browser for CORS preflight request.
- For browser, we will force the request to use `POST`

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -288,7 +288,9 @@ export function request(
       
       if (fetchOptions.method === "GET") {
         // Prevents token from being passed in query params when hideToken option is used.
-        if (params.token && options.hideToken) {
+        if (params.token && options.hideToken && 
+          // Sharing API does support preflight check required by modern browsers https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -288,7 +288,10 @@ export function request(
       
       if (fetchOptions.method === "GET") {
         // Prevents token from being passed in query params when hideToken option is used.
-        if (params.token && options.hideToken) {
+        /* istanbul ignore if - window is always defined in a browser. Test case is covered by Jasmine in node test */
+        if (params.token && options.hideToken && 
+          // Sharing API does support preflight check required by modern browsers https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
```
tomwayson, line 293: should this be:

```suggestion
          // Sharing API does not support preflight check required by modern browsers https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
```

tomwayson:
Thanks @ssylvia - this has been released as [2.13.1](https://github.com/Esri/arcgis-rest-js/releases/tag/v2.13.1).

Issue #699, opened 2020-04-28
Make refresh token TTLs consistent


mpayson:
According to the REST doc the [expiration /authorize](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm) parameter sets the refresh token TTL for server-based OAuth. Currently we handle these differently.

Expiration is defined by `duration` (default 20160) in [`authorize`](https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L455) while the refresh token TTL is set by `refreshTokenTTL` param (default 1440) in [`exchangeAuthorizationCode`](https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L481) without actually specifying this to the server.

A couple ideas to approach:
1. Update defaults to be the same
2. Respect `refreshTokenTTL` instead of `duration` in `authorize` (or fall back to `duration` for compatibility)?
3. Maybe pass in the `refreshTokenTTL` as state (ref #660 ) then set it in `exchangeRefreshToken` to ensure it's consistent between the two methods?

Let me know your thoughts and happy to PR if there's an updated approach!

patrickarlt:
This will be resolved in the upcoming 4.0 release which is currently on the `beta` branch

Issue #660, opened 2020-02-24
Include state in server OAuth?


mpayson:
I noticed state isn't passed to `authorize` for server-based OAuth, should this be included?

https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L460

jgravois:
from the JSDoc it looks like it was a conscious choice in this library to flag state as `@browserOnly`

https://github.com/Esri/arcgis-rest-js/blob/25778e3e6b16d73cc22a79a7b9bc1c97f4f90ac4/packages/arcgis-rest-auth/src/UserSession.ts#L139-L144

but looking at the REST API doc, I'm not seeing anything that indicates it isn't valid for server-based flows.

https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm

> should this be included?

do you have a use case that is improved by including it?
if so, if you've tested it and it works, that sounds like enough of a reason to me to open a PR and continue the discussion there.

mpayson:
👋 @jgravois 

📝 Partners often have their own log in system and want to allow Esri-users to connect their ArcGIS account. This is commonly done on the server to store refresh tokens. So, the server needs to know in the `redirect` who the user is in the partner's system. One approach would be sessions, another would be the `state` param for stateless auth mechanisms (eg JWTs passed from the client)
✅ It works

I've been playing around with these workflows, so curious to hear if this is considered best practice or not. Thanks and LMK if you'd like a PR on my end!

jgravois:
👋 old friend.

given the REST doc I linked to above and your own feedback, I think the library would benefit from passing through `state`. feel free to open a PR at your leisure.

patrickarlt:
This will be resolved at 4.0 which is currently on the `beta` branch.

PR #704, opened 2020-05-20
Update server-based oauth 2.0 for state and token consistency

mpayson:
Addresses #699 & #660 as well as a couple other tweaks:
* change the 'duration' URL parameter to 'expiration' in `authorize()` to match the [REST endpoint](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)
* update 'refreshTokenExpires' calculation to reflect that 'refreshTokenTTL' is in minutes

Should we also update the default 'duration' value for browser-based OAuth 2.0 to be 2 hours like the [REST api](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)? Currently it's 2 weeks.


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -451,17 +451,22 @@ export class UserSession implements IAuthenticationManager {
...
-    const { portal, clientId, duration, redirectUri }: IOAuth2Options = {
+    const { portal, clientId, duration, redirectUri, refreshTokenTTL, state }: IOAuth2Options = {
       ...{ portal: "https://arcgis.com/sharing/rest", duration: 20160 },
       ...options
     };
+    const expiration = refreshTokenTTL ? refreshTokenTTL : duration;
+
+    let url = `${portal}/oauth2/authorize?client_id=${clientId}&expiration=${expiration}&response_type=code&redirect_uri=${encodeURIComponent(
+      redirectUri
+    )}`;
```
jgravois, line 462: the tweak below would simplify things. am i losing anything in translation?

```suggestion
    const { portal, clientId, duration, redirectUri, refreshTokenTTL, state }: IOAuth2Options = {
      ...{ portal: "https://arcgis.com/sharing/rest", refreshTokenTTL: 20160 },
      ...options
    };
 
    let url = `${portal}/oauth2/authorize?client_id=${clientId}&expiration=${refreskTokenTTL}&response_type=code&redirect_uri=${encodeURIComponent(
      redirectUri
    )}`;
```

mpayson: I was thinking backwards compatibility... since the `duration` param previously set the refresh token expiration on the server, it should still be respected if the developer specified it, but update the default behavior to respect the `refreshTokenTTL` since if set this is likely the intent

jgravois: makes sense. thanks for humoring me.

noahmulfinger:
@patrickarlt Since you initially wrote a lot of this, you might want to take a look.

gavinr:
Hm, looks like I messed up this merge somehow. I'll close this and re-open a separate PR with the same changes.

PR #707, opened 2020-06-18
Make cleanUrl resilient to null urls

dbouwman:
Add a guard to `cleanUrl` so it does not try to call `.trim()` if the fn is not passed a string.

Does not require a release, but thought I'd add it as we found this while debugging something else in Hub


packages/arcgis-rest-request/src/utils/clean-url.ts
```diff
@@ -5,6 +5,10 @@
  * Helper method to ensure that user supplied urls don't include whitespace or a trailing slash.
  */
 export function cleanUrl(url: string) {
+  // Guard so we don't try to trim something that's not a string
+  if (typeof url !== "string") {
+    return url;
+  }
```
jgravois, line 11: just as long as the TypeScript gods never see this 😆

dbouwman: Howdy John! That's the thing w/ consuming a ts lib, outside typescript. And there is also the fact that TS does not assign default values to `null` - i.e. if we have `clearUrl(url: string = '')`, calling `cleanUrl(null) !== cleanUrl("")`

jgravois: i know you're still in Emberland, i just couldn't resist the urge to give you a hard time from the grave 😉

Issue #711, opened 2020-06-28
UserSession. completeOAuth2 doesn't work when popup is true


pfcstyle:
The oauth2 popup window didn't close and callback function didn't be called.

arcgis-rest-auth version: 2.13.0
Chrome version: Version 83.0.4103.97 (Official Build) (64-bit)

I noticed this in `UserSession.ts`:
```
    if (
          popup &&
          win.opener &&
          win.opener.parent &&
          win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`]
        ) {
          const handlerFn =
            win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`];
          if (handlerFn) {
            handlerFn(
              error ? JSON.stringify(error) : undefined,
              JSON.stringify(oauthInfo)
            );
          }
          win.close();
          return undefined;
        }
```

But I found the `__ESRI_REST_AUTH_HANDLER_${clientId}` was exists in win.opener only, so it can't get the handleFn.

pfcstyle:
So I submit the pr  to  get handlerFn from win.opener if it can't get the callback from win.opener.parent. I think it can resolve this issue and won't result in other issues.

```
    if (
          popup &&
          win.opener &&
          win.opener.parent &&
          (win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`] || win.opener[`__ESRI_REST_AUTH_HANDLER_${clientId}`])
        ) {
          const handlerFn =
            win.opener.parent[`__ESRI_REST_AUTH_HANDLER_${clientId}`] || win.opener[`__ESRI_REST_AUTH_HANDLER_${clientId}`];
          if (handlerFn) {
            handlerFn(
              error ? JSON.stringify(error) : undefined,
              JSON.stringify(oauthInfo)
            );
          }
          win.close();
          return undefined;
        }
```

https://github.com/Esri/arcgis-rest-js/pull/712

dbouwman:
Thanks @pfcstyle - I was literally about to open an issue about this *exact* issue!

To date this has not been an issue but we've started doing some dev't work using [Storybook](https://storybook.js.org/) and that loads the components into iframes. When we started using auth in that scenario, we ran into this issue where window.parent (the page hosting the iframe) did not have the handler, but window.opener (the page loaded into the iframe) did.

Thanks for the PR - we'll review it asap and move to get it merged.

dbouwman:
[...]

PR #714, opened 2020-07-01
fix(arcgis-rest-auth): enable oAuth from within an IFrame

dbouwman:
Adds a new check for oauth handler on window.opener, and cleans up previous handler checks.

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #711

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,7 @@
+codecov:
+  require_ci_to_pass: yes
+
 coverage:
   round: up
-  range: 0..10
+  range: 90..95
+  precision: 2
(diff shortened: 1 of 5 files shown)


dbouwman:
I'm going to verify this by `yalc`-ing this into a StoryBook and verifying the iframed oauth process works

dbouwman:
Verified this works in scenarios where the caller is in an `<iframe>` in a page... it correctly (at least in chrome) locates the handler on `window.opener`. Unclear what's up the travis and codecov

PR #717, opened 2020-07-15
Adding Get User Content

mikeringrose:
Adds support for fetching [user content](https://developers.arcgis.com/rest/users-groups-and-items/user-content.htm).

### DIFF
--- a/packages/arcgis-rest-portal/src/items/content.ts
+++ b/packages/arcgis-rest-portal/src/items/content.ts
@@ -8,7 +8,8 @@ import { getPortalUrl } from "../util/get-portal-url";
 export type UnixTime = number;
 
 export interface IUserContentRequestOptions extends IPagingParams, IRequestOptions {
-  username: string;
+  username?: string;
+  folder?: string;
 }
 
 export interface IFolder {
@@ -22,6 +23,7 @@ export interface IUserContentResponse extends IPagedResponse {
   username: string;
   currentFolder?: IFolder;
   items: IItem[];
+  folders: IFolder[];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-portal/src/items/content.ts
```diff
@@ -0,0 +1,68 @@
...
+import { getPortalUrl } from "../util/get-portal-url";
+
+export type UnixTime = number;
+
+export interface IUserContentRequestOptions extends IPagingParams, IRequestOptions {
+  username?: string;
+  folder?: string;
+}
+
+export interface IFolder {
```
tomwayson, line 15: I'm surprised we don't have this defined elsewhere already. Maybe this should go in the types package? cc @MikeTschudi

mikeringrose: I didn't find it define else where. I agree it should probably be in types. I'll move it.

packages/arcgis-rest-portal/src/items/content.ts
```diff
@@ -0,0 +1,68 @@
...
+ *    authentication
+ * })
+ * ```
+ * Returns a listing of the user's content. If the `username` is not supplied, it defaults to the username of the authenticated user. If `start` is not specificed it defaults to the first page.
+ * If the `num` is not supplied it is defaulted to 10. See the [REST Documentation](https://developers.arcgis.com/rest/users-groups-and-items/user-content.htm) for more information.
+ *
+ * @param requestOptions - Options for the request
+ * @returns A Promise<IUserContentResponse>
+ */
+export const getUserContent = (requestOptions: IUserContentRequestOptions): Promise<IUserContentResponse> => {
```
mikeringrose, line 47: @tomwayson @MikeTschudi I didn't see [IFolderIdOptions](https://github.com/mikeringrose/arcgis-rest-js/blob/c3682322f7aca69c0dd3907a603304d232d8b43c/packages/arcgis-rest-portal/src/items/helpers.ts#L22) until now. I think `requestOptions` would probably be best defined as the following:
```
requestOptions: IFolderIdOptions & IPagingParams
```
The only issue is that the documentation generator doesn't seem to like intersection types. I could also define it using multiple inheritance like:
```
export interface IUserContentRequestOptions extends IPagingParams, IFolderIdOptions {}
```
Thoughts?

tomwayson: I don't know enough about TS (nor do I want to) to have an opinion on that.

packages/arcgis-rest-portal/src/items/content.ts
MikeTschudi: I didn't, either. Nice to have this function.

packages/arcgis-rest-portal/src/items/content.ts
MikeTschudi: I think that multiple inheritance is more readable here than trying to chain the types.

mikeringrose: Actually the type i need is slightly different. Mine allows the owner to be optional. I'll rename my properties to be consistent with the one that already exists.

PR #724, opened 2020-08-13
add note about manually updating peerDependencies

tomwayson:


RELEASE.md
```diff
@@ -20,12 +20,16 @@ npm run release:review
 
 **IMPORTANT** 
 
-Before publishing, you will likely need to make a few chages to `CHANGELOG.md` b/c the system to automates this is [far from perfect](https://github.com/Esri/arcgis-rest-js/issues/688). For example:
+**Before publishing, you will likely need to make a few chages manually.**
```
jgravois, line 23:
```suggestion
**Before publishing, you will likely need to make a few changes manually.**
```

RELEASE.md
```diff
@@ -20,12 +20,16 @@ npm run release:review
...
+
+First, lerna does **not** manage `peerDepedencies`. We have to update them ourselves as needed. For example, if you've added a feature to a package (i.e. bumping minor) _and you've also updated other packages w/in the monorepo to use that feature_, you will need to manually update the `peerDependencies` of those other packages. Note that if you add features but they are not used w/in the monorepo, then you should **not** have to update any `peerDependencies`.
```
jgravois, line 25:
```suggestion
First, lerna does **not** manage `peerDependencies`. We have to update them ourselves as needed. For example, if you've added a feature to a package (i.e. bumping minor) _and you've also updated other packages w/in the monorepo to use that feature_, you will need to manually update the `peerDependencies` of those other packages. Note that if you add features but they are not used w/in the monorepo, then you should **not** have to update any `peerDependencies`.
```

Issue #726, opened 2020-08-14
Documentation: Node version support


gavinr:
What version(s) of NodeJS do we support? [This guide page](https://esri.github.io/arcgis-rest-js/guides/) says `ArcGIS REST JS is supported in Node.js 6.x and 8.x` - I think that's out of date, because our travis tests [looks like](https://github.com/Esri/arcgis-rest-js/blob/4b3905ca6517443c9237a44c0fc3249e579db8f5/.travis.yml#L8-L9) we're testing against the latest major version (v14 currently) and the latest LTS version (v12 currently).

jgravois:
the guide page should definitely be updated to say that Node.js 6.x and above are supported.
I don't see any reason not to ask Travis to add a few more versions to the matrix while you're at it.

```iml
node_js:
  - "node" # 14
  - "lts/*" # 12
  - 10
  - 8
  - 6
```

gavinr:
Thanks @jgravois. Would it make sense to define an [`engines` property](https://docs.npmjs.com/files/package.json#engines) in our `package.json` file? And then would be great if we could configure travis to test all major versions that are covered in the "engines" property instead of having it duplicated in `.travis.yml` ... but doesn't look like that's currently supported, based on some googling.

jgravois:
i've never seen anything that used 'engines' but it certainly couldn't hurt anything.

PR #727, opened 2020-08-14
nodejs versions

gavinr:
#726


jgravois:
not the end of the world, but it looks like this is causing duplicate builds on 14/12.

https://travis-ci.org/github/Esri/arcgis-rest-js/builds/718001512

Screen Shot 2020-08-14 at 11 07 49 AM

PR #737, opened 2020-08-25
Add removeGroupUsers. Refactor shareItemWithGroup and unshareItemWith…

rweber-esri:
…Group to unlock additional sharing capabilities natively supported by the REST API. Admin users temporarily join groups when necessary to perform shares, then get removed if added.

[144791](https://esriarlington.tpondemand.com/entity/144791-search-filter-and-select-groups-to)

The current sharing implementation was too restrictive for Hub's needs. This PR:

1. Adds `removeGroupUsers` method that batches user requests w/ max 25 users per request per [REST API docs](https://developers.arcgis.com/rest/users-groups-and-items/remove-users-from-group.htm) similar to how `addGroupUsers` is implemented.
2. Separates logic for existing `shareItemWithGroup` and `unshareItemWithGroup` into separate methods to make logic easier to reason about.
3. Loosens unshare restrictions that prevented org admins from unsharing org items they do not own from groups they are not members of.
4. Item org admins are temporarily added to edit/update groups as a `member` before sharing, then remove themselves from the group afterwards (only happens if the admin was added, not if they were already a member)
5. Item org admin are temporarily added to view groups as a `member` before sharing, then remove themselves from the group afterwards (only happens if the admin was added, not if they were already a member)


rweber-esri:
> there's a lot here. I just glanced over it. I assume the logic is copied out of existing logic in the Hub app and therefore sound.

@tomwayson Most of the work here was splitting out (most copied verbatim) sharing vs unshare logic that was previously all in a single method with special handling conditions intermingled which made it difficult to reason about and manage. Additionally, the previous implementation enforced some limitations that prevented the calls from going through to the API despite the API supporting those use cases, which forced us in `Hub` to do things like temporarily add/promote org admins' group membership just so the share call would go through to the API. This loosens those restrictions & performs any membership adding/promotions that are required for the share call to succeed.

tomwayson:
I really think it would be best to have @dbouwman review this b/c he understands the dark magic of the sharing API better than anyone and would know if there was any reason to have the sharing fns grouped.

Are there breaking changes in this?

rweber-esri:
> I really think it would be best to have @dbouwman review this b/c he understands the dark magic of the sharing API better than anyone and would know if there was any reason to have the sharing fns grouped.
> 
> Are there breaking changes in this?

There should be no _breaking_ changes; everything that _could_ be shared/unshared previously should be able to be shared/unshared in the future. Some changes in this PR prevent these API wrapper methods from short-circuiting via throwing errors in a couple legitimate sharing scenarios that were previously blocked (but supported by the API) and forced us to manually work around on our end. The existing tests were split out as part of this work, too, which _should_ validate no regressions. I added additional tests to cover the sharing scenarios that were unlocked. Regardless, I agree, the more eyes the better.

dbouwman (approved):
Main question is if this constitutes a breaking change. The API is the same, but it looks like the xhrs behind the scenes may have changed... and thus any projects using this, and using fetchMock, will run into problems... not exactly sure what to recommend

tomwayson:
it's not a breaking change on the account of fetchMock - the API calls made under the hood are an implementation detail that consumers should not rely on.

As long as the API has not changed, and what is returned by those functions is not materially different (despite the fact that those fns are now calling different end points), then this is not a breaking change.

PR #743, opened 2020-09-01
Updates search query builder error message

phpmaps:
I've found if you try and pass bad / mangled dates to the query builder that it will not resolve the expression and omit the intended expression from the query.  When throwing the warning, it would be nice to draw attention to this case / reason for failure.

bad dates happen... : )

### DIFF
--- a/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts
+++ b/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts
@@ -216,7 +216,7 @@ export class SearchQueryBuilder implements IParamBuilder {
     if (this.hasTerms) {
       warn(
         // apparently-p*rettier-ignore causes prettier to strip *all* comments O_o
-        `\`from(...)\` is not allowed after \`match(...)\` try using \`.from(...).to(...).in(...)\`. Your query was not modified.`
+        `\`from(...)\` is not allowed after \`match(...)\` try using \`.from(...).to(...).in(...)\`. Optionally, you may see this because dates are incorrectly formatted, not in milliseconds.  Your query was not modified.`
       );
       return this;
     }
@@ -238,7 +238,7 @@ export class SearchQueryBuilder implements IParamBuilder {
     if (this.hasTerms) {
       warn(
         // apparently-p*rettier-ignore causes prettier to strip *all* comments O_o
-        `\`to(...)\` is not allowed after \`match(...)\` try using \`.from(...).to(...).in(...)\`. Your query was not modified.`
+        `\`to(...)\` is not allowed after \`match(...)\` try using \`.from(...).to(...).in(...)\`. Optionally, you may see this because dates are incorrectly formatted, not in milliseconds. Your query was not modified.`
       );
       return this;
     }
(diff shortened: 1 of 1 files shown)


jgravois:
i dig this.

if you wanted to update the JSDoc for `SearchQueryBuilder` to show how to pass through a date _correctly_, that'd be a very welcome addition too!

https://github.com/Esri/arcgis-rest-js/blob/8d18794ec30f346b3af1269207c1953dd475bc9f/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts#L2-L32

phpmaps:
@jgravois  done.  w/ doc.

packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts
```diff
@@ -3,10 +3,16 @@ import { IParamBuilder, warn } from "@esri/arcgis-rest-request";
  * `SearchQueryBuilder` can be used to construct the `q` param for [`searchItems`](/arcgis-rest-js/api/portal/searchItems#searchItems-search) or [`searchGroups`](/arcgis-rest-js/api/portal/searchGroups#searchGroups-search). By chaining methods, it helps build complex search queries.
  *
  * ```js
+ * const startDate = new Date("2020-01-01").valueOf();
+ * const endDate = new Date(2020-09-01).valueOf();
```
jgravois, line 7:
```suggestion
 * const startDate = new Date("2020-01-01");
 * const endDate = new Date(2020-09-01);
```
based on the code below, i was under the impression that users could pass a plain old instance of `Date()`.
is that not the case?
https://github.com/Esri/arcgis-rest-js/blob/8d18794ec30f346b3af1269207c1953dd475bc9f/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts#L308-L310

phpmaps: @jgravois dunno.  i can circle back after trying plain old dates.  i do know Date().valueOf() works.

jgravois: yup. it definitely works and i'd recommend that we document it that way, because its more straightforward.

https://codepen.io/jgravois/pen/WNwZeez

phpmaps:
Thanks @jgravois.  How about this update, now? I fixed a few more things.  Also double confirmed that plain old dates do work.  Kinda makes sense that an API would use real Date() objects.  I guess having come from straight REST calls via HTTP I missed this.  

I do know having flubbed dates with the Query Builder recently - that these errors have nothing to do with the actual problem which could be Date related:

https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts#L212

https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-portal/src/util/SearchQueryBuilder.ts#L234

phpmaps:
@jgravois - yeah, makes sense.  good idea. i can revise.

tomwayson (approved):
Thanks @phpmaps and sorry for the delay!

PR #748, opened 2020-09-03
v2.17.0

rweber-esri:
Release commit, couldn't push directly to master post release.

### DIFF
--- a/packages/arcgis-rest-auth/package.json
+++ b/packages/arcgis-rest-auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@esri/arcgis-rest-auth",
-  "version": "2.16.0",
+  "version": "2.17.0",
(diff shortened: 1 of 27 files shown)


packages/arcgis-rest-auth/package.json
```diff
@@ -13,11 +13,11 @@
     "dist/**"
   ],
   "dependencies": {
-    "@esri/arcgis-rest-types": "^2.16.0",
+    "@esri/arcgis-rest-types": "^2.17.0",
```
tomwayson, line 16: why is this a dep and not a devDep?

tomwayson (approved):

packages/arcgis-rest-auth/package.json
rweber-esri: Nice catch, it was a dep prior to the 2.17.0 release. I'll submit a follow-up PR to move.

PR #755, opened 2020-09-10
F/get item info

tomwayson:


rweber-esri:
> @rweber-esri
> 
> So far `getItemInfo()` only supports text (XML) files, but it looks like it also would need to at least support JSON files for surveys (i.e. `${formName}.json`). I can add that. Do you know what format the `${fomName}.info` files are? Will `response.text()` work for those?

@tomwayson 

Looks like the contents of of the `.info` files are JSON. An easy way to get at these files is to view the edit page in EAPS for a `Solution` created from a survey. Check out [this Solution in EAPS](https://eapsqaext.surge.sh/items/7d06e0b390f547518ec27781c2411bb0/edit) as `paige_pa`

rweber-esri (approved):
Looks good to me

PR #758, opened 2020-09-10
getItemInfo can return the raw response as an escape hatch

### LINKED ISSUE #757: getItemInfo() should handle more than just text files
Surveys for example could have JSON files, images, excel spreadsheets(!). For example, [this survey](https://qaext.arcgis.com/sharing/rest/content/items/7b1a434793df480d84680e8581a56eb7?f=json) (on QA) seems to have all of these files under info:

image

It won't be easy to rely on file extension either (see https://github.com/Esri/arcgis-rest-js/pull/755#issuecomment-690689168). 

Maybe instead we could base how the response is parsed on the `f` parameter, which could be any of these values:

https://github.com/Esri/arcgis-rest-js/blob/2092bb31dd6a1429f94580a51df30e6f16b370e8/packages/arcgis-rest-request/src/utils/ResponseFormats.ts#L4-L10

This is what `request()` does, but it defaults to `.blob()`. I think we want the default to be `.text()` since `iteminfo.xml` is the default file.

tomwayson:
resolves #757

### DIFF
--- a/packages/arcgis-rest-portal/test/items/get.test.ts
+++ b/packages/arcgis-rest-portal/test/items/get.test.ts
@@ -22,7 +22,8 @@ import {
   ItemGroupResponse,
   RelatedItemsResponse,
   ItemInfoResponse,
-  ItemMetadataResponse
+  ItemMetadataResponse,
+  ItemFormJsonResponse
 } from "../mocks/items/item";
 
 import { GetItemResourcesResponse } from "../mocks/items/resources";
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-portal/test/items/get.test.ts
```diff
@@ -191,6 +192,29 @@ describe("get", () => {
...
 
+  it("should return raw response for non-text item info files", done => {
+    fetchMock.once("*", ItemFormJsonResponse);
+
+    getItemInfo("3ef", {
+      fileName: "form.json",
+      rawResponse: true
+    } as IGetItemInfoOptions)
+      .then(response => response.json())
+      .then(formJson => {
```
tomwayson, line 203: @rweber-esri this shows how to fetch a JSON info file. A little verbose, but enough of a stop gap to cut a release for now.

rweber-esri (approved):

packages/arcgis-rest-portal/test/items/get.test.ts
rweber-esri: > @rweber-esri this shows how to fetch a JSON info file. A little verbose, but enough of a stop gap to cut a release for now.

Passing an argument for the _expected_ output format and having `getItemInfo` resolve the result of `json()`, `text()` (already doing), `blob()` accordingly feels like a less jarring update for consuming apps to make down the road if/when we add more elegant content type detection.

```
getItemInfo('3ef', { fileName: 'form.json', type: 'json' })
```

Once content type detection is added, we could just deprecate the `type` prop, consuming apps wouldn't _need_ to remove that param. The current approach, the consuming apps would need to remove the call to `.json()`. Not sure if this tweak would have implications on minor vs major down the road?

tomwayson: I actually started down that road (using `readAs: 'json'), but was unsure if I we'd end up needing to support `arrayBuffer()` and `formData()`. Then I thought, we'd could just use `rawResonse` as an escape hatch for those edge cases. Then I realized that the original implementation broke `rawResponse`, so I thought I'd start there and we could add the `readAs` param as an enhancement later.

Although I have no plan to do content type detection at this point (so many edge cases), I do think you make a good point about refactoring consumer code, so I'll add that now.

tomwayson:
OK, here's the final signature, I think this is GTG (as long as you're OK w/ `readAs` as the option name):

image

PR #762, opened 2020-09-15
FEAT: Add postMessage auth wrappers

dbouwman:
Adds static and instance methods to UserSession to support `postMessage` based authentication.

Adds a Guide to docs re: how to work with this

Note: The message types and payloads have been reviewed in internal meetings

### DIFF
--- a/docs/src/guides/embedded-apps.md
+++ b/docs/src/guides/embedded-apps.md
@@ -1,0 +1,95 @@
+---
+title: Passing Authentication to iFramed apps
+navTitle: Embedded Authentication
+description: Learn how to pass authentication into iframe embedded applications.
+order: 50
+group: 2-authentication
+---
+
+**Note** This is a new api feature, and currently there are *no* platform applications that support the authentication flows discussed in this guide. However, this guide exists to help other platform app teams add this functionality to their apps.
+
+# Authentication with Embedded Applications
+
+Sometimes an application will need to embed another application using an `<iframe>`. If both applications are backed by items that are publicly accessible, things will just work.
+
+But, if the "Host" application is not public and the embedded application is not public, we then run into the question of how to pass authentication from the "Host" to the embedded application.
+
+### Cross Origin Embedding
+Cross-Origin embedding occurs when the "host" app and the "embedded" application are served from different locations. This is only supported for ArcGIS Platform apps that support embedding.
+
+
+For example, you can build a custom app, hosted at `http://myapp.com` and iframe in a "platform app" that supports embedding. However, you can not embed your custom app into a storymap, and expect the storymap to pass authentication to your app. This is done for security reasons.
+
+
+## Using `postMessage` 
+The browser's `postMessage` api is designed to allow communication between various "frames" in a page, and it is how this works internally. You can read more about [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) at the Mozilla Developer Network.
+
+# Authentication Flow
+We will walk through the flows at a high-level
+
+## 1 Host App uses oAuth
+
+The application acting as the host, should use oAuth to autentication the user, and *before* rendering the iframe with the embedded application it must call `session.enablePostMessageAuth(validOrigins)`. This sets up an event handler that will process the requests from the embedded applications.
+
+The `validOrigins` argument is an array of "orgins" your app expects to get auth requests from. **NOTE** This should be a constrained list of just the domains this particular application will actually be embedding.
+
+```js
+  // register your own app to create a unique clientId
+  const clientId = "abc123"
+
+  UserSession.beginOAuth2({
+    clientId,
+    redirectUri: 'https://yourapp.com/authenticate.html'
+  })
+  .then(session => {
+    // for this example we will only send auth to embeds hosted on storymaps.arcgis.com
+    const validOrigins = ['https://storymaps.arcgis.com'];
+    session.enablePostMessageAuth(validOrigins);
+  })
+```
+
+#### 2  Host App adds params to embed url
+Let's suppose the host app is embedding `https://storymaps.arcgis.com/stories/15a9b9991fff47ad84f4618a28b01afd`. To tell the embedded app that it should request authentication from the parent we need to add two url parameters:
+
+- `embed=iframe` - tells the app it's embedded in an iframe. This allows the app to make ui changes like hiding headers etc
+- `parentOrigin=https://myapp.com` - this tells the app what 'origin' to expect messages from, and also to ignore other origins. **note** this should be uri encoded
+
+```js
+const originalUrl = 'https://storymaps.arcgis.com/stories/15a9b9991fff47ad84f4618a28b01afd';
+const embedUrl = `${originalurl}?embed=true&parentOrigin=${encodeURIComponent(window.location.origin)}`;
+// then use embedUrl in your component that renders the <iframe>
+```
+
+#### 3 Embed App boots and Requests Auth
+In the embedded application, early in it's boot sequence it should read the query string parameters and make the determintation that it is running inside an iframe, and that it can request authentication information from the parent. 
+
+```js
+  // Parse up any url params
+  let params = new URLSearchParams(document.location.search.substring(1));
+  const embedStyle = params.get('embed');
+  const parentOrigin = params.get('parentOrigin'); 
+  if (embedStyle === 'iframe' && parentOrigin) {
+    UserSession.fromParent(parentOrigin)
+    .then((session) => {
+      // session is a UserSession instance, populated from the parent app
+      // the embeded app should exchange this token for one specific to the application
+    })
+    .catch((ex) => {
+      // if the origin of the embedded app is not in the parent's validOrigin array
+      // this will throw with a message "Rejected authentication request."
+    })
+  }
+```
+
+#### 4 Parent App Transitions Away
+If the parent app has the ability to transition to another route (i.e. an Angular, Ember, React etc app with a router) then **before** the transition occurs, the postMessage event listener must be disposed of. Unfortuately we can't automate this inside ArcGIS Rest Js as it will require using the life-cycle hooks of your application framework. The example below uses the `disconnectedCallback` that is part of the Stenci.js component life-cycle.
+
+```js
+  // Use your framework's hooks...
+  disconnectedCallback () {
+    // stop listening for requests from children
+    state.session.disablePostMessageAuth();
+  }
+```
+
+
--- a/docs/src/guides/embedded-apps.md
+++ b/docs/src/guides/embedded-apps.md
@@ -68,7 +68,7 @@ In the embedded application, early in it's boot sequence it should read the quer
   let params = new URLSearchParams(document.location.search.substring(1));
   const useEmbedAuth = params.get('arcgis-auth-embed');
   const parentOrigin = params.get('parentOrigin'); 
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -438,6 +438,66 @@ export class UserSession implements IAuthenticationManager {
...
+      handler = (event:any) => {
+        // ensure we only listen to events from the specified parent
+        if (event.origin === parentOrigin){
+          try {
+            return resolve(UserSession.parentMessageHandler(event));
+          } catch (err) {
+            return reject(err);
+          }
+        } else {
+          return reject(new Error('Rejected authentication request.'));
```
dasa, line 473: Should it do anything here? The app may be getting other valid messages.

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,95 @@
...
+    // for this example we will only send auth to embeds hosted on storymaps.arcgis.com
+    const validOrigins = ['https://storymaps.arcgis.com'];
+    session.enablePostMessageAuth(validOrigins);
+  })
+```
+
+#### 2  Host App adds params to embed url
+Let's suppose the host app is embedding `https://storymaps.arcgis.com/stories/15a9b9991fff47ad84f4618a28b01afd`. To tell the embedded app that it should request authentication from the parent we need to add two url parameters:
+
+- `embed=iframe` - tells the app it's embedded in an iframe. This allows the app to make ui changes like hiding headers etc
```
dasa, line 54: This parameter isn't needed since the app can detect this using `window !== window.parent`.

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,95 @@
...
+- `parentOrigin=https://myapp.com` - this tells the app what 'origin' to expect messages from, and also to ignore other origins. **note** this should be uri encoded
```
dasa, line 55:
```suggestion
- `parentOrigin=https://myapp.com` - this tells the app what 'origin' to expect messages from, what origin to post messages to, and also to ignore other origins. **note** this should be uri encoded
```

dasa (changes requested):

docs/src/guides/embedded-apps.md
dasa: If we want to keep this param as a kind of signal telling the app that it should use `postMessage`, we should probably name it to be less generic in case the app has its own `embed` parameter. Maybe "arcgis:auth:embed=true"?

packages/arcgis-rest-auth/src/UserSession.ts
dbouwman: @dasa good call. I'll just drop the else.

docs/src/guides/embedded-apps.md
dbouwman: @dasa I'll change it to your suggestion

packages/arcgis-rest-auth/src/UserSession.ts
dbouwman: actually - since we return a Promise, we have to either resolve or reject. I'm going to resolve(null) for now and we can change if/when we operationally hit an issue

dasa: Wouldn't that make is resolve too early? It should still resolve eventually when it get's the right message.

dbouwman: @dasa Nice catch! I added a test which fires two `message`, the first one from another origin (which we ignore) and the second, from the parent origin, which we process.

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### 3 Embed App boots and Requests Auth
+In the embedded application, early in it's boot sequence it should read the query string parameters and make the determintation that it is running inside an iframe, and that it can request authentication information from the parent. 
+
+```js
+  // Parse up any url params
+  let params = new URLSearchParams(document.location.search.substring(1));
+  const embedStyle = params.get('embed');
+  const parentOrigin = params.get('parentOrigin'); 
+  if (embedStyle === 'iframe' && parentOrigin) {
```
dasa, line 71: Should be updated to use look for "arcgis-auth-embed=true"

dbouwman: gah! thanks for the close reading!

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### 3 Embed App boots and Requests Auth
+In the embedded application, early in it's boot sequence it should read the query string parameters and make the determintation that it is running inside an iframe, and that it can request authentication information from the parent. 
+
+```js
+  // Parse up any url params
+  let params = new URLSearchParams(document.location.search.substring(1));
+  const useEmbedAuth = params.get('arcgis-auth-embed');
+  const parentOrigin = params.get('parentOrigin'); 
+  if (useEmbedAuth && parentOrigin) {
```
dasa, line 71:
```suggestion
  if (useEmbedAuth === "true" && parentOrigin) {
```

dasa (approved):
Just one small suggestion and 👍🏻

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,93 @@
...
+group: 2-authentication
+---
+
+**Note** This is a new api feature, and currently there are *no* platform applications that support the authentication flows discussed in this guide. However, this guide exists to help other platform app teams add this functionality to their apps.
+
+# Authentication with Embedded Applications
+
+Sometimes an application will need to embed another application using an `<iframe>`. If both applications are backed by items that are publicly accessible, things will just work.
+
+But, if the "Host" application is not public and the embedded application is not public, we then run into the question of how to pass authentication from the "Host" to the embedded application.
```
tomwayson, line 15:
```suggestion
However, if the embedded application is not public and the user has already logged into the "Host" application, we then run into the question of how to pass authentication from the "Host" to the embedded application.
```

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Using `postMessage` 
+The browser's `postMessage` api is designed to allow communication between various "frames" in a page, and it is how this works internally. You can read more about [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) at the Mozilla Developer Network.
+
+# Authentication Flow
+We will walk through the flows at a high-level
+
+## 1 Host App uses oAuth
+
+The application acting as the host, should use oAuth to authenticate the user, and *before* rendering the iframe with the embedded application it must call `session.enablePostMessageAuth(validOrigins)`. This sets up an event handler that will process the requests from the embedded applications.
```
tomwayson, line 32:
```suggestion
The application acting as the host, should use oAuth to authenticate the user, and *before* rendering the iframe with the embedded application it must call `session.enablePostMessageAuth(validOrigins)`. This sets up a listener that will process the requests from the embedded application(s).
```

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,93 @@
...
+    const validOrigins = ['https://storymaps.arcgis.com'];
+    session.enablePostMessageAuth(validOrigins);
+  })
+```
+
+#### 2  Host App adds params to embed url
+Let's suppose the host app is embedding `https://storymaps.arcgis.com/stories/15a9b9991fff47ad84f4618a28b01afd`. To tell the embedded app that it should request authentication from the parent we need to add two url parameters:
+
+- `arcgis-auth-embed=true` - tells the app it's embedded in an iframe and should request auth from the parent. This allows the app to make ui changes like hiding headers etc
+- `parentOrigin=https://myapp.com` - this tells the app what 'origin' to expect messages from, what origin to post messages to, and also to ignore other origins. **note** this should be uri encoded
```
tomwayson, line 55: These should use consistent casing; both should use either camel or kebab. Also, should the be consistent about using an `arcgis` prefix?

@dasa, since they both appear to be required, can we instead get away w/ a single `arcgis-auth-origin=myapp.com`? If that's present assume `arcgis-auth-embed=true`?

docs/src/guides/embedded-apps.md
```diff
@@ -0,0 +1,93 @@
...
+      // the embeded app should exchange this token for one specific to the application
+    })
+    .catch((ex) => {
+      // if the origin of the embedded app is not in the parent's validOrigin array
+      // this will throw with a message "Rejected authentication request."
+    })
+  }
+```
+
+#### 4 Parent App Transitions Away
```
tomwayson, line 84: It seems like transitioning away is a specific reason to clean up and we may want to make this section more generic like:

```suggestion
`### Clean up`
```

For example, if the host app has a sign out button, does this need to be called there?

tomwayson: hm... looks like from below that once the authentication has been successfully passed from the host to the embedded app that the listener is removed. I think we should mention that here, and call out the scenarios where you might need to do it manually. I still think the transition example can be made more generic, like, "If the host app is no longer going to render the iframe before the embedded app has a chance to request the authentication, then it should manually clean up the listener by calling:`

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -809,6 +871,36 @@ export class UserSession implements IAuthenticationManager {
...
+   * Internally this adds an event listener on window for the `message` event
+   * 
+   * @param validChildOrigins Array of origins that are allowed to request authentication from the host app
+   */
+  public enablePostMessageAuth (validChildOrigins: string[], win?:any ): any {
+    /* istanbul ignore next: must pass in a mockwindow for tests so we can't cover the other branch */
+    if (!win && window) {
+      win = window;
+    }
+    this.hostHandler = this.createPostMessageHandler(validChildOrigins);
```
tomwayson, line 888: should we name this `_hostHandler`?

tomwayson (commented):
Great work @dbouwman - I think my main hope would be that we could use a single query param.

docs/src/guides/embedded-apps.md
dasa: That sounds good to me 👍🏻

docs/src/guides/embedded-apps.md
```diff
@@ -51,12 +51,11 @@ The `validOrigins` argument is an array of "orgins" your app expects to get auth
 #### 2  Host App adds params to embed url
 Let's suppose the host app is embedding `https://storymaps.arcgis.com/stories/15a9b9991fff47ad84f4618a28b01afd`. To tell the embedded app that it should request authentication from the parent we need to add two url parameters:
 
-- `arcgis-auth-embed=true` - tells the app it's embedded in an iframe and should request auth from the parent. This allows the app to make ui changes like hiding headers etc
-- `parentOrigin=https://myapp.com` - this tells the app what 'origin' to expect messages from, what origin to post messages to, and also to ignore other origins. **note** this should be uri encoded
+- `arcgis-auth-origin=https://myapp.com` - this tells the app what 'origin' to expect messages from, what origin to post messages to, and also to ignore other origins. **note** this should be uri encoded
```
dasa, line 54: This sentence from the old param may be good to copy over too:

> tells the app it's embedded in an iframe and should request auth from the parent.

dbouwman: Cool - I'll address w/ Tom's "Clean Up" section comment in a moment...

PR #764, opened 2020-09-18
getItemResource and scrubControlChars functions

drewctate:
…scrubControlChars

AFFECTS PACKAGES:
@esri/arcgis-rest-portal

### DIFF
--- a/packages/arcgis-rest-portal/src/items/get.ts
+++ b/packages/arcgis-rest-portal/src/items/get.ts
@@ -216,6 +216,9 @@ export interface IGetItemResourceOptions extends IRequestOptions {
  * ```
  * Fetches an item resource and optionally parses it to the correct format.
  *
+ * Note: provides JSON parse error protection by sanitizing out any unescaped control
+ * characters before parsing that would otherwise cause an error to be thrown
+ *
  * @param {string} itemId
  * @param {IGetItemResourceOptions} requestOptions
  */
@@ -404,7 +407,7 @@ export function getItemMetadata(
 
 // overrides request()'s default behavior for reading the response
 // which is based on `params.f` and defaults to JSON
-// Also add JSON parse error protection by sanitizing out any unescaped control characters before parsing
+// Also adds JSON parse error protection by sanitizing out any unescaped control characters before parsing
 function getItemFile(
   id: string,
   // NOTE: fileName should include any folder/subfolders
--- a/packages/arcgis-rest-portal/src/items/get.ts
+++ b/packages/arcgis-rest-portal/src/items/get.ts
@@ -404,6 +404,7 @@ export function getItemMetadata(
 
 // overrides request()'s default behavior for reading the response
 // which is based on `params.f` and defaults to JSON
+// Also add JSON parse error protection by sanitizing out any unescaped control characters before parsing
 function getItemFile(
   id: string,
   // NOTE: fileName should include any folder/subfolders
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-portal/src/util/scrub-control-chars.ts
```diff
@@ -0,0 +1,12 @@
+const CONTROL_CHAR_MATCHER = /[\x00-\x1F\x7F-\x9F\xA0]/g;
+
+/**
+ * Returns a string with all control characters removed.
+ *
+ * Doesn't remove characters
```
tomwayson, line 6:
```suggestion
 * Returns a new string with all control characters removed.
 *
 * Doesn't remove characters from input string.
```

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -185,6 +186,31 @@ export interface IGetItemGroupsResponse {
...
+  const url = `${getPortalUrl(
+    requestOptions
+  )}/content/items/${itemId}/resources/${resourceName}`;
+
+  // We need the raw response because there may be control characters
+  // that need to be scrubbed prior to parsing the JSON
+  const options = Object.assign({ params: { f: "json" }, rawResponse: true }, requestOptions);
+  return request(url, options)
+    .then(res => res.text())
+    .then(text => JSON.parse(scrubControlChars(text)));
```
tomwayson, line 210: why is `scrubControlChars()` needed here?

tomwayson (changes requested):
Nice work, but I think it would be better if rather than `getJsonResource()` we should implement `getItemResource()` as I propose here: 

https://github.com/Esri/arcgis-rest-js/pull/701#issuecomment-691164727

If there's some reason that we need to call `scrubControlChars()` every time a JSON resource is fetched, I'm sure you could work that into the above.

packages/arcgis-rest-portal/src/items/get.ts
drewctate: JSON resources in the wild occasionally come back with unescaped control characters which breaks `res.json()`. Still trying to track down why/how those make it in there, but this is a backstop to make sure that we at least get the JSON back without control chars instead of a total failure. We think this is fine because control characters are usually added inadvertently by copy-pasting from MS word, etc.

drewctate: cc: @dbouwman

tomwayson: that's great, would be good to have that in comments

drewctate:
Not sure how to deal with the failing test. It's angry about `response.blob` in NodeJS. I know that newer versions of `node-fetch` include this support.

tomwayson:
You have to skip that test in node by checking for `window`.

tomwayson:
or, I guess if newer versions really do support blobs, then maybe install one of those as a dev dependency and explicitly pass that to this test.

drewctate:
> See how the other tests for request() handle blobs.

@tomwayson Are you talking about [this](https://github.com/Esri/arcgis-rest-js/blob/d75fe7c35f7ed2346171e60662021c759e8ceed1/packages/arcgis-rest-portal/test/items/get.test.ts#L86-L87)?

tomwayson (approved):
yep, that's what I was talking about

PR #772, opened 2020-10-23
Update getTokenForServer to pass request options to generateToken 

charlietilt:
## Issue
When passing your own `fetch` reference in the request options, those options are not being passed into the `generateToken` call in the `getTokenForServer` method.

## Summary
- updated `getTokenForServer` to pass `requestOptions` into `generateToken` call to a


tomwayson:
Seems reasonable to me, but I'm curious what others think.

Also, I notice that `getTokenForServer()` makes other requests w/o passing along `requestOptions` like:

https://github.com/Esri/arcgis-rest-js/blob/08208d3780cb9daaa0024388c13864f75f527e72/packages/arcgis-rest-auth/src/UserSession.ts#L1018

tomwayson:
For now we think it's best to _only_ pass in the custom fetch rather than all of the `requestOptions` - that should fix your issue w/o potentially introducing breaking changes.

gavinr:
Closing due to inactivity. Please feel free to re-submit following Tom's suggestion.

PR #775, opened 2020-11-06
Make "layers" parameter of IExportParameters optional

rgwozdz:
We want the option of not setting `layers` in IExportParameters. That will force the /export service to include all the layers in the delivered export.


haoliangyu (approved):

tomwayson:
But this says `layers` is required: https://developers.arcgis.com/rest/users-groups-and-items/export-item.htm - in fact, that appears to be the only parameter on that page that's called out as required?

rgwozdz:
@tomwayson - it does say that, but in fact, it is not true. I've tested it and can send you the curl if you like.  There are cases where you want to export all layers but still set a `targetSR`.  In such cases, you would leave out the layers parameter.

tomwayson (approved):
only suggestion would be to put this into a code comment:

https://github.com/Esri/arcgis-rest-js/pull/775#issuecomment-724034959

rgwozdz:
Confirmed that the docs are out of date;  will add code-comment to note the discrepancy.

dbouwman:
@rgwozdz fyi - CI fails on node 15.1.0 due to https://github.com/npm/cli/issues/2084
My PR will remove latest node from running in CI until that issue is resolved... that should merge later today...

rgwozdz:
@dbouwman - would it help if I do a separate PR for removing latest Node from Travis?

PR #777, opened 2020-11-10
Allow custom encoders in `request`

MikeTschudi:
Proposed as an alternate approach to PR #686 to minimize impact.  Modeled after the way that `request` supports a custom `fetch` implementation.

### DIFF
--- a/packages/arcgis-rest-request/src/utils/IRequestOptions.ts
+++ b/packages/arcgis-rest-request/src/utils/IRequestOptions.ts
@@ -37,6 +37,14 @@ export interface IRequestOptions {
    */
   fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
   /**
+   * The implementation for converting parametrs into a query string. Defaults to `encodeQueryString` in this package.
+   */
+  encodeQueryString?: (params: any) => string;
+  /**
+   * The implementation for converting parametrs into a FormData. Defaults to `encodeFormData` in this package.
+   */
+  encodeFormData?: (params: any, forceFormData?: boolean) => FormData | string;
+  /**
    * If the length of a GET request's URL exceeds `maxUrlLength` the request will use POST instead.
    */
   maxUrlLength?: number;
--- a/packages/arcgis-rest-request/src/utils/IRequestOptions.ts
+++ b/packages/arcgis-rest-request/src/utils/IRequestOptions.ts
@@ -39,11 +39,11 @@ export interface IRequestOptions {
   /**
    * The implementation for converting parametrs into a query string. Defaults to `encodeQueryString` in this package.
    */
-  encodeQueryString?: (params: any) => string;
+  encodeQueryString?: (params: IParams) => string;
   /**
    * The implementation for converting parametrs into a FormData. Defaults to `encodeFormData` in this package.
    */
-  encodeFormData?: (params: any, forceFormData?: boolean) => FormData | string;
+  encodeFormData?: (params: IParams, forceFormData?: boolean) => FormData | string;
   /**
    * If the length of a GET request's URL exceeds `maxUrlLength` the request will use POST instead.
    */
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,10 @@ notifications:
   slack:
     secure: KghYpsaoCZB1sPOqk1d+JhfBOkg76Cq8eMK665ozdf484RdPvObqLrxyrt9i+sDMDv/oZT+VuvGXsbHj6R+TUh87r2Yhfizcwvdotc4VTHei4y/oNX6cRhV8ryQox/0j6GClqI43A4kIsuOcUm8pI4FTGOhDQOK7+UDcuvcSa5VR7N7GJ97zfzC9gQtRsbfD5WKY4BEr+KlcWt0KCCbp1+Jh7craEO30ztUl38BuUuW7F1/AykrTX6kgS5E5ssUQ0pJPBlzpjb3G5UEeodzOD9wUBVLIBAbgQsofAPLi0pHeyCCeEOuTxXEo7SD8fG+kBstPlncR+V8goCEBnAJysIvBznUxLjrnxcU7V40g+iTENKwhVTIiZzZN3Z9buJd001MRaWjOTCWX80Ig6oraH8WJk7W1qtKzsDK9Lsp2N9pxNTPASbAQoFDlMCGcWYGixq6IhIAwsk1Pkb689F67vHaES8qMXkIHifB0j7yRMwrHinYFqUYASQOECPNEMJv3iqByXntmMP8X/r3VBujMehGodZ1dnbWvAOdWdEmG3KCVpn42ocKCj6k4WZL62nSxs3W00E1UVPnXujT3ENk/3L/VJE3tRWFCSbR9HnfBSDVA6WaFu4EFDg8WV6fn2T0FHKTBgCtTugQo8bsoyReYZvEU/27O/B+d5/9nqMHqyy4=
 node_js:
-  - "node"
+  # Currently `npm ci` fails when installing a GH repo dep that has a prepare script that uses a command from a devDependency
+  # https://github.com/npm/cli/issues/2084 is the issue, it's closed but node has not cut a release yet 11/9/2020
+  # Will return this once it's resolved
+  # - "node"
   - "lts/*"
   - 10
 cache: npm
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -241,6 +241,14 @@ export function request(
     );
   }
 
+  // Check for custom encoders
+  if (!options.encodeQueryString) {
+    options.encodeQueryString = encodeQueryString;
+  }
+  if (!options.encodeFormData) {
+    options.encodeFormData = encodeFormData;
+  }
+
   const { httpMethod, authentication, rawResponse } = options;
 
   const params: IParams = {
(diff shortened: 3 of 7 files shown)


MikeTschudi:
In TravisCI, passes "Node.js: lts/*" and "Node.js: 10"; fails "Node.js: node" with a lerna problem:
```
$ npm ci 
npm ERR! code 127
npm ERR! git dep preparation failed
npm ERR! command /home/travis/.nvm/versions/node/v15.1.0/bin/node /home/travis/.nvm/versions/node/v15.1.0/lib/node_modules/npm/bin/npm-cli.js install --only=dev --prod --ignore-prepublish --no-progress --no-save --cache=/home/travis/.npm/_cacache --prefer-offline=false --prefer-online=false --offline=false --before=
npm ERR! > cz-lerna-changelog@1.2.1 prepublish
    :             :             :
```

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -36,6 +36,14 @@ export interface IRequestOptions {
...
    */
   fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
+  /**
+   * The implementation for converting parametrs into a query string. Defaults to `encodeQueryString` in this package.
+   */
+  encodeQueryString?: (params: any) => string;
+  /**
+   * The implementation for converting parametrs into a FormData. Defaults to `encodeFormData` in this package.
+   */
+  encodeFormData?: (params: any, forceFormData?: boolean) => FormData | string;
```
tomwayson, line 46:
```suggestion
  encodeFormData?: (params: IParams, forceFormData?: boolean) => FormData | string;
```

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -36,6 +36,14 @@ export interface IRequestOptions {
    * The implementation of `fetch` to use. Defaults to a global `fetch`.
    */
   fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
+  /**
+   * The implementation for converting parametrs into a query string. Defaults to `encodeQueryString` in this package.
+   */
+  encodeQueryString?: (params: any) => string;
```
tomwayson, line 42:
```suggestion
  encodeQueryString?: (params: IParams) => string;
```

tomwayson (commented):
Nice work @MikeTschudi!

This approach seems like like a simpler (in terms of implementation) to achieve what @patrickarlt [proposed previously](https://github.com/Esri/arcgis-rest-js/pull/686#pullrequestreview-404273419).

The only thing I don't like is that this introduces 2 new request options, and there's already a lot. Furthermore they're functions that people would have to understand and we would have to support.

Ultimately I'd prefer Pat's API, but it could get messy inside `request()`, for example in dealing w/ the `hideToken` logic.

I'm wondering if this seemingly specific case of categories is actually an instance of more generic case. Maybe we could have an option that would indicate that we should encode arrays as multiple query/form parameters? Or maybe instead of a boolean option, the option could take a list of param names that we should encode that way. Then search could pass in something like `encodeMultiple: ['categories']`.

MikeTschudi:
Thanks for the comments, Tom. I tried to implement Pat's API, but `request` did indeed get messy as the parameters wove around the token handling.

The two new functions are simply the signatures of the existing `encodeFormData` (encode-form-data.ts) and `encodeQueryString` (encode-query-string.ts), which `request` would continue to call by default. I think that the only maintenance necessary in arcgis-rest-js is keeping the signatures in `IRequestOptions` sync'ed with the existing functions. (I'll submit a PR update changing `params: any` to `params: IParams` in the existing functions and the interface.)

An advantage of permitting custom encoding functions is that it supports what Pat wrote in his comment: "But this does provide for future cases where methods might need to implement their own special encoding methods."

dbouwman:
@MikeTschudi  while this is an elegant solution, I'm leery about passing encoding functions into `request`. It means we have to add more logic to correctly trap failures within the encoder function and return that in a useful way to the developer. 

While code in rest-js requires 100% coverage, and is likely pretty bomber, opening up these params allows any developer to pass in any function... 

So - going back to Pat's original suggestion -  would it simplify things if we state that if you pass in `params.encodedFormData` or `params.encodedQueryString`, you as the developer take full responsibility for headers/method/token etc etc?

MikeTschudi:
@dbouwman, Pat pointed out how complicated `request` is at this point, and so passing in `params.encodedFormData` or `params.encodedQueryString` has the risk of introducing destabilizing changes. E.g., at the very beginning of `request`, we have
```ts
const options: IRequestOptions = {
  ...{ httpMethod: "POST" },
  ...DEFAULT_ARCGIS_REQUEST_OPTIONS,
  ...requestOptions,
  ...{
    params: {
      ...DEFAULT_ARCGIS_REQUEST_OPTIONS.params,
      ...requestOptions.params
    },
    headers: {
      ...DEFAULT_ARCGIS_REQUEST_OPTIONS.headers,
      ...requestOptions.headers
    }
  }
};
```
Changing params' signature ripples through the system. We probably don't want to change `IParams`, so that means changing `params?: IParams;` in `IRequestOptions` to a mixed type. And the spread operator would no longer be supported for `params`. We then have to jump around through `request` with an alternate path. It would be safer to have a `request` clone.

While a developer could pass in any function for encoding, if he/she doesn't follow the pattern set in `encodeFormData` and `encodeQueryString`, he/she will get junk results. A developer can stomp on any JavaScript function in arcgis-rest-js, but then what's the point of using the library?

dbouwman:
@MikeTschudi was mulling @tomwayson's point about a "generic" solution to this...

First thought was to use nested arrays...
i.e. `params.categories = [["/Categories/Water","/Categories/Forest",] ["/Region/United States"]]`

and we change the generic encode functions to handle arrays-of-arrays as indicating that we should "encode multiple" entries in the form/query... but... after looking at the encoder implementations maybe we just do this with arrays like your original PR, but dropping the `categories` specific clause... 

`params.categories = ["/Categories/Water, Categories/Forest", "/Region/United States"]`

Simply put, if you pass an array as a param value, we'll send multiple entries for that key.

```js
        if (Array.isArray(newParams[key])) {
          newParams[key].forEach(
            (group: any) => formData.append(key,group)
          );
        } 
```

This does not address the more general, but to-date-not-needed, idea of allowing for custom encoding... but it should address the need for `categories`, and does not require messing w/ `request` itself... we'd have to check that this does not mess things up for the few top-level params that are returned as arrays (item.tags, item.typeKeywords) but I'm pretty sure those have to be sent as comma delimited strings... and thus are encoded before landing at `request` Thoughts?

MikeTschudi:
@dbouwman, handling arrays as param values would certainly work for the categories, and would not require software seeking to search by category to do extra work; I also appreciate not touching `request`. But changing how arrays are handled could have unintended consequences. While we can check item.tags and item.typeKeywords, there might be software out in the world that relies on the current implementation's merging of values for repeat uses of keys. Admittedly a hypothetical, but we can't know for sure.

This proposed PR, on the other hand, puts all of the burden for custom encoding on the developer who chooses to supply custom encoders at the cost of a trivial modification to `request`. There is no effect on current uses of `request` that continue to use the standard encoders.

tomwayson:
I think @dbouwman makes a good point about the idea of allowing for custom encoding has not yet come up, so I think we should take that out of consideration here since supporting that involves opening ourselves up to maintenance headaches.

I like the idea of using a nested array to signal to the encoding functions that they should include this parameter multiple times once for each value.

I agree w/ @MikeTschudi that changing the default way arrays are encoded "could have unintended consequences" but I think that it is unlikely that anyone is relying on the current behavior. Furthermore, the current behavior only really works if your intent is to flatten the nested arrays into a comma separated string, and if those nested arrays contain non-scalar elements, they are serialized as `'[object Object]'`.

I chatted w/ @dbouwman, and we think the best solution is to amend this logic:

https://github.com/Esri/arcgis-rest-js/blob/2d36c52c2149915abd57cc596f58efec1d382784/packages/arcgis-rest-request/src/utils/process-params.ts#L77-L86

To check for `param.length === 1` and `param[0]` to be an array (i.e. a nested array like `[["/Categories/Water", "/Categories/Forest", "/Region/United States"]]`, and if so set `value = param[0]`. That would be the only way that `processParams` would return an array for a given param.

Then the encoding functions could treat any arrays that come out `processParams` the way that @dbouwman [proposes above](https://github.com/Esri/arcgis-rest-js/pull/777#issuecomment-725827736), i.e. the logic that @MikeTschudi introduces in #686 just w/o the `key === "categories"` check.

Even though that is technically a breaking change, I would say we could get away w/ it b/c we didn't really intend the previous behavior to work.

What do you think @MikeTschudi?

tomwayson:
FWIW, the default behavior of HTML checkboxes in a form is to include the parameter multiple times, once for each checked value:

image

Given that, you could argue that should be the default behavior whenever we are encoding an array, especially since consumers could still get the existing behavior themselves by joining or strngifying parameter values before passing them to `request()`. However, as @MikeTschudi points out, that would not only require us to go through the code and make changes for how we handle `item.types`, etc, but it would almost certainly break consuming code that's out in the wild. I just don't see how we could do that other than w/ a major release.

The other problem is that as far as the ArcGIS REST API is concerned, the `categories` param is the exception. In fact, [the update item end point expects a single `categories` as a comma separated string](https://developers.arcgis.com/rest/users-groups-and-items/update-item.htm), so it seems that the API is not even consistent in how it handles `categories` between endpoints (good thing we didn't merge #686).

dbouwman:
@tomwayson I think that's a little off

With the intent being to express the following condition...
```
(categories.includes("/categories/a") OR categories.includes("/categories/b")) AND categories.includes('/categories/c')
```

And an API that needs that communicated as a query/form that looks like (not withstanding the lack or uri encoding)
```
?categories=/categories/a,/categories/b&categories=/categories/c
```

If we do this as a double-nested, single element array, like this...

```
params.categories = [["/categories/a, /categories/b", "/categories/c"]]
```
That means whatever code is constructing the elements in the array, needs to stuff in the AND vs OR ing

Which is doable... but if we wanted, we could make this more array-ish... and expect nested arrays like this

```
params.categories = [["/categories/a", "/categories/b"], ["/categories/c"]]
```

When we talked I was suggesting that if the param is an array, and the first entry in that array is ALSO and array, we return the whole nested geezer

so the output of `processParams` would be `[['/categories/a, /categories/b', '/categories/c']]`

Then encode params can map over things to construct the multiple entries in the FormData or query string

tomwayson:
ah, I didn't realize the desired output is:

```
?categories=/categories/a,/categories/b&categories=/categories/c
```

Your proposal works for me.

MikeTschudi:
Replaced by [Support array of arrays when encoding query strings #779](https://github.com/Esri/arcgis-rest-js/pull/779)

PR #778, opened 2020-11-13
fix: platformSelf to send cookies and append f=json in url

dbouwman:
Update the recently added `platformSelf` function to send cookies as well as `?f=json` in the url. 

The latter should be a short-term fix as the API should not require it, but currently it will 405 on the OPTIONS pre-flight if that's no in the url as shown

image

### DIFF
--- a/packages/arcgis-rest-auth/src/app-tokens.ts
+++ b/packages/arcgis-rest-auth/src/app-tokens.ts
@@ -71,12 +71,19 @@ export function platformSelf(
   redirectUri: string,
   portal: string = "https://www.arcgis.com/sharing/rest"
 ): Promise<IPlatformSelfResponse> {
-  const url = `${portal}/oauth2/platformSelf`;
+  // TEMPORARY: the f=json should not be needed, but currently is
+  const url = `${portal}/oauth2/platformSelf?f=json`;
   const ro = {
     method: "POST",
     headers: {
       "X-Esri-Auth-Client-Id": clientId,
       "X-Esri-Auth-Redirect-Uri": redirectUri,
+    },
+    // We need to ensure the cookies are sent as that's
+    // what the API uses to convert things
+    credentials: "include",
+    params: {
+      f: "json",
     },
   } as IRequestOptions;
   // make the request and return the token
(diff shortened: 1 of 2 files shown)


tomwayson (approved):

noahmulfinger:
@dbouwman @tomwayson  Is this setup working for you? `request` doesn't pass through the `credentials` property as far as I can tell.

tomwayson:
good catch @noahmulfinger. I didn't try it, myself.

Looks like he's right @dbouwman:

https://github.com/Esri/arcgis-rest-js/blob/b978605f08810bbd5f0b568b36afd3f8f2adbdb2/packages/arcgis-rest-request/src/request.ts#L257

dbouwman:
Thanks @noahmulfinger  - the issue is that it's pretty difficult to setup to actually test this fn. I was able to verify that the api works using `fetch` devtools in a tab on a page where the esri_aopc cookie is set, but I have not created an actual test app using rest-js yet. I'll have to puzzle this a bit... I *think* we can use the `credential` value if passed... but I need to verify with the API team how/if/?? should happen in IWA scenarios with this particular function.

packages/arcgis-rest-auth/src/app-tokens.ts
```diff
@@ -71,13 +71,20 @@ export function platformSelf(
...
 ): Promise<IPlatformSelfResponse> {
-  const url = `${portal}/oauth2/platformSelf`;
+  // TEMPORARY: the f=json should not be needed, but currently is
+  const url = `${portal}/oauth2/platformSelf?f=json`;
   const ro = {
     method: "POST",
     headers: {
       "X-Esri-Auth-Client-Id": clientId,
       "X-Esri-Auth-Redirect-Uri": redirectUri,
     },
```
tomwayson, line 81: @dbouwman @noahmulfinger I wonder if we can use the existence of `headers` that start w/ `"X-Esri-Auth-"` as the signal to `request()` that it should set `credentials: "include"`?

noahmulfinger: That seems reasonable. Another option would be adding a `fetchOptions` param or something similar. That way we could allow extra overrides for any future edge cases like this.

PR #779, opened 2020-11-16
Support array of arrays when encoding query strings

MikeTschudi:
Proposed replacement for #777


dbouwman (approved):
Thanks @MikeTschudi!

PR #782, opened 2020-11-22
fix: ensure platformSelf sends cookie

dbouwman:
the `/platformSelf` call is commonly a cross domain request, and needs to send the `esri_aopc` cookie, thus for that specific request we need to ensure that `credential` is set to `include`

Discussion of this solution here: https://github.com/Esri/arcgis-rest-js/pull/778#discussion_r526391636

Also adds verification of the `credential` param in both the `platformSelf` and `request` test suites.

Apparently `prettier` also wanted to format things in `request`... :-/ 

@noahmulfinger / @tomwayson - I can cut a release once CI is green and one of you give this the 👍 

I'd like to get this out so the Urban team can use it next week

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1667,7 +1667,7 @@
     "babylon": {
       "version": "6.18.0",
       "resolved": "https://registry.npmjs.org/babylon/-/babylon-6.18.0.tgz",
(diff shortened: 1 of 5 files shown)


dbouwman:
CI passed... not clear why the status did not update
image

noahmulfinger (approved):
LGTM!

PR #783, opened 2020-12-07
fix: export app-tokens

dbouwman:
Missed exporting `app-token` functions from the package 🙃


btpschroeder (approved):
Tested this locally in Velocity. Imports now work like a charm!

PR #797, opened 2021-02-04
fix(arcgis-rest-auth): postMessage internals

dbouwman:
- Ensure `event.source === window.parent`
- no response unless origin and message type are expected
- update guide and spec

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

### DIFF
--- a/packages/arcgis-rest-auth/post-message-auth-spec.md
+++ b/packages/arcgis-rest-auth/post-message-auth-spec.md
@@ -1,5 +1,12 @@
 # Post Message Authentication Specification
 
+The general idea is that the HOST application will only respond when the following conditions are true:
+
+- the event.source.origin is a location that the HOST trusts
+- the message type is `arcgis:auth:requestCredential`
+
+Under any other conditions, the HOST application will not send any response.
+
 ## Message Types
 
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-auth/post-message-auth-spec.md
```diff
@@ -47,13 +54,17 @@ Message Object
 
 ## `arcgis:auth:rejected`
```
dasa, line 55: arcgis:auth:error

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -635,8 +639,8 @@ export class UserSession implements IAuthenticationManager {
     if (event.data.type === "arcgis:auth:credential") {
       return UserSession.fromCredential(event.data.credential);
     }
-    if (event.data.type === "arcgis:auth:rejected") {
-      throw new Error(event.data.message);
+    if (event.data.type === "arcgis:auth:error") {
+      throw new Error(event.data.error);
```
dasa, line 643: Pass the message to the constructor and then set the error's name property.

dasa (changes requested):

dasa (approved):

PR #805, opened 2021-02-22
Service Name and Service Info Functions

dbouwman:
During work on the Solution.js vNext code, I wrote some functions that should to be in rest-js.

I've broken them into two places:
in arcgis-rest-portal:
- `getUniqueServiceName`
- `isServiceNameAvailable`
- Not sure if these are available on an ArcGIS Server instance, but the key end-point lives at `/portal/<self|id>/isServiceNameAvailable` so I put these in the portal package

in `arcgis-rest-service-admin`
- `getServiceAdminInfo`
- `getViewSources`
I also added some interfaces into the `arcgis-rest-types` package. Those interfaces are very very loose - just enough to make typescript happy for my needs. If someone has cycles, we could expand them

I'm totally open to moving these around, changing signatures etc etc. Just wanted to get this up to start discussions

I also have two more functions to add, but not entirely sure where to add them 
- `projectGeometry` and `findTransformations`

These work w/ any geometry service, so I'm not sure if we should make a geometry package and add more geometry service functions in there, or just slap'em into Portal or.... other ideas.

For now, Solutions vNext will use it's internal versions so no rush on this

### DIFF
--- a/packages/arcgis-rest-portal/src/index.ts
+++ b/packages/arcgis-rest-portal/src/index.ts
@@ -44,10 +44,13 @@ export * from "./sharing/unshare-item-with-group";
 export * from "./sharing/is-item-shared-with-group";
 export * from "./sharing/helpers";
 
+export * from "./services/get-unique-service-name";
+export * from "./services/is-service-name-available";
+
 export * from "./util/get-portal";
 export * from "./util/get-portal-settings";
 export * from "./util/get-portal-url";
-export * from './util/scrub-control-chars';
+export * from "./util/scrub-control-chars";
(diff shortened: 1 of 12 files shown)


gavinr (approved):
looks good to me

tomwayson:
@dbouwman

> These work w/ any geometry service, so I'm not sure if we should make a geometry package and add more geometry service functions in there, or just slap'em into Portal or.... other ideas.

on our call today the consensus was to create an `arcgis-rest-geometry` package for those fns and the README should clarify that it's wrapping geometry engine.

PR #807, opened 2021-02-25
feat(UserSession): make enablePostMessageAuth idempotent, add session…

andrewctate:
….postMessageAuthEnabled flag

AFFECTS PACKAGES:
@esri/arcgis-rest-auth


andrewctate:
Upon further research, I've found that `enablePostMessageAuth` is already idempotent because of the nature of the `addEventListener` function.

andrewctate:
This is cart-before-the-horse. We'll start with an RFC.

Issue #816, opened 2021-03-12
Enh: support ArcGIS Online 9.1 credentials


MikeTschudi:
Credentials in ArcGIS Online &ge;9.1 do not include the `expires` or `ssl` properties. This is a breaking change for arcgis-rest-js that prevents tokens from being fetched or refreshed.

The arcgis-rest-js, the search code calls `getFreshToken`. Even though a token exists inside `UserSession`, the missing expiration date prevents `UserSession` from returning that value, and so `getFreshToken` returns an entry on its `_pendingTokenRequests` while it tries to refresh the token. But the refresh fails because the `refreshToken` inside `UserSession` is not defined:
![image](https://user-images.githubusercontent.com/2125181/110954150-0492d300-82fd-11eb-8f54-b5895b395cb5.png)

A test page for reproducing the problem in Chrome and Firefox is shown below. It offers buttons to sign in and out; after signin, it lists the account name and the number of items owned by that account. (*Note that an appId for your system needs to be supplied on line 33.*)

The page as-is will only show the count of public items because line 45 (`credential.expires = undefined;`) emulates what the ArcGIS Online 9.1 authorization returns as a credential that we supply to arcgis-rest-js' `UserSession.fromCredential`. Comment out line 45 to emulate the ArcGIS Online 8.4 credential, and you will see a count that includes private items.

```html
<!doctype html>
<html lang="en" dir="ltr">
<head>
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
  <script>
    var dojoConfig = {
      "paths": {
        "@esri/arcgis-rest-auth": "https://unpkg.com/@esri/arcgis-rest-auth@3.0.2/dist/umd/auth.umd",
        "@esri/arcgis-rest-portal": "https://unpkg.com/@esri/arcgis-rest-portal@3.0.2/dist/umd/portal.umd",
        "@esri/arcgis-rest-request": "https://unpkg.com/@esri/arcgis-rest-request@3.0.2/dist/umd/request.umd"
      }
    }
  </script>
  <script src="https://js.arcgis.com/4.19/dojo/dojo.js"></script>
</head>
<body>
  <button id="signinBtn" onclick="signinFcn()" disabled>Sign in</button>
  <button id="signoutBtn" onclick="signoutFcn()" disabled>Sign out</button>
  <br/><br/>
  <div id="info"></div>

  <script>
    var signinFcn, signoutFcn;

[...]

PR #817, opened 2021-03-12
Support ArcGIS Online 9.1 credentials

MikeTschudi:
#816

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -52,9 +52,9 @@ export type AuthenticationProvider =
  * object used to access a secure ArcGIS resource.
  */
 export interface ICredential {
-  expires: number;
+  expires?: number; // removed from ArcGIS Online credentials at 9.1
   server: string;
-  ssl: boolean;
+  ssl?: boolean; // removed from ArcGIS Online credentials at 9.119
   token: string;
   userId: string;
 }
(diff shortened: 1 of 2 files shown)


tomwayson (commented):
LGTM, thanks @MikeTschudi!

dbouwman:
For whatever reason... my brain would not let this go... @MikeTschudi  / @tomwayson Could we use a default value if expires is `undefined` in `fromCredential`? Something like 2 hours? 

As this stands, it will require a major bump, and going from `.expires` required (i.e. always has some value) to being optional, (i.e. always requiring a guard prior to use) is not a particularly "clean" sort of break - i.e. even for those consuming rest-js in TypeScript, I'm not sure they'd get any warnings if they did something like `new Date(session.getToken())` which will blow up when their app is iframed into another app... which is the worst place to try to debug things.

MikeTschudi:
I like this variant. And default `ssl` to `true`?

MikeTschudi:
Put in a commit with Dave's idea. Simplifies the PR; what do you think of it, @tomwayson & @dbouwman ?

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -620,14 +620,15 @@ export class UserSession implements IAuthenticationManager {
    * @returns UserSession
    */
   public static fromCredential(credential: ICredential) {
+    const expires = credential.expires || (Date.now() + 7200000 /* 2 hours */);
```
tomwayson, line 623: I think it would be a good idea to add a comment here indicating why we did this and I'd suggest moving the default `ssl` value up here too just to keep it close to the commant.

tomwayson (approved):
This approach works for me, I just would like to document why we did it.

dbouwman:
Thanks @MikeTschudi  & @tomwayson !

Issue #808, opened 2021-02-25
Confusion about layers and tables types in IFeatureServiceDefinition


sandromartis:
I'm struggling to understand the types inside `IFeatureServiceDefinition`.

In  [`IFeatureServiceDefinition`](https://esri.github.io/arcgis-rest-js/api/types/IFeatureServiceDefinition/):
- layers are of type: `ILayerDefinition[] | ILayerDefinition`
- tables are of type: `ITable[] | ITable`

Why are tables not of type `ILayerDefinition[] | ILayerDefinition` as well?

My issue is that `ILayerDefinition` has `id` with type `number`, but `ITable` has `id` with type `string`.

So when I use [`getService`](https://esri.github.io/arcgis-rest-js/api/feature-layer/getService/) to get the feature service definition, I will get quite a mix of types when extracting `layers` and `tables`.

If I call the feature service directly I seem to be getting the same data structure for both layers and tables so I'm not sure why they have different types:
https://developers.arcgis.com/rest/services-reference/feature-service.htm

Could you shed some light on this? Are these types correct? :)

tomwayson:
It looks like `ITable.id` should be a number. It _might_ have been defined as a `string` by mistake. I'd accept a PR that updates the definition.

sandromartis:
Sounds good, I can do a PR for that 👍

PR #819, opened 2021-03-26
Change type of ITable.id to number

sandromartis:
It seems that the type of `ITable.id` should be `number` instead of `string`.

See issue: #808 

AFFECTS PACKAGES:
@esri/arcgis-rest-types

ISSUES CLOSED: #808


tomwayson (approved):
Thanks @sandromartis!

Issue #824, opened 2021-04-12
Cannot update item resource with prefix


thollingshead:
The prefix request option does not get passed through to the item /updateResources request.

PR #825, opened 2021-04-12
fix(updateItemResources): support item resource prefix

thollingshead:
AFFECTS PACKAGES:
@esri/arcgis-rest-portal

ISSUES CLOSED: #824

### DIFF
--- a/packages/arcgis-rest-portal/src/items/update.ts
+++ b/packages/arcgis-rest-portal/src/items/update.ts
@@ -133,6 +133,7 @@ export function updateItemResource(
     requestOptions.params = {
       file: requestOptions.resource,
       fileName: requestOptions.name,
+      resourcesPrefix: requestOptions.prefix,
       text: requestOptions.content,
       ...requestOptions.params
     };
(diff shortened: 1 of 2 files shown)


gavinr:
@thollingshead so it looks like this is adding a parameter `prefix` that maps to the `resourcesPrefix` REST API parameter - any specific reason why we need this wording change? Is it to keep consistent with a previous version or something? Or is it documented that `prefix` shoudl work?

If not, I would generally lean toward keeping the exact same parameter name just for consistency and less confusion

thollingshead:
@gavinr It is [documented](https://esri.github.io/arcgis-rest-js/api/portal/updateItemResource/) that `prefix` (from [IItemResourceOptions](https://esri.github.io/arcgis-rest-js/api/portal/IItemResourceOptions/)) should work. This is also consistent with [ the code from addItemResource](https://github.com/Esri/arcgis-rest-js/blob/98be7a74b8362cb20170c3f62d4a61c842a1f8d5/packages/arcgis-rest-portal/src/items/add.ts#L130).

gavinr (approved):
Great, looks good to me then.

PR #826, opened 2021-04-12
Update Origin Destination Matrix Response Typings

mpayson:
I learned from the product team that `messages` in the origin destination service response has a structure like:
```
{
  "messages": [{
      "type": 0,
      "description": "A message description"
    }, {
      "type": 50,
      "description": "Another message description"
    }],
  ...rest
}
```

Updated the typings to reflect this.

### DIFF
--- a/packages/arcgis-rest-routing/src/originDestinationMatrix.ts
+++ b/packages/arcgis-rest-routing/src/originDestinationMatrix.ts
@@ -55,7 +55,7 @@ interface IFeatureSetWithGeoJson extends IFeatureSet {
 }
 
 export interface IOriginDestinationMatrixResponse {
-  messages: string[];
+  messages: [ { type: number, description: string } ];
   /**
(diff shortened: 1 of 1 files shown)


gavinr:
Thanks @mpayson. I would prefer if we could get that [documented publicly](https://developers.arcgis.com/rest/network/api-reference/origin-destination-cost-matrix-synchronous-service.htm), but lacking that, I am able to confirm experimentally if you send an invalid input for the `restrictionAttributeNames` we do see an object consistent with the type in this PR:
image

So, looks good!

gavinr (approved):

goodefroi:
Hey @mpayson @gavinr, I'm afraid the typings don't exactly match the described structure.

`[ { type: number, description: string } ]` means "an array containing exactly one object of the type `{ type: number, description: string }`".

What you're probably looking for is `{ type: number, description: string }[]`, meaning "an array containing an arbitrary number of objects of type `{ type: number, description: string }`".

gavinr:
Thanks @goodefroi - I'll create an issue. PRs to fix this certainly welcome.

PR #831, opened 2021-05-03
Add support for sending all credentials to trusted domains

patrickarlt:
In the process of chasing down a very specific bug in Vector Tile Style Editor for a very specific customer with a very specific security setup we found out that ArcGIS REST JS needs to support sending all credentials to servers defined in the `authorizedCrossOriginDomains` property on `portals/self`.

The reason for this is that in some high security scenarios organizations ALSO authenticate with BOTH some sort of cookie based auth (PKI/IWA) AND token based auth. In this case the domains that should also get the cookie based auth are defined in the `authorizedCrossOriginDomains` property on `portals/self` so we have to request that before trying to federate otherwise getting the server info will fail.

@dasa we discussed that a protocol is not strictly required by the UI for this. In this case we treat `secure.esri.com` as `https://secure.esri.com` to always force HTTPS in these kinds of high security setups. I think you told us the JS API will treat this as `https://secure.esri.com` OR `http://secure.esri.com` for purposes of matching. I think we should support HTTPS only in this case. Let me know if you think this is a problem.

@noahmulfinger @tomwayson @dbouwman @MikeTschudi  this makes a few internal chagnes to `UserSession`:

* Renames `trustedServers` to `federatedServers` so we can have the new `trustedDomains` property and not have it be confusing.
* Adds a new `getPortal` like `getUser` however this doesn't have any type information because I didn't want to move `IPortal` into common types and make this PR even larger.
* There is a new optional methods on `IAuthenticationManager.getDomainCredentials` which will return the proper value for Fetch `credentials` based on a URL this only needs to be implemented for `UserSession` and is used internally by `request`.
* Most tests have been updated to also mock the new `portal/self` call which is required.


dasa:
> @dasa we discussed that a protocol is not strictly required by the UI for this. In this case we treat `secure.esri.com` as `https://secure.esri.com` to always force HTTPS in these kinds of high security setups. I think you told us the JS API will treat this as `https://secure.esri.com` OR `http://secure.esri.com` for purposes of matching. I think we should support HTTPS only in this case. Let me know if you think this is a problem.

The [doc](https://enterprise.arcgis.com/en/portal/latest/administer/windows/configure-security.htm#ESRI_SECTION1_70CC159B3540440AB325BE5D89DBE94A) isn't specific about this either, other than saying that the protocol is optional, but it does seem reasonable. It'd only be an issue if the user expected it to work over `http` and the app was using `http` and the URL to the service was also `http`. I'm not sure how common this is.

dasa (approved):
Looks good 👍🏻

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -830,6 +830,44 @@ export class UserSession implements IAuthenticationManager {
     }
   }
 
+  /**
+   * Returns information about the currently logged in users [portal](https://developers.arcgis.com/rest/users-groups-and-items/portal-self.htm). Subsequent calls will *not* result in additional web traffic.
```
noahmulfinger, line 834:
```suggestion
   * Returns information about the currently logged in user's [portal](https://developers.arcgis.com/rest/users-groups-and-items/portal-self.htm). Subsequent calls will *not* result in additional web traffic.
```

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -1222,4 +1289,40 @@ export class UserSession implements IAuthenticationManager {
...
+  /**
+   * ensures that the authorizedCrossOriginDomains are obtained from the portal and cached
+   * so we can check them later.
+   *
+   * @returns this
+   */
+  private fetchAuthorizedDomains() {
+    // if this token is for a specific server or we don't have a portal
+    // don't get the portal info because we cant get the authorizedCrossOriginDomains
+    if (this.server || !this.portal) {
```
noahmulfinger, line 1302: The second part of this if statement isn't necessary because `this.portal` always is truthy. It defaults to www.arcgis.com in the constructor if one isn't passed in.

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -715,11 +700,17 @@ export class UserSession implements IAuthenticationManager {
    */
   private _user: IUser;
 
+  /**
+   * Hydrated by a call to [getPortal()](#getPortal-summary).
+   */
+  private _portal: any;
```
noahmulfinger, line 706: Might be worth `_portalInfo` or something similar since there's already a `this.portal` variable that refers to the url, not the data

packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -1839,4 +1920,327 @@ describe("UserSession", () => {
...
+
+  describe(".getPortal()", () => {
+    afterEach(fetchMock.restore);
+
+    it("should cache metadata about the portal", (done) => {
+      // we intentionally only mock one response
+      fetchMock.once(
+        "https://www.arcgis.com/sharing/rest/portals/self?f=json&token=token",
+        {
+          fetchAuthorizedDomains: ["gis.city.com"],
```
noahmulfinger, line 1932: Wrong variable name here and in a few other places:
```suggestion
          authorizedCrossOriginDomains: ["gis.city.com"],
```

noahmulfinger (approved):
Overall LGTM. I have a few minor comments.

dbouwman (approved):
Nothing to add beyond @noahmulfinger's suggestions. Thanks for digging deep into the weeds and resolving this.

MikeTschudi (approved):
Thanks, Patrick; I don't think that it will break our use of the library.

packages/arcgis-rest-auth/src/UserSession.ts
patrickarlt: @noahmulfinger this would catch cases where a user explicitly set `portal` to `null` or `undefined` which would make this error. I'm not sure what other errors that might cause but it seemed best to check for it here.

Issue #840, opened 2021-05-13
Token duration is improperly set in `ouath/authorize` workflow


hogpilot:
[This](https://github.com/Esri/arcgis-rest-js/blob/e59e49968ae9bc88c76035e81aaffc9c971080f1/packages/arcgis-rest-auth/src/UserSession.ts#L537) line of code improperly sets the query param `duration` instead of `expiration` (see [here](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)).

* Invalid parameter `duration` (current code)
  ``` js
  response.writeHead(301, {​
    Location: `${​portal}​/oauth2/authorize?client_id=${​clientId}​&duration=${​duration}​&response_type=code&redirect_uri=${​encodeURIComponent(
      redirectUri
    )}​`,
  }​);
  ```
* Correct parameter `expiration` (proposed fix)
  ``` js
  response.writeHead(301, {​
    Location: `${​portal}​/oauth2/authorize?client_id=${​clientId}​&expiration=${​duration}​&response_type=code&redirect_uri=${​encodeURIComponent(
      redirectUri
    )}​`,
  }​);
  ```

PR #841, opened 2021-05-13
Fix UserSession token duration for oauth

hogpilot:
Addresses an issue where token duration was not honored in oauth flow due to improper \`duration\`

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

ISSUES CLOSED: #840


noahmulfinger (approved):
LGTM!

tomwayson (approved):
Thanks @hogpilot - I can't believe no one has caught this up until now.

We should add an issue to add an `expiration` to  and deprecate `duration` from `IOAuth2Options`.

PR #845, opened 2021-05-18
fix(UserSession): switch "duration" to "expiration" in IOAuth2Options

gavinr:
AFFECTS PACKAGES:
@esri/arcgis-rest-auth

BREAKING CHANGE:
IOAuth2Options: "duration" is removed. Please use "expiration"

ISSUES CLOSED: #843

@tomwayson is there a better way to deprecate?

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -103,9 +103,9 @@ export interface IOAuth2Options {
   provider?: AuthenticationProvider;
 
   /**
-   * Duration (in minutes) that a token will be valid. Defaults to 20160 (two weeks).
+   * The requested validity in minutes for a token. Defaults to 20160 (two weeks).
    */
-  duration?: number;
+   expiration?: number;
 
   /**
    * Determines whether to open the authorization window in a new tab/window or in the current window.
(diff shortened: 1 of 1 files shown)


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -103,9 +103,9 @@ export interface IOAuth2Options {
   provider?: AuthenticationProvider;
 
   /**
-   * Duration (in minutes) that a token will be valid. Defaults to 20160 (two weeks).
+   * The requested validity in minutes for a token. Defaults to 20160 (two weeks).
    */
-  duration?: number;
```
tomwayson, line 108: I was thinking we'd have both until the next breaking change so we would **not** have to do a breaking change for this release. We'd leave `duration`, but change the comments to indicate that it is deprecated and to use `expiration` instead. Then in the code below we'd use `expiration || duration` until the next major release.

tomwayson (commented):
thanks @gavinr, but I think we should either hold this until the next release, or change it to deprecate `duration` instead of remove it.

packages/arcgis-rest-auth/src/UserSession.ts
gavinr: got it. That's what I initlally was doing, using the `@deprecated` tag, but then the build was erroring at me for using it. I think I was probably just not doing it correctly. I'll bring that code back into this PR then re-ping you. Thanks!

PR #854, opened 2021-06-15
Admin's share items via user item route

dbouwman:
Prepping for Sharing API v 9.2, there is a change that impacts `shareItemsToGroup` when the user doing the sharing is not the owner, but rather an `org_admin`.

Prior to 9.2, the `/content/items/:id/share` endpoint would allow an `org_admin` to share a private item, owned by another user, to a group (item owner must already be a member in said group).

At 9.2. that no longer works; instead we need to use the `/content/users/:ownername/items/:id/share` route, which is what this PR changes

This is an internal change.

Have verified that this works in PROD, as well as test envs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1440,7 +1440,7 @@
     "@types/mz": {
       "version": "0.0.32",
       "resolved": "https://registry.npmjs.org/@types/mz/-/mz-0.0.32.tgz",
-      "integrity": "sha1-6CSLTkFCTAUu3Bcl3TNlDDE6Nlk=",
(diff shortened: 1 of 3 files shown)


tomwayson (approved):
I'm just rubber stamping this b/c:
- it looks like 90% formatting changes
- you said you have a script that verifies it works in prod!

dbouwman:
Travis-ci.org is basically shut down.

image

dbouwman:
We are collectively unclear what's up w/ Travis - it _seems_ like we should be able to simply switch to travis-ci.com, but the activation is failing w/o any messages. I'm proceeding to merge this as we need the change in Hub

PR #848, opened 2021-05-19
Switch out tslint for eslint

COV-GIS:
As suggested by https://github.com/Esri/arcgis-rest-js/issues/844.

@tomwayson, @patrickarlt and/or @gavinr you may want to pull this down and see what you think of how updated prettier transforms many of the src and test files. I didn't realize the how the styling differed between packages, e.g. trailing commas.

I didn't commit the prettied files, but lints and builds without errors.

PR #855, opened 2021-06-16
Migrate repo from tslint to eslint

MikeTschudi:
tslint was deprecated over a year ago, and eslint is a recommended replacement.

Got eslint working with a subset of rule set, and then restored some of the disabled rules, with code changes for those rules
* Added no-useless-escape
* Added prefer-const
* Restored no-control-regex
* Restored no-redeclare


gavinr:
We were doing the same thing over here: https://github.com/Esri/arcgis-rest-js/pull/848 .. not sure how this compares to that one, but maybe we should discuss the different approaches

MikeTschudi:
@gavinr, I'll put this PR into draft; #848 is the better way to go since it is farther along.

dbouwman:
Switched in #863 - closing

PR #858, opened 2021-06-16
doc/Corrected function return type

MikeTschudi:


dbouwman (approved):
That's interesting... and we need to discuss w/ Tom et al how to release this... for non-typescript consumers, this is not a change at all... but for typescripters, this is a breaking change...

tomwayson:
I say go w/ a patch release. Those devs knew what they were getting into when they chose TS.

dbouwman:
Travis still seems to be non-functional... 😢

I'll pull and verify, then merge and release... 🤖

PR #867, opened 2021-06-21
removed unnecessary `package-lock.json` files

gavinr:
There is currently a `package-lock.json` file within each of our `packages/*` folders.  they seem to be out-of-sync with the `package.json` files, probably because our release process is not keeping them up to date. I think the best way to clean this up is to remove them, as it seems like they're not fulfilling a useful purpose.


tomwayson (approved):
thanks @gavinr, I've been confused about whether or not those should be there and according to https://www.linkedin.com/pulse/things-i-wish-had-known-when-started-javascript-monorepo-gorej/ (see "A word about package-lock.json files") at least, they should  not.

gavinr:
Sounds good. Thanks @tomwayson for the research and good link, and thank you @char0n for the great article!

char0n:
@gavinr np, will update in near future with section how things look now as we have npm@7 and native workspaces with it.

Issue #804, opened 2021-02-20
Broken link to browser-based login image in documentation


MikeTschudi:
On page https://esri.github.io/arcgis-rest-js/guides/browser-authentication/, the link to "browser-based login" is broken:
![image](https://user-images.githubusercontent.com/2125181/108599623-21179d00-7347-11eb-8176-5f4f799964d6.png)

Should this link be removed, or will this image do the job: https://developers.arcgis.com/documentation/static/5a68e29d341f1a1e7641b6bc289c9c2f/4cdf7/4-oauth-screen.png?

jf990:
Image referenced above has been taken down, we just published updated Security and Authentication docs with new images

https://developers.arcgis.com/documentation/mapping-apis-and-services/security/arcgis-identity/serverless-web-apps/

PR #870, opened 2021-06-22
docs(docs): fix dead image link

gavinr:
ISSUES CLOSED: #804 

There is no permanent link to that image anymore, so hosting here instead of link out to developers.arcgis.com


MikeTschudi (approved):
Thanks for the straightforward solution!

PR #879, opened 2021-07-19
Improve adminServiceInfo on IServiceInfo type

sandromartis:
Taken from the response example found here:
https://developers.arcgis.com/rest/services-reference/online/hosted-feature-service.htm

**Questions:**
- I made `adminServiceInfo` non-optional. Do we know if the response always returns `adminServiceInfo`?
- Can we narrow `type` and `status` further down? I'm unsure about all the possible values.

@dbouwman

### DIFF
--- a/packages/arcgis-rest-types/src/service.ts
+++ b/packages/arcgis-rest-types/src/service.ts
@@ -117,7 +117,16 @@ export interface IFeatureServiceDefinition {
  * /arcgis/rest/admin/services/<service-name>/FeatureServer?f=json response
  */
 export interface IServiceInfo extends Record<string, unknown> {
-  adminServiceInfo?: Record<string, unknown>;
+  adminServiceInfo: {
+    name: string;
+    type: string;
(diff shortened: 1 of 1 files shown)


dbouwman:
@sandromartis I _think_ this is ok - I had a vague recollection of some cross org issue accessing the this route, and sometimes not getting that node in the response, but looking in `templates-js` I don't see any checks for a undefined `adminServicveInfo` node. As for `type` and `status` values - I'm really unsure what the valid values are. I kind of suspect that what's under `.database` has more variability... if you are running into typescript complaints about accessing deep properties that the typing is not defined for, I use `getProp(obj, 'path.to.prop')` from `@esri/hub-common`... i.e.

```js
   const serviceName = getProp(serviceInfo, 'adminServiceInfo.name');
```

There is also `getWithDefault(obj, 'path.to.prop', 'Default Value')`

gavinr (approved):
Looks good

PR #883, opened 2021-07-29
outStatisticFieldName should be optional

dangrussell:
Documentation states "If outStatisticFieldName is empty or missing, the server will assign one."


gavinr (approved):
Thank you for the submission!

Looks correct to me:

image

https://developers.arcgis.com/rest/services-reference/enterprise/query-feature-service-layer-.htm

PR #849, opened 2021-05-24
Update server-based oauth 2.0 for state and token consistency

gavinr:
Updates from #704 merged with the latest code.

Addresses #699 & #660 as well as a couple other tweaks:
* change the 'duration' URL parameter to 'expiration' in `authorize()` to match the [REST endpoint](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)
* update 'refreshTokenExpires' calculation to reflect that 'refreshTokenTTL' is in minutes

Also related to #842

Issue #842, opened 2021-05-14
Refresh token expiration is not accurate


hogpilot:
In a `UserSession` object (specifically with an `authorize` workflow), the `refreshTokenTTL` property is what drives the expiration time of the refresh token (`refreshTokenExpires`).  There are a couple issues with this:

1. It is set directly via the constructor options (`refreshTokenTTL`), or defaulted to `1440`.  This seems to only be used internally to the `UserSession` object and not know by ArcGIS API calls.

    * It isn't representative of when the refresh token actually expires, which is described [here](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm) as defaulting to `20160` (2 weeks) or as specified in the `expiration` parameter of the  `ouath2/authorize` API call

    * This will cause an unhandled exception [here](https://github.com/Esri/arcgis-rest-js/blob/e59e49968ae9bc88c76035e81aaffc9c971080f1/packages/arcgis-rest-auth/src/UserSession.ts#L1267) and [here](https://github.com/Esri/arcgis-rest-js/blob/e59e49968ae9bc88c76035e81aaffc9c971080f1/packages/arcgis-rest-auth/src/UserSession.ts#L1291) when the refresh token actually has expired, but `UserSession` *thinks* it is still valid.
   
2. It is [initially](https://github.com/Esri/arcgis-rest-js/blob/e59e49968ae9bc88c76035e81aaffc9c971080f1/packages/arcgis-rest-auth/src/UserSession.ts#L579) treated as "number of seconds", but then [later](https://github.com/Esri/arcgis-rest-js/blob/e59e49968ae9bc88c76035e81aaffc9c971080f1/packages/arcgis-rest-auth/src/UserSession.ts#L1297) on (after a call to `refreshSession`) as "number of minutes".

This can be reproduced by waiting at least 30 minutes (the default accest token duration) between initial creation and refresh (due to #840, you can't currently set the expiration with `arcgis-rest-js`).

Once that issue is fixed, then reproducing it can be done more quickly by specifying a `duration` of `1` when creating a new `UserSession` (authorize) instance... waiting a minute, then trying to call `refreshSession`.

mpayson:
Ref #704 too, it's gone stale but I think it tackles this

patrickarlt:
This will be resolved at v4 which is currently on the beta branch.

PR #884, opened 2021-08-02
refreshTokenTTL fix

gavinr:
- expected: `refreshTokenTTL` is set to 14 days by default
- actual: `refreshTokenTTL` is set at 1 day by default

#842 
follow-up from #849.

### DIFF
--- a/packages/arcgis-rest-auth/test/UserSession.test.ts
+++ b/packages/arcgis-rest-auth/test/UserSession.test.ts
@@ -1625,7 +1625,6 @@ describe("UserSession", () => {
     it("should return a UserSession where refreshTokenExpires is 2 weeks from now (within 10 ms)", (done) => {
       fetchMock.postOnce("https://www.arcgis.com/sharing/rest/oauth2/token", {
         access_token: "token",
-        expires_in: 1800,
         refresh_token: "refreshToken",
         username: "Casey",
         ssl: true,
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -1621,6 +1621,33 @@ describe("UserSession", () => {
           fail(e);
         });
     });
+
+    it("should return a UserSession where refreshTokenExpires is 2 weeks from now (within 10 ms)", (done) => {
+      fetchMock.postOnce("https://www.arcgis.com/sharing/rest/oauth2/token", {
+        access_token: "token",
+        expires_in: 1800,
```
gavinr, line 1628: should remove this line

hogpilot: I agree that we should test the default value and not the `expires_in` option with this test.

hogpilot (changes requested):

packages/arcgis-rest-auth/test/UserSession.test.ts
gavinr: thanks - removed.

tomwayson (approved):
thanks @gavinr!

hogpilot (approved):

PR #886, opened 2021-08-03
Remove 'new' keyword when calling UserSession.fromCredential()

dangrussell:
Only a void function can be called with the 'new' keyword.


gavinr:
Thank you for this PR. This covers the code in the README, but should we also make the equivalent change in the code?  https://github.com/Esri/arcgis-rest-js/blob/22b67cf65555c91b8e18dd19f0c83959dc648f53/demos/jsapi-integration/index.html#L83

gavinr (approved):
Looks good. While testing this I also made some general updates to the jsapi-integration sample.

PR #894, opened 2021-08-04
Clarified output of UserSession's beginOAuth2 and fixed minor typo

MikeTschudi:


packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -387,7 +388,7 @@ export class UserSession implements IAuthenticationManager {
   }
 
   /**
-   * Completes a browser-based OAuth 2.0  in. If `options.popup` is `true` the user
+   * Completes a browser-based OAuth 2.0 sign in. If `options.popup` is `true` the user
```
gavinr, line 391: Great catch!

packages/arcgis-rest-auth/src/UserSession.ts
```diff
@@ -297,8 +297,9 @@ export class UserSession implements IAuthenticationManager {
 
   /**
    * Begins a new browser-based OAuth 2.0 sign in. If `options.popup` is `true` the
-   * authentication window will open in a new tab/window otherwise the user will
-   * be redirected to the authorization page in their current tab/window.
+   * authentication window will open in a new tab/window and the function will return
+   * Promise&lt;UserSession&gt;. Otherwise, the user will be redirected to the
+   * authorization page in their current tab/window and the function will return `undefined`.
```
gavinr, line 302: If it's possible to return `undefined`, should we also update that in our typescript types? Changing line 307 to:


```ts
public static beginOAuth2(options: IOAuth2Options, win: any = window): Promise<UserSession> | undefined {
```

MikeTschudi: Indeed--thanks, Gavin!

gavinr (approved):
looks good to me.

PR #904, opened 2021-09-01
Update TypeDoc to use the new --package option

patrickarlt:
This PR updates TypeDoc to `^0.21.0` to use the new `--packages` option  which also requires updating TypeScript to `^4.0.0` and `tslib` to `^2.3.0`.

This work is in preparation to migrate the API reference the ArcGIS Developer site.

**After merging this PR:** We bump a new major version because our underlying dependencies have new major versions and release.  The new release script builds a TypeDoc site and JSON and packages it into the release.

Then on our internal repo that builds  https://developers.arcgis.com/arcgis-rest-js/ we can:

1. We add a configuration option to build of ArcGIS REST JS for the developer site to indicate which release we should pull doc source from.
2. When we build https://developers.arcgis.com/arcgis-rest-js/ download the latest release un-zip it and access the `typedoc.json` file so we can build our custom rendering of the TypeDoc.
3. Develop our custom  TypeDoc rendering to replace https://esri.github.io/arcgis-rest-js/api/
4. Migrate remaining doc from https://esri.github.io/arcgis-rest-js/guides/ to https://developers.arcgis.com/arcgis-rest-js/ 
5. Once the full migration is complete we can sunset https://esri.github.io/arcgis-rest-js/api/ and point users to https://developers.arcgis.com/arcgis-rest-js/  for all doc. Users can still download the built TypeDoc HTML for each release.


dbouwman:
@patrickarlt seems like your local was behind master, and thus there's a mess of conflicts, but otherwise this looks good to me. Later today I'll pull it and run things, but I assume it all works. Thanks for jumping in and setting this up - I'll try to get a story in our backlog to apply the same updates to Hub.js

tomwayson:
I was going to say the same as @dbouwman - if you resolve the conflicts, I'll pull and test it out.

Also, @gavinr - aren't we supposed to not have package-lock.json files at the individual package level anymore?

patrickarlt:
@gavinr it looks like we have run afoul of https://jasmine.github.io/pages/faq.html#module-spy. As of TypeScript 4.0 we can no longer spy on module exports like `request`. I rewrote the tests to look at what `fetchMock` sees rather then what we can spy on from `request`.

tomwayson:
@dbouwman 

> As of TypeScript 4.0 we can no longer spy on module exports like request.

That is very bad news for Hub-land. Hub.js has 762 invocations of `spyOn`, I'm sure most of which are spying on module exports.

patrickarlt:
@tomwayson @dbouwman from https://jasmine.github.io/pages/faq.html#module-spy:

> Use dependency injection for things you’ll want to mock, and inject a spy or a mock object from the spec. This approach usually results in maintainability improvements in the specs and the code under test. Needing to mock modules is often a sign of tightly coupled code, and it can be wise to fix the coupling rather than work around it with testing tools.

Maybe we need to have `request` be an option that overrides the default request handler entirely. Then you can pass a spy in from the tests to keep track of the calls. still requires updating 762 tests though...

gavinr (approved):
Looks good!

PR #909, opened 2021-09-07
od matrix - fixed type on `messages` parameter

### LINKED ISSUE #834: Follow-up fix to OD-Matrix types
Need to modify https://github.com/Esri/arcgis-rest-js/pull/826/files per suggestion https://github.com/Esri/arcgis-rest-js/pull/826#issuecomment-833531257:

>[ { type: number, description: string } ] means "an array containing exactly one object of the type { type: number, description: string }".

> What you're probably looking for is { type: number, description: string }[], meaning "an array containing an arbitrary number of objects of type { type: number, description: string }".

gavinr:
This is a follow-up fix to #826.  Makes the change to the types and also adds some unit tests that verify that the types are working properly.

Resolves #834


mpayson (approved):
LGTM! Thank you for picking this up

Issue #833, opened 2021-05-04
ISearchOptions - Enumerate options for search sortOrder


gavinr:
When using [ISearchOptions](https://esri.github.io/arcgis-rest-js/api/portal/ISearchOptions/), `sortOrder` is typed as any string, when [according to the REST API doc, there are only two possible options](https://developers.arcgis.com/rest/users-groups-and-items/search.htm#GUID-A2C8BFD1-89B2-483F-9275-65D2786F6A6C): `asc | desc`.

Should we make our types reflect this?

Also on [ISearchOptions](https://esri.github.io/arcgis-rest-js/api/portal/ISearchOptions/), the descriptions are empty -
![image](https://user-images.githubusercontent.com/209355/117022631-e8e9ee80-ace7-11eb-983a-cf6a29e4fb7d.png)

... it would be nice to fill these in to help developers:

![image](https://user-images.githubusercontent.com/209355/117022845-19318d00-ace8-11eb-98fb-77bc100118b9.png)

PR #910, opened 2021-09-07
added sortOrder options and documentation

gavinr:
#833


tomwayson (approved):
thanks @gavinr 

prob best to do this as part of the 4.0 release since we're so close, not sure if that's going to it's own branch tho.

PR #913, opened 2021-09-10
fix(arcgis-rest-portal): query terms containing colons need to be wra…

andrewctate:
…pped in quotes to be valid

AFFECTS PACKAGES:
@esri/arcgis-rest-portal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,22711 @@
 {
   "name": "@esri/arcgis-rest-js",
   "version": "3.3.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@esri/arcgis-rest-js",
+      "version": "3.3.0",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "cross-fetch": "^3.0.0",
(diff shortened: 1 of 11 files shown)


tomwayson (commented):
Thanks @andrewctate 

This stands to reason, but:
1. Looks like no additional test case was needed for coverage, but would you mind adding one to help ensure we don't regress on this in the future
2. @gavinr, @patrickarlt I think we're in a pickle here b/c master has unreleased breaking changes in #904

> After merging this PR: We bump a new major version because our underlying dependencies have new major versions and release. The new release script builds a TypeDoc site and JSON and packages it into the release.

I think to be safe we should revert #904 on master, then merge this.

andrewctate:
@tomwayson thanks. The test case is [here](https://github.com/Esri/arcgis-rest-js/pull/913/files#diff-704937f1df8c53f5d05cef4037d337a9979a3055bb868d2f7b0361c0147501feR42).

patrickarlt (approved):
LGTM thanks @andrewctate. 

@tomwayson if we want to cut a new release with this in it we can revert https://github.com/Esri/arcgis-rest-js/pull/904/files?file-filters%5B%5D=.json and put it on the 4.0 branch.

tomwayson:
Please wait for #915 before merging this.

Issue #917, opened 2021-09-13
[regression] Admins can no longer add items from outside their org to a group


oevans:
As mentioned in PR https://github.com/Esri/arcgis-rest-js/pull/854, the endpoint used to add an item to a group was recently changed to account for a change in the Sharing API v9.2.

> Prior to 9.2, the `/content/items/:id/share` endpoint would allow an `org_admin` to share a private item, owned by another user, to a group (item owner must already be a member in said group).
>
> At 9.2. that no longer works; instead we need to use the `/content/users/:ownername/items/:id/share` route, which is what this PR changes.

This change, however, created a regression such that admin users can no longer add items owned by someone outside their org to a group. This affects not only standard group/sharing workflows but also favoriting (since favoriting is implemented as a group).

A specific case of this was already reported in issue #882, but this issue more broadly affects an admin sharing (or favoriting) _any item_ outside their org, not just Living Atlas items.

patrickarlt:
@tomwayson @dbouwman I think both of you wrote most of the original sharing code can you take a look?

dbouwman:
Reading the code and the issue noted, I think...

we want to use `/content/users/:ownername/items/:id/share` only:

- current user is not the owner of the item
- current user has `role: "org_admin"`
- item access !== public
- item is in the same org as the Admin OR item owner is in same org as Admin

otherwise use `/content/items/:id/share`.

Currently we don't get the item, so if we really want to restrict this flow to only non-public items, we'd have to do yet-another xhr. During the flow we do fetch the item owner if the `currentUser !=== owner`, and so we could verify that the owner and the current user are in the same org (thus implying the item is in the same org)... this will take some refactoring, which is always scary w/ this part of the API - We do have a harness that we can use to verify this against the live api, so that's a plus... 

I can try to get to this early next week ~9/20/21 - if someone inside Esri wants to tackle this sooner, the harness lives in devtopia at /dc/portal-api-check

tomwayson:
@oevans this is released in v3.4.1

oevans:
Great -- thanks all!!

oevans:
@tomwayson @dbouwman I'm not sure if anything needs to be done in `rest-js`, but it looks like the back-end made a change related to this previous fix...

https://devtopia.esri.com/WebGIS/arcgis-portal-app/issues/921

Issue #882, opened 2021-07-28
Unable to favorite any Living Atlas item as an organization administrator using shareItemWithGroup


aravindesri:
### Issue description
I, as an org admin, am unable to favorite a Living Atlas item (Owned by a user outside of my organization) to my favorite's group.

### Context
Since https://github.com/Esri/arcgis-rest-js/commit/e8ab1bc3ad4498e5c1ac19fc0f1c24013de0a8df, the `shareItemWithGroup` now handles a different API route to share an item to a group if the sharing was performed by an org_admin. This route uses the `/content/users/:ownername/items/:itemid/share` which when favoriting a Living Atlas item, does not have the right permissions to add to favorite group since the item and the user don't belong to the same organization as the org admin and therefore no item control.

### Steps to reproduce
1. Use the following inputs to run the shareItemWithGroup function: 
    A. id: bf7c462f912f4d41ba262ba3c0e9008f
    B. groupId: "Org admin favorite group ID"
    C. owner: "cfrye_UO"
    D. authentication: Org admin user session

### Seen behavior
I get an error that I do not have permissions to favorite the item.

### Expected behavior
I would expect that an admin is able to favorite a Living Atlas item just like any other member of the organization.

### Version
Browser: Chrome, Firefox, Safari (all OS)
ArcGIS-Rest-Js: v3.2.0

### Use case
The ArcGIS StoryMaps app offers several item browser views in its ecosystem. There's a browser to explore and search for Living Atlas web maps, scenes, storymaps, apps and files. This browser view can be accessed by both an admin and any non-admin user. 
<img width="1400" alt="Screen Shot 2021-07-28 at 12 15 59 PM" src="https://user-images.githubusercontent.com/17153133/127382488-21cc7d6a-a9ad-4446-8e4a-a2c158ab4db8.png">

We're running into this issue specifically for the org_admin use case.

patrickarlt:
@dbouwman @tomwayson can you take a look?

dbouwman:
Addressed by #919

tonybattsesri:
We are currently having what seems to be the same issue, only occuring in our lower environments that are using `"@esri/arcgis-rest-portal": "^4.0.0"`

We're getting the error message: `You do not have permissions to access this resource or perform this operation` when calling shareItemWithGroup() as a user that does not own the item they are trying to favorite.

In staging the share request is hitting this end point:
[...]

PR #919, opened 2021-09-20
fix(arcgis-rest-portal): admin uses items/:id/share route for items outside their org

dbouwman:
AFFECTS PACKAGES:
@esri/arcgis-rest-portal

ISSUES CLOSED: #917 #882

Verified against PROD API using https://devtopia.esri.com/dc/portal-api-check/blob/master/spec/admin-share-living-atlast-to-own-group.spec.ts#L36

### DIFF
--- a/packages/arcgis-rest-portal/src/sharing/share-item-with-group.ts
+++ b/packages/arcgis-rest-portal/src/sharing/share-item-with-group.ts
@@ -66,6 +66,8 @@ export function shareItemWithGroup(
       if (itemOwner !== username) {
         // need to track if the user is an admin
         let isAdmin = false;
+        // track if the admin & owner are in the same org
+        let isCrossOrgSharing = false;
         // next perform any necessary membership adjustments for
         // current user and/or item owner
         return Promise.all([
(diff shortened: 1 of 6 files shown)


packages/arcgis-rest-portal/test/sharing/share-item-with-group.test.ts
```diff
@@ -1325,4 +1325,58 @@ describe("shareItemWithGroup() ::", () => {
         });
     });
   });
+  describe("share item from another org to admin user's favorites group ::", () => {
+    it("should share item", (done) => {
```
dbouwman, line 1329: This describe is very similar to the one above, but we just change things so the owner user is from another org, and verify that the sharing call happens on the /items/:id/share route, and not the /users/:username/items/:id/share route

packages/arcgis-rest-portal/src/sharing/share-item-with-group.ts
```diff
@@ -207,8 +211,8 @@ function shareToGroup(
 
   // but if they are the owner, or org_admin, use this route
   // Note: When using this end-point as an admin, apparently the admin does not need to be a member of the group (the itemOwner does)
-  // Have not validated this but came up in conversations w/ Sharing API team
-  if (itemOwner === username || isAdmin) {
+  // Note: Admin's can only use this route when the item is in the same org they are admin for
+  if (itemOwner === username || (isAdmin && !isCrossOrgSharing)) {
```
dbouwman, line 215: This is the crux of the fix - if the item is from some other org, even an admin can't use the `/content/users/:username/items/:id/share` route.

tomwayson (approved):
thank you @dbouwman!

PR #922, opened 2021-10-01
fix(arcgis-rest-portal): ensure item owner is temporarily member of group

juliannemarik:
[1050](https://devtopia.esri.com/dc/hub/issues/1050)

currently we only make membership adjustments for non-owners trying
to share an item. In the case that the item owner (also an admin),
tries to share an item to a group, we also need to ensure they are
(temporarily) a member of the group

AFFECTS PACKAGES:
@esri/arcgis-rest-portal


juliannemarik:
@dbouwman @rweber-esri, I believe you have much more familiarity with this function. Let me know if this change seems like a reasonable solution.

dbouwman:
@juliannemarik I think this looks reasonable, but before we move ahead with this, I'd like to chat w/ Brian re: alternative fixes within the Hub application. I _think_ this join-group-share-item-leave-group flow is something that _may_ have worked in the past, but is not how the Sharing API really wants things to work. I'll email you a link to the functional matrix, and you'll see a) it's crazy complex, and b) there are known scenarios where it allows things it should not (Overshare)... and the team is actively fixing those things.

PR #925, opened 2021-10-12
types(feature-layer): make objectIds optional

roemhildtg:
objectIds is currently required by the `IDeleteFeaturesOptions` - this should not be the case since you can also use where clauses to delete features.

### DIFF
--- a/packages/arcgis-rest-feature-layer/src/delete.ts
+++ b/packages/arcgis-rest-feature-layer/src/delete.ts
@@ -22,7 +22,7 @@ export interface IDeleteFeaturesOptions
   /**
    * Array of objectIds to delete.
    */
-  objectIds: number[];
+  objectIds?: number[];
 }
 
 /**


COV-GIS:
There is a couple reasons why `objectIds` is required. First, it's the most commonly used parameter for deleting features. Second, there's no conditional logic in typescript interfaces. Not only can you use `where` but you can also use `geometry` to delete features. And when providing `geometry` you are also are required to provide `geometryType` and possibly `inSR`. But one or some combination of the three has to be included to prevent a server error.

I believe you can still use `where` or `geometry` with `objectIds: []` to make TS happy and still successfully delete features.

Or there would need to be extensive interface splitting to account for all possibilities of parameters that may be passed to `deleteFeatures()`.

I recall this being brought up when the CRUD methods were added to the package but couldn't hunt down the specific comment.

gavinr:
Thanks all, and sorry about the delay on this. We'll merge this since setting the type to "optional" is not a breaking change, and then we'll follow-up by creating an issue to re-factor this issue holistically, taking into account some of the things that @COV-GIS mentioned:

- you can delete via a `where` or `objectIds` or `geometry` (or others too perhaps?)
- If you pass a `geometry` you also need to pass a geometry `geometryType`
- etc

gavinr (approved):

Issue #789, opened 2020-12-18
Move non-shared types to 


tomwayson:
Related to #786, many, if not most interfaces in the @esri/arcgis-rest-types are not shared between packages, and therefore could move to correct package. For example, I'm pretty sure none of these are used outside of portal:

https://github.com/Esri/arcgis-rest-js/blob/0aaeb093166d57be0e5e19c3f209eef9154a7978/packages/arcgis-rest-portal/src/index.ts#L55-L67

tomwayson:
I looked into this, and at least as far as the portal types, it only ends up being a handful of types that can be moved out of types, at least unless #786 ends up moving shared types to the "core" package (currently request).

PR #926, opened 2021-10-15
moved common types to arcgis-rest-request package

gavinr:
Completes #789.

- Move the "Item" types to `arcgis-rest-portal` package.
- Moves other common packages to `arcgis-rest-request` package.
- Removes `arcgis-rest-types` package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,7 +75,7 @@
         "slug": "^0.9.1",
         "sri-toolbox": "0.2.0",
         "trash-cli": "^4.0.0",
-        "ts-node": "^10.2.1",
+        "ts-node": "^10.3.1",
         "typedoc": "^0.21.0",
         "typescript": "^4.0.0",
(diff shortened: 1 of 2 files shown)


tomwayson:
I've been resistant to it, b/c originally it was one package (arcgis-rest-feature-service) that @jgravois and I [split into two at v2.0](https://esri.github.io/arcgis-rest-js/guides/whats-new-v2-0/).

However, tree-shaking is much better now, so I'm good to merge them back into one. I don't think we should `arcgis-rest-feature-layer` a name though. I don't mind the original name `arcgis-rest-feature-service`. there shouldn't be any issues resurrecting that as we're just going from v1.x to v4.0.0, but there's potential for some minor complications there.

Might be best to go w/ a new name. I think `arcgis-rest-features` is most consistent w/ the other packages (i.e. it's `arcgis-rest-geocoding` and not `arcgis-rest-geocode-server`).

Issue #914, opened 2021-09-10
Authentication manager design


patrickarlt:
This has been bugging me since our discussion with @hogpilot in https://github.com/Esri/arcgis-rest-js/issues/893. 

We want to encourage static methods for creating `UserSession` as opposed to `new UserSession()`, I.e. `UserSession.login()`, `UserSession.beginOAuth2()`, ect... but we don't encourage them for `ApiKey` or `ApplicationSession`.

Generally we didn't cover it because ApiKey and ApplicationSession are super simple whereas `UserSession` can accept a whole range of options that are difficult to construct but this seems weird from an API perspective.

@alaframboise has pointed out in https://github.com/Esri/arcgis-rest-js/issues/786#issuecomment-917115707 that `new APIKey` from the `request` package might imply a new key is created.

Similarly there are other issues:

* We have 3 classes `APIKey`, `ApplicationSession` and `UserSession`
* We have 3 general methods of authentication https://developers.arcgis.com/documentation/mapping-apis-and-services/security/#authentication-methods API Keys, Application Credentials and ArcGIS Identity.

Should we align these?

* `APIKey` > `ApiKeyManager` - manages authentication via an API key for ArcGIS REST JS authentication
* `ApplicationSession` > `ApplicationCredentialsManager` - manages authentication via Application Credentials for ArcGIS REST JS authentication
* `UserSession` > `IdentityManager`/`ArcGISIdentityManager` - manages authentication via an ArcGIS Identity for ArcGIS REST JS. Are we concerned with having an `IdentityManager` class in ArcGIS REST JS?

Then add static methods to `ApiKeyManager`  and `ApplicationCredentialsManager` to match `IdentityManager`. The preference will be that users always use these factory methods instead of `new ...` if possible.

```js
import {
  ApiKeyManager,
  ApplicationCredentialsManager
} from "@esri/arcgis-rest-request";

const myKey = ApiKeyManager.fromApiKey("AAPK...")
const myKey = new ApiKeyManager({
  apiKey: "AAPK..."
})

request("...", {
  authentication: myKey
})
```

```js
import {
  ApplicationCredentialsManager
} from "@esri/arcgis-rest-request";


const myApp = ApplicationCredentialsManager.fromCredentials({
  clientId: "•••",
  clientSecret: "•••"
});
const myApp = new ApplicationCredentialsManager({
  clientId: "•••",
  clientSecret: "•••"
});

request("...", {
  authentication: myApp
})
```

[...]

Issue #893, opened 2021-08-03
Clarify the recommended ways to create UserSession


gavinr:
Right now, [the documentation for UserSession](https://esri.github.io/arcgis-rest-js/api/auth/UserSession/) implies that the constructor (`new UserSession()`) should be used. 

After #887, the idea will be that in most cases a user should only ever call the static methods:

- `UserSession.beginOAuth2({...})`
- `UserSession.authorize({...})`
- `UserSession.login({...})` (new in #887)

.. so we should update the documentation page to make this more obvious.

patrickarlt:
Updated in #941

Issue #892, opened 2021-08-03
Add logic to allow sign-out (revoke token) for a UserSession


gavinr:
The ArcGIS REST API now includes a [`/revokeToken` endpoint](https://developers.arcgis.com/rest/users-groups-and-items/revoke-token.htm).

> when a user signs out of the app so that no abandoned tokens are left valid.

tomwayson:
related: #297

tomwayson:
I think we should start by adding a standalone `revokeToken()` function that wraps that end point.

Then we should add `session.signOut()` that calls `revokeToken` for the access and/or refresh tokens that the session is holding.

patrickarlt:
Reading the docs for the a [`/revokeToken` endpoint](https://developers.arcgis.com/rest/users-groups-and-items/revoke-token.htm) endpoint I think there are a few key things:

> An access token is issued from implicit grant and a refresh token can be revoked with this operation. 

This means that we can do 2 things:

* Revoke token granted by `beginOAuth2`/`completeOAuth2` cycle (implict grant)
* Revoke a refresh token and all associated tokens from `authorize`/`exchangeAuthorizationCode`

This also means (I think) that we cannot invalidate JUST an access token and force users to get a new token from their refresh token.

> Revoking an app access token and tokens generated from generate token are not supported with this operation.

So you cant revoke tokens from `ApplicationSession` or API keys with /revokeToken

---

I think because of these 2 things that this needs be a static `UserSession.destory()` or ``UserSession.signOut()`` method that calls `/revokeToken` for either the `refreshToken` (if available) or `token`. A generic `revokeToken` method could still be available but would need to clear that it only revoked the token in really specific situations.

patrickarlt:
Closing this. This will be part of the upcoming 4.0 release.

Issue #890, opened 2021-08-03
Add `style` to IOAuth2Options


gavinr:
- Would be added to https://esri.github.io/arcgis-rest-js/api/auth/IOAuth2Options/
- REST endpoint defined [here](https://developers.arcgis.com/rest/users-groups-and-items/authorize.htm)

patrickarlt:
Closing this. This will be part of the upcoming 4.0 release.

PR #931, opened 2021-10-23
Authentication improvments for 4.0

patrickarlt:
Covers https://github.com/Esri/arcgis-rest-js/issues?q=is%3Aopen+is%3Aissue+milestone%3Av4.0+assignee%3Apatrickarlt Each commit is a separate issue so this can easily reviewed. I plan on handling https://github.com/Esri/arcgis-rest-js/issues/893 in a separate PR.

Summary

* `UserSession` serialize/deserialize now includes the `server` property. #885
* Adds `UserSession.canRefresh` as a getter. #888
* Adds `style` to the oAuth 2.0 options for light/dark themes. #890
* Adds `revokeToken` helper and `UserSession.destroy()`. #892
* Adds a new `UserSession.signIn()` for username/password sign ins. #887
* Adds a new `UserSession.fromToken()` for generating a User Session from an existing token. #887
* `requestOptions.authentication` can now be a plain string. #889
* Rename auth classes for #914:
  * `UserSesison` > `ArcGISIdentityManager`
  * `ApplicationSession` > `ApplicationCredentialsManager`
  * `ApiKey` > `ApiKeyManager`
  Old classes have been deprecated but are still exported. We can remove them at 5.0.


packages/arcgis-rest-request/src/UserSession.ts
```diff
@@ -697,6 +698,17 @@ export class UserSession implements IAuthenticationManager {
     }
   }
 
+  /**
+   * Revokes all active tokens for a provided {@linkcode UserSession}. The can be considered the equivalent to signing the user out of your application.
+   */
+  public static destroy(session: UserSession) {
```
tomwayson: a `session.sigOut()` feels better to me than  a static `Session.destroy(session)`, especially as a compliment to `signIn()`. Although, maybe your reasoning was that `Session.signIn()` is just one way to create a session, and `Session.destroy()` works equally well for sessions created w/ `Session.beginOAuth2()`, etc.

This is not something I feel strongly about, and it works for me as is.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -239,6 +240,36 @@ export function request(
...
+    const rawToken = options.authentication;
+
+    authentication = {
+      portal: "https://www.arcgis.com/sharing/rest",
+      getToken: () => {
+        return Promise.resolve(rawToken);
+      }
+    };
+
+    /* istanbul ignore else - we don't need to test NOT warning people */
```
tomwayson, line 257: 👍

tomwayson (approved):
Epic work @patrickarlt 

Thank you for your patience. I'm sorry it took so long to review, but wanted to spend time and go through pretty thoroughly. Despite doing that, I don't have any real improvements to suggest.

The whole `UserSession` -> `ArcGISIdentityManager` is a bit jarring for me. I definitely expect that and other XyzManager classes to be able to juggle more than one set of credentials. That is likely b/c of my long history w/ this library and the ArcGIS API for Javascript.

That said, I don't think my feels on the names should hold anything up (any longer).

LGTM

packages/arcgis-rest-request/src/UserSession.ts
patrickarlt: @tomwayson this is absolutely what I was going for with `destroy` vs `signOut`

patrickarlt: @tomwayson I would be open to `session.signOut()` as an instance method.

PR #935, opened 2021-10-29
fix(arcgis-rest-auth): postMessage returns error if session is expired

dbouwman:
Add gating to the postMessage handler so it does not return the credential is the token is expired. Instead, it returns an error, that the Identity Manager can handle today (i.e. a JSAPI patch is not required)

A customer ran into this scenario, where they have a Hub site that embeds Dashboards _after_ the site loads. If a user loads the site, leaves the tab open until their session expires, and THEN they open the Dashboard, the Site hands the Dashboard (which is using JSAPI Identity Mgr) an expired credential. The Identity Manager does not expect an expired token so it infinitely loops hitting /generateToken... like _millions of times_ - which was noticed by the ops team, and eventually resulted in this PR

We need to land this quickly, so it can go out with the Hub release on Nov 2nd, 2021. We also need to ensure it gets into the 4.x code stream (I'll open a second PR to that branch)

### DIFF
--- a/packages/arcgis-rest-auth/src/UserSession.ts
+++ b/packages/arcgis-rest-auth/src/UserSession.ts
@@ -105,14 +105,14 @@ export interface IOAuth2Options {
   /**
    * The requested validity in minutes for a token. Defaults to 20160 (two weeks).
    */
-   expiration?: number;
+  expiration?: number;
 
   /**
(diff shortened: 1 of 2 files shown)


gavinr (approved):
looks good

tomwayson (approved):

PR #949, opened 2022-02-16
Test in Node 18 too.

patrickarlt:
Ideally I think we should support all LTS versions of Node.


patrickarlt:
Looks like GitHub actions don't support Node 18 yet.

PR #951, opened 2022-03-01
Rewrite oauth 2 functions to use PKCE

patrickarlt:
This adds support for dynamic client secrets with PKCE. This allows using `beginOAuth2()`/`completeOAuth2()` to get refresh tokens and use short lived tokens that get refreshed frequently.

Breaking changes:

* `state` is deprecated. We handle the state variable now.
* `completeOAuth2()` is now async and will return a Promise with the `UserSession` for inline redirects

Major changes:

* We now verify that `beginOAuth2()`/`completeOAuth2()` are called on the same client by saving a random string to local storage and checking it against the oauth 2 `state` param.
* Adds support for browser only tests. `beginOAuth2()`/`completeOAuth2()` now ONLY run in the browser and have 100% coverage.
* Remove the oauth 2 retry demo. This is almost identical to the main oauth 2 demo so I removed it and we can demo retrying in the main demo.
* The main demo ONLY does PKCE now since that is the recommended path now and figuring out how to make it do PKCE + implicit was difficult.

After this gets merged I'm going to:

* Do a follow up PR to automatically try to refresh tokens if we get an invalid token error.
* Add examples of retry handling to the main oauth 2 demo


dbouwman (approved):
I'm not deep into pkce details, but looks good and I tested out the oauth demo, and made some changes to show how to work against an Enterprise instance (or qaext)

demos/oauth2-browser/README.md
```diff
@@ -1,17 +1,25 @@
...
 See the ArcGIS Rest JS guide to [Authentication in Browser-based Apps](https://esri.github.io/arcgis-rest-js/guides/browser-authentication/) for more info.
 
 ## Running this demo
 
 1. Run `npm run build` in the root directory
-1. Register a new application using https://developers.arcgis.com/applications/new (you will be prompted to sign in or create a free account first).
-1. Add a redirect URL of `http://localhost:8080/authenticate.html` to your new app.
-1. Either copy the `config.js.template` file and rename it to `config.js` with your own ClientID or supply it in the prompt when the app is launched.
-1. Run `npm start` to spin up the development server.
+1. Run `npm start` in this directory to spin up the development server.
```
noahmulfinger, line 10: We still need a step here about copying the config:
```suggestion
1. Copy the config object from `config.js.template` into a new `config.js` file.
1. Run `npm start` in this directory to spin up the development server.
```

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -374,159 +339,313 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+            detail: {
+              ...oauthInfo
+            }
+          })
+        );
+
+        win.close();
+      }
+
+      win.history.replaceState(win.history.state, "", originalUrl);
```
noahmulfinger, line 563: Just for my understanding, what is the purpose of replacing the state with the original url? Doesn't the OAuth flow already return to the original page?

demos/oauth2-browser/README.md
patrickarlt: @noahmulfinger it is actually safe to reveal the client ID here so to make it easier for users I'm just going to commit `config.js` to the repo.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -374,159 +339,313 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+      if (popup && win.opener) {
+        win.opener.dispatchEvent(
+          new CustomEvent(`arcgis-rest-js-popup-auth-${clientId}`, {
+            detail: {
+              ...oauthInfo
+            }
+          })
+        );
+
+        win.close();
```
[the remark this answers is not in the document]
patrickarlt: added the return back in and updated tests.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: This was feedback from Dasa from the JS API team. The purpose is to make sure the code/token doesn't get left in the URL after successful auth. A user could potentially copy the token/code into an email or similar.

PR #953, opened 2022-03-03
Automatically refresh session and try again with new token on invalid token errors

patrickarlt:
This PR adds extra support for `Invalid token.` errors. When `request` encounters an invalid token it checks if:

1. The `authentication` option `canRefresh` and has `refreshSession`.
2. The error is `498: Invalid token.`

In this scenario we call `e.retry()` calling `refreshCredentials()` to exchange for a fresh token and retry the request with a new token. 

This should handle edge cases where the server thinks a token is invalid but ArcGIS REST JS thinks it is valid OR were ArcGIS REST JS doesn't have an expiration stored for the token.

In order to avoid edge cases where a server continuously responds with `Invalid token` (for example for a resource that a user doesn't have access too) we have an `internalRequest` function which this implements everything EXCEPT the automatic retry logic. `internalRequest` is what is used when we retry auth errors. The public  `request` function is just `internalRequest` with the retry logic.

---

This also renames `refreshSession` to `refreshCredentials` to align with the new naming conventions.

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -75,10 +75,12 @@ export class ArcGISAuthError extends ArcGISRequestError {
       code === "AUTHENTICATION_ERROR_CODE" ? message : `${code}: ${message}`;
   }
 
-  public retry(getSession: IRetryAuthError, retryLimit = 3) {
+  public retry(getSession: IRetryAuthError, retryLimit = 1) {
     let tries = 0;
 
     const retryRequest = (resolve: any, reject: any) => {
+      tries = tries + 1;
+
       getSession(this.url, this.options)
         .then((session) => {
           const newOptions = {
@@ -86,8 +88,7 @@ export class ArcGISAuthError extends ArcGISRequestError {
             ...{ authentication: session }
           };
 
-          tries = tries + 1;
-          return request(this.url, newOptions);
+          return internalRequest(this.url, newOptions);
         })
         .then((response) => {
           resolve(response);
@@ -95,7 +96,11 @@ export class ArcGISAuthError extends ArcGISRequestError {
         .catch((e) => {
           if (e.name === "ArcGISAuthError" && tries < retryLimit) {
             retryRequest(resolve, reject);
-          } else if (e.name === "ArcGISAuthError" && tries >= retryLimit) {
+          } else if (
+            e.name === this.name &&
+            e.message === this.message &&
+            tries >= retryLimit
+          ) {
             reject(this);
           } else {
             reject(e);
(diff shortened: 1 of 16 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -270,6 +262,10 @@ export function request(
     authentication = options.authentication;
   }
 
+  // for errors in GET requests we want the URL passed to the error to the the URL before
+  // query params are applied.
```
noahmulfinger, line 266:
```suggestion
  // for errors in GET requests we want the URL passed to the error to be the URL before
  // query params are applied.
```

packages/arcgis-rest-request/test/request.test.ts
```diff
@@ -560,4 +562,113 @@ describe("request()", () => {
...
+        httpMethod: "GET",
+        authentication: session
+      }).catch((e) => {
+        expect(e.code).toBe(498);
+        expect(e.message).toBe("498: Invalid refresh_token");
+        return;
+      });
+    });
+
+    it("should throw throw an error is it also fails with the new token", () => {
```
noahmulfinger, line 644:
```suggestion
    it("should throw an error if it also fails with the new token", () => {
```

packages/arcgis-rest-request/src/request.ts
```diff
@@ -444,3 +440,47 @@ export function request(
...
+      e instanceof ArcGISAuthError &&
+      e.code === 498 &&
+      e.message === "498: Invalid token." &&
+      requestOptions.authentication &&
+      (requestOptions.authentication as any).canRefresh &&
+      (requestOptions.authentication as any).refreshSession
+    ) {
+      return e.retry(() => {
+        return (requestOptions.authentication as any).refreshSession();
+      }, 1);
```
noahmulfinger, line 481: Since the default is already `1`, do we need to explicitly pass `1` as the second argument here?

packages/arcgis-rest-request/src/request.ts
```diff
@@ -444,3 +440,47 @@ export function request(
...
+  requestOptions: IRequestOptions = { params: { f: "json" } }
+): Promise<any> {
+  return internalRequest(url, requestOptions).catch((e) => {
+    if (
+      e instanceof ArcGISAuthError &&
+      e.code === 498 &&
+      e.message === "498: Invalid token." &&
+      requestOptions.authentication &&
+      (requestOptions.authentication as any).canRefresh &&
+      (requestOptions.authentication as any).refreshSession
```
noahmulfinger, line 477: Would make sense to add `canRefresh` and `refreshSession` as optional props to the IAuthenticationManager interface? So you wouldn't have to type these as `any` here

packages/arcgis-rest-request/src/request.ts
```diff
@@ -444,3 +440,47 @@ export function request(
...
+    ) {
+      return e.retry(() => {
+        return (requestOptions.authentication as any).refreshSession();
```
noahmulfinger, line 480: I don't see the rename from `refreshSession` to `refreshCredentials` in this PR.

noahmulfinger (changes requested):
Makes sense to me. I just had a few small suggestions and typo fixes.

patrickarlt:
@noahmulfinger this should be ready for a re-review now.

dbouwman (approved):
Looks good to me! Glad we're getting all this in for the 4.x release

PR #957, opened 2022-03-10
Publish raw typedoc json file w/ docs

dbouwman:
To enable the ArcGIS Developers site to pull raw typedoc from a stable location, we are now creating and publishing a "raw" (unprocessed) typedoc json file at the root of the docs site.

When we deprecate the acetate based doc site we will continue to publish this single file to the same location.


patrickarlt (changes requested):
@dbouwman at 4.0 I took a look at some new typedoc config. It would probably be easier to run the new `npm run typedoc` task and just adjust the output in `typedoc.json`

dbouwman:
@patrickarlt updated to have the `typedoc` script generate the json file directly in the docs/build/arcgis-rest-js folder, using the typedoc config.

patrickarlt (approved):

PR #959, opened 2022-03-14
Fix refresh token duration

patrickarlt:
This fixes the behavior around refresh token lifespans detailed in https://github.com/Esri/arcgis-rest-js/issues/699 and https://github.com/Esri/arcgis-rest-js/issues/842. In the next version on Online/Enterprise we will actually get a `refresh_token_expires_in` from the server so we can quit guessing at the token lifespan.

This PR totally removes `refreshTokenTTL` in favor of `expiration`. If you get a refresh token from a version of server that doesn't return `refresh_token_expires_in` we will just assume your refresh token is always valid.

If your server doesn't support `refresh_token_expires_in` the best thing to do would be to call the new `.exchangeRefreshToken()` method to update the refresh token every time the app loads. If this method errors then the refresh token is likely invalid. If it does not error the session was extended for 2 more weeks.

To get the absolute most long lived sessions possible you can just call `.exchangeRefreshToken()` every time the app loads to get another refresh token that lasts 2 more weeks. This will only invalidate the session when the user doesn't access the app for 2 weeks OR signs out (`.destroy()`).

### DIFF
--- a/demos/express/package.json
+++ b/demos/express/package.json
@@ -8,7 +8,7 @@
     "start": "node server.js"
   },
   "dependencies": {
-    "@esri/arcgis-rest-request": "^3.3.0",
+    "@esri/arcgis-rest-request": "file:../../packages/arcgis-rest-request",
     "express": "^4.16.3"
   },
   "author": ""
--- a/packages/arcgis-rest-feature-service/test/addToServiceDefinition.test.ts
+++ b/packages/arcgis-rest-feature-service/test/addToServiceDefinition.test.ts
@@ -30,7 +30,6 @@ describe("add to feature service", () => {
       tokenExpires: TOMORROW,
       refreshToken: "refreshToken",
       refreshTokenExpires: TOMORROW,
-      refreshTokenTTL: 1440,
       username: "casey",
       password: "123456",
       portal: "https://myorg.maps.arcgis.com/sharing/rest"
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -87,9 +87,9 @@ export interface IOAuth2Options {
   /**
    * The requested validity in minutes for a refresh token/access token. Defaults to 20160 (2 weeks).
    *
-   * When using PKCE or server based OAuth this will control the duration of the refresh token. In this scenario access tokens will always have a 30 minute validity.
+   * When using PKCE or server-based OAuth this will control the duration of the refresh token. In this scenario, access tokens will always have a 30 minute validity.
    *
-   * When using implict auth (`pkce: false`) in {@linkcode ArcGISIdentityManager.beginOAuth2} this controls the duration of the access token and no refresh token will be granted.
+   * When using implicit auth (`pkce: false`) in {@linkcode ArcGISIdentityManager.beginOAuth2}, this controls the duration of the access token and no refresh token will be granted.
    */
   expiration?: number;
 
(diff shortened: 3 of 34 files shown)


demos/express/package.json
```diff
@@ -8,7 +8,7 @@
     "start": "node server.js"
   },
   "dependencies": {
-    "@esri/arcgis-rest-request": "^3.3.0",
+    "@esri/arcgis-rest-request": "file:../../packages/arcgis-rest-request",
```
noahmulfinger, line 11: Is this change to `file:` intentional? We should make sure to standardize the import behavior across the demos.

packages/arcgis-rest-feature-service/test/addToServiceDefinition.test.ts
```diff
@@ -30,7 +30,6 @@ describe("add to feature service", () => {
       tokenExpires: TOMORROW,
       refreshToken: "refreshToken",
       refreshTokenExpires: TOMORROW,
-      refreshTokenTTL: 1440,
```
noahmulfinger, line 33: Not blocking, but considering this has to be changed in so many tests, we could switch to importing the MOCK_USER_SESSION from a shared file like `packages/arcgis-rest-portal/test/mocks/sharing/sharing.ts`

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -85,16 +85,13 @@ export interface IOAuth2Options {
...
   /**
-   * The requested validity in minutes for a token. Defaults to 20160 (two weeks).
-   */
-  expiration?: number;
-
-  /**
-   * Duration (in minutes) that a token will be valid. Defaults to 20160 (two weeks).
+   * The requested validity in minutes for a refresh token/access token. Defaults to 20160 (2 weeks).
+   *
+   * When using PKCE or server based OAuth this will control the duration of the refresh token. In this scenario access tokens will always have a 30 minute validity.
```
noahmulfinger, line 90:
```suggestion
   * When using PKCE or server-based OAuth this will control the duration of the refresh token. In this scenario, access tokens will always have a 30 minute validity.
```

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -85,16 +85,13 @@ export interface IOAuth2Options {
...
    *
-   * @deprecated use 'expiration' instead
+   * When using implict auth (`pkce: false`) in {@linkcode ArcGISIdentityManager.beginOAuth2} this controls the duration of the access token and no refresh token will be granted.
```
noahmulfinger, line 92:
```suggestion
   * When using implicit auth (`pkce: false`) in {@linkcode ArcGISIdentityManager.beginOAuth2}, this controls the duration of the access token and no refresh token will be granted.
```

noahmulfinger (approved):
Looks good to me, pending a few small comments.

demos/express/package.json
patrickarlt: We should standardize to `file:` because that is how NPM workspaces do it now if you want to test against the local packages. I'm going to do a follow up PR once I'm done with my auth work.

packages/arcgis-rest-feature-service/test/addToServiceDefinition.test.ts
patrickarlt: I agree that test standardization would be great but we have almost 600 tests now so it would be a lot of work 😄

PR #960, opened 2022-03-15
Server side state

patrickarlt:
This should resolve https://github.com/Esri/arcgis-rest-js/issues/660 allowing the `state` param to be passed through for server side auth so you can associate specific clients to specific auth states.

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -343,7 +343,6 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
         popup: true,
         popupWindowFeatures:
           "height=400,width=600,menubar=no,location=yes,resizable=yes,scrollbars=yes,status=yes",
-        state: options.clientId,
         locale: "",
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -709,15 +714,18 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
-    };
+    const { portal, clientId, expiration, redirectUri, state }: IOAuth2Options =
+      {
+        ...{ portal: "https://arcgis.com/sharing/rest", expiration: 20160 },
+        ...options
+      };
+
+    const url = `${portal}/oauth2/authorize?client_id=${clientId}&expiration=${expiration}&response_type=code&redirect_uri=${encodeURIComponent(
+      redirectUri
+    )}&state=${state}`;
```
noahmulfinger, line 725: Good candidate for using `encodeQueryString` here like we do in beginOAuth2.

noahmulfinger (approved):
This looks good to me. I had two small comments

Issue #796, opened 2021-02-04
Better UserSession Error Reporting


andrewctate:
Just spent the morning debugging a project and this `UserSession` ended up being the issue.
```js
const session = new UserSession({
  username: process.env.AGO_USERNAME,
  password: process.env.AGO_PASSWORD
});
```
Unbeknownst to me, the environment variables were undefined. The error I got when I tried to make a request was
```js
{
    "name": "ArcGISAuthError",
    "message": "Unable to refresh token.",
    "originalMessage": "Unable to refresh token.",
    "code": "AUTHENTICATION_ERROR_CODE"
}
```
I found this error misleading because
1. I was expecting to hear about fetching a token for the first time instead of "refreshing" a token.
1. It didn't tell me what was actually wrong (no `username` and `password`)

I'm not sure what to do about the first problem, but for the second maybe we could at least change the [current message](https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L941) to be something like "Unable to refresh token. You must provide either username and password or a clientId and refreshToken."

gavinr-maps:
@patrickarlt was going to follow up a fix with this one on #961

Issue #621, opened 2019-09-11
getTokenForServer fails to refresh expired tokens


oselz:
When a UserSession has a server with username and password it will only ever fetch one token, and then fails to update the token after it expires.  
```
const session = new UserSession({
    username: abc,
    password: def,
    server: my_enterprise_server_url,
    tokenDuration: 1,
})
request(someresource, {authentication: session})
--> works
(wait 1 minute)
request(someresource, {authentication: session})
--> fails (invalid token)
```

The reason is getTokenForServer checks for pending requests:

```
if (this._pendingTokenRequests[root]) {
      return this._pendingTokenRequests[root];
    }
```
however this check will _always_ be hit as the pending request is a Promise, thus will still exist as an object even after it is resolved.

The solution is to clear _pendingTokenRequests[root] when the promise resolves.

See https://github.com/Esri/arcgis-rest-js/pull/620

patrickarlt:
This will be resolved in the upcoming 4.0 release which is currently on the `beta` branch.

PR #620, opened 2019-09-11
fix(UserSession): will now update expired tokens on non-federated ser…

oselz:
…vers

AFFECTS PACKAGES:
@esri/arcgis-rest-auth

see #621

Issue #597, opened 2019-07-12
Set a specific error code when unable to refresh token


qlqllu:
When unable to refresh token, the user session just through a general `ArcGISAuthError` at this line: https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L775, it would be better if we can give it a error code.

patrickarlt:
@qlqllu Good catch. If you want to PR a change you could just add a second param to that line with whatever you wanted the error code to be. Something like `TOKEN_REFRESH_FAILED`

patrickarlt:
In 4.0 There is a new `ArcGISTokenRequestError` which will be thrown whenver a request for a token fails. THis is different from `ArcGISAuthError` which throws only for invalid token errors. More details will be forthcoming in the 4.0 release notes.

PR #961, opened 2022-03-16
Unable to refresh error

patrickarlt:
This PR completes https://github.com/Esri/arcgis-rest-js/issues/597 introducing a new `ArcGISTokenRequestError` which is thrown when there is an error generating or getting a token for a particular URL OR when a token refresh fails. There are different error codes for different scenarios (see doc in PR)

This also dug up some lingering issues from https://github.com/Esri/arcgis-rest-js/issues/161 / https://github.com/Esri/arcgis-rest-js/pull/233. In #233 John added support for refreshing a session if username and password was defined but not for using a refresh token. I ended up rewriting this and adding more tests so that we always get a fresh token before trying federation.

This also dug up some more skeletons that were a part of https://github.com/Esri/arcgis-rest-js/pull/620#pullrequestreview-334175193 / https://github.com/Esri/arcgis-rest-js/issues/621. The tests for this are a little odd and appear to be making multiple simultaneous requests as the tests were easy to fix by changing the `postOnce`/`getOnce` to `post`/`get`. I'm fairly sure this doesn't result in duplicate requests because if I change `canUseOnlineToken()` to always return `false` (thereby forcing everything to federate) I don't see any duplicate requests in the console just the `portal/self` (to get cross origin domains), server info, portal info, generate token, and the final request.

---

After this I'm going to make a follow up PR for https://github.com/Esri/arcgis-rest-js/issues/796 to add more error handling for common misktakes when creating sessions and another PR to add support for more error handling when doing server side oAuth.


packages/arcgis-rest-request/src/utils/ArcGISAccessDeniedError.ts
```diff
@@ -5,6 +5,33 @@
 // and https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work
 //
 // This code is from MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types.
+
+/**
+ * This error code will be thrown by the following methods when the user cancels or denys an authroization request on the OAuth 2.0
```
gavinr, line 10:
```suggestion
 * This error code will be thrown by the following methods when the user cancels or denies an authorization request on the OAuth 2.0
```

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1419,6 +1432,34 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+      params: {
+        token: this.token,
+        serverUrl,
+        expiration: this.tokenDuration
+      }
+    })
+      .then((response) => {
+        return {
+          token: response.token,
+          expires: new Date(response.expires - 1000 * 60 * 5)
```
gavinr, line 1449: Why do you subtract 5 minutes here?

packages/arcgis-rest-request/src/request.ts
```diff
@@ -52,6 +52,28 @@ export function getDefaultRequestOptions() {
   );
 }
 
+/**
+ * This error is thrown when a request encounters an invalid token error. Requests that use {@linkcode ArcGISIdentityManager} or
+ * {@linkcode ApplicationCredentialsManager} in the `authentication` option the authentication manager will automatically try to generate
+ * a fresh token using either {@linkcode ArcGISIdentityManager.refreshCredentials} or
+ * {@linkcode ApplicationCredentialsManager.refreshCredentials}. If the request with the new token fails you will receive an `ArcGISAuthError`
+ * if refreshing the token fails you will revice an instance of {@linkcode ArcGISTokenRequestError}.
```
gavinr, line 60:
```suggestion
 * if refreshing the token fails you will receive an instance of {@linkcode ArcGISTokenRequestError}.
```

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
```diff
@@ -0,0 +1,117 @@
...
+    code = ArcGISTokenRequestErrorCodes.UNKNOWN_ERROR_CODE,
+    response?: any,
+    url?: string,
+    options?: IRequestOptions
+  ) {
+    // 'Error' breaks prototype chain here
+    super(message);
+
+    // restore prototype chain, see https://stackoverflow.com/questions/41102060/typescript-extending-error-class
+    // we don't need to check for Object.setPrototypeOf as in the answers becasue we are ES2017 now
```
gavinr, line 105:
```suggestion
    // we don't need to check for Object.setPrototypeOf as in the answers because we are ES2017 now
```

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  public name: string;
+
+  /**
+   * Formatted error message. See the [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) class for more details.
+   */
+  public message: string;
+
+  /**
+   * The errror message return from the request.
```
gavinr, line 61:
```suggestion
   * The error message return from the request.
```

gavinr (commented):
Reading through the code, this looks good to me. I made a few spelling fixes. When I run `npm test` I'm getting an error - not sure if that's me or with this branch, I'll investigate it with you.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -356,8 +378,14 @@ export function internalRequest(
...
       };
 
+      // This should have the same conditional for Node JS as ArcGISIdentityManager.refreshWithUsernameAndPassword()
+      // to ensure that generated tokens have the same referer when used in Node with a username and password.
       /* istanbul ignore next - karma reports coverage on browser tests only */
-      if (typeof window === "undefined" && !fetchOptions.headers.referer) {
+      if (
+        (typeof window === "undefined" ||
+          (window && typeof window.document !== "undefined")) &&
+        !fetchOptions.headers.referer
```
noahmulfinger, line 387: Should the second part of the OR statement be inverted here? Won't this return true if `window.document` exists?

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
```diff
@@ -0,0 +1,117 @@
...
+import { IRequestOptions } from "./IRequestOptions.js";
+
+/**
+ * There are 5 potential error codes that might be thrown by {@linkcode ArcGISTokenRequestError}. 2 of these codes are used by both
+ * {@linkcode ArcGISIdentityManager} or {@linkcode ApplicationCredentialsManager}:
+ *
+ * * `TOKEN_REFRESH_FAILED` - When a request for an new access token fails.
+ * * `UNKNOWN_ERROR_CODE` - The error is unknown. More information may be available in {@linkcode ArcGISTokenRequestError.response}
+ *
+ * The 3 remaining error codes will inly be thrown when using {@linkcode ArcGISIdentityManager}:
```
noahmulfinger, line 13:
```suggestion
 * The 3 remaining error codes will only be thrown when using {@linkcode ArcGISIdentityManager}:
```

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
```diff
@@ -0,0 +1,117 @@
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
```
noahmulfinger, line 1:
```suggestion
/* Copyright (c) 2022 Environmental Systems Research Institute, Inc.
```

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export enum ArcGISTokenRequestErrorCodes {
+  TOKEN_REFRESH_FAILED = "TOKEN_REFRESH_FAILED",
+  GENERATE_TOKEN_FOR_SERVER_FAILED = "GENERATE_TOKEN_FOR_SERVER_FAILED",
+  REFRESH_TOKEN_EXCHANGE_FAILED = "REFRESH_TOKEN_EXCHANGE_FAILED",
+  NOT_FEDERATED = "NOT_FEDERATED",
+  UNKNOWN_ERROR_CODE = "UNKNOWN_ERROR_CODE"
+}
+
+// TypeScript 2.1 no longer allows you to extend built in types. See https://github.com/Microsoft/TypeScript/issues/12790#issuecomment-265981442
```
noahmulfinger, line 27: Regarding this comment, I'm a little confused because it seems like we are successfully extending the `Error` type below.

packages/arcgis-rest-request/test/utils/ArcGISTokenRequestError.test.ts
```diff
@@ -0,0 +1,50 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
```
noahmulfinger, line 1:
```suggestion
/* Copyright (c) 2022 Environmental Systems Research Institute, Inc.
```

package-lock.json
```diff
@@ -6233,52 +6233,6 @@
         "node": ">= 6.9.0"
       }
     },
-    "node_modules/@lerna/gitlab-client/node_modules/node-fetch": {
```
noahmulfinger, line 6236: I don't see any package changes, here, so I'm wondering if this package-lock.json change should be included?

packages/arcgis-rest-request/test/ArcGISIdentityManager.test.ts
```diff
@@ -831,9 +989,11 @@ describe("ArcGISIdentityManager", () => {
       expect(session.canRefresh).toBe(false);
 
       session.refreshCredentials().catch((e) => {
-        expect(e instanceof ArcGISAuthError).toBeTruthy();
-        expect(e.name).toBe("ArcGISAuthError");
-        expect(e.message).toBe("Unable to refresh token.");
+        expect(e instanceof ArcGISTokenRequestError).toBeTruthy();
```
noahmulfinger, line 992: Just a note here that we should make sure include error response changes in the list of breaking changes.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1454,20 +1495,49 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
-        password: this.password,
-        expiration: this.tokenDuration
-      },
-      ...requestOptions
+    const params = {
+      username: this.username,
+      password: this.password,
+      expiration: this.tokenDuration,
+      client: "referer",
+      referer:
```
noahmulfinger, line 1503: Just curious, are these newly added params expected by the rest/info call?

noahmulfinger (commented):
A few small comments/questions.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: It gives us some padding between when we think the token expires and when it actually expires. We probably could pad with less time but it lets us refresh the token early if the user is using the app heavily.

gavinr: sounds good and reasonable. Maybe adding that description in a comment here might be helpful to future devs to understand that? Or not, either way, I was just not fully understanding the "why" here.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: I don't think we were previously doing this right in `refreshWithUsernameAndPassword`. But I think this went unnoticed because we did it correctly in the `generateToken` function so we would always generate the token correctly. It was a little weird. The REST API for this is https://developers.arcgis.com/rest/users-groups-and-items/generate-token.htm

packages/arcgis-rest-request/src/request.ts
patrickarlt: Good catch will push a fix.

packages/arcgis-rest-request/src/utils/ArcGISTokenRequestError.ts
patrickarlt: Your right this comment is confusing because we do do this successful but we have to jump though hoops to maintain the prototype chain. I'm going to remove these comments and move them into the constructor for each error where they are close to the actual prototype chain code.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1454,20 +1495,49 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+    )
+      .then((response: any) => {
         this.updateToken(response.token, new Date(response.expires));
         return this;
-      }
-    );
+      })
+      .catch((e) => {
+        throw new ArcGISTokenRequestError(
+          e.messsage,
```
noahmulfinger, line 1534:
```suggestion
          e.message,
```

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1454,20 +1495,49 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
-        password: this.password,
-        expiration: this.tokenDuration
-      },
-      ...requestOptions
+    const params = {
+      username: this.username,
+      password: this.password,
+      expiration: this.tokenDuration,
+      referer: "client",
+      client:
```
noahmulfinger, line 1503: This looks backwards based on the generateToken docs: https://developers.arcgis.com/rest/users-groups-and-items/generate-token.htm

Should be `client: "referer"` and `referer: <url>`

noahmulfinger (changes requested):

gavinr (approved):

noahmulfinger (approved):

Issue #610, opened 2019-08-12
UserSession.fromCredential using incorrect url?


TravisMS1:
We have an in-house application that uses both the ArcGIS Javascript API and the ArcGIS REST JS API for connecting to a variety of services.  When trying to access a GeocodeServer secured with token-based security on an on-prem ArcGIS Server instance using the ArcGIS REST JS, our request fails due to an Auth error (which is expected since we haven't provided a token or credentials).  We prompt the user for a login using the JS API's IdentityManager signIn method to obtain a Credential object, and then we use that Credential object as a parameter to UserSession.fromCredential.  

When we retry the request using the new UserSession object, we're getting a NOT_FEDERATED error.  I believe this is due to the UserSession.fromCredential method constructing the portal url from the credential.server url.  In our case, that results in a portal url of https://&lt;host&gt;/server/sharing/rest instead of https://&lt;host&gt;/portal/sharing/rest

The GeocodeServer is at https://&lt;host&gt;/server/rest/services/&lt;GeocodeServer&gt;
The associated Portal is at https://&lt;host&gt;/portal

tomwayson:
> The associated Portal is at https://<host>/portal

Is that something that's specific to your environment?

If so, you should be able to just modify the session's portal after calling `fromCredential()`

TravisMS1:
I'm not aware of it being specific to our environment.  Our server instance is federated with our portal, and the GeocodeServer is secured.  Since the documentation for UserSession.fromCredential states 'Translates authentication from the format used in the ArcGIS API for JavaScript', and what we're providing is a Credential object returned from the JS API IdentityManager after a successful authentication, it seemed like it might be a limitation of the fromCredential method.  We can definitely just construct the ICredential manually.

Could there be Credential objects returned from the JS API that wouldn't be expected to work with fromCredential?  Also, maybe it's important to note that we are on JS API 3.29 - are there differences in the 3.x and 4.x IdentityManager that could be the culprit?

tomwayson:
I'm just not familiar enough w/ enterprise installations to know if portal is typically under `/portal` or that's an arbitrary choice that the administrator makes during installation.

> are there differences in the 3.x and 4.x IdentityManager that could be the culprit?

dunno

Here's the current logic we use:

https://github.com/Esri/arcgis-rest-js/blob/1635eb2c01c8096f53ec4097a72de91ba1ebbc5b/packages/arcgis-rest-auth/src/UserSession.ts#L519-L521

[...]

PR #965, opened 2022-03-21
fix(arcgis-rest-request): add support for server credentials

patrickarlt:
This fixes https://github.com/Esri/arcgis-rest-js/issues/610. The reason this broke is that `fromCredential` doesn't know if a given URL is a server credential or a portal URL. We might be able to use the actuall URL but the JS API normalizes Portal URLs removing the `sharing/rest` part. The JS API knows if a given URL is a server or a portal, but you have to call the `findServerInfo` method and pass the result to ArcGIS REST JS.

Since this changed the `fromCredential` method I also had to change a bunch of tests related to post message auth which @dbouwman  and @tomwayson  should look at.


packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -801,27 +814,48 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+    if (serverInfo.hasServer) {
+      return new ArcGISIdentityManager({
+        server: credential.server,
+        ssl,
+        token: credential.token,
+        username: credential.userId,
+        tokenExpires: new Date(expires)
+      });
+    }
+    console.log(credential.server.includes("sharing/rest"));
```
gavinr, line 852:
```suggestion
```

demos/jsapi-integration/index.html
```diff
@@ -79,11 +81,17 @@
...
-            });
+          // Once the web map loads the JS API should be "aware" of the ArcGIS.com portal.
+          // Using the findCredential and findServerInfo methods we can create an ArcGISIdentityManager
+          // for use in REST JS from the JS API
+          webmap.watch("loaded", function (loaded) {
+            if(loaded) {
+              const credential = esriId.findCredential("https://arcgis.com/sharing/rest/");
+              const serverInfo = esriId.findServerInfo("https://arcgis.com/sharing/rest/");
+              const session2 = arcgisRest.ArcGISIdentityManager.fromCredential(credential, serverInfo);
+              console.log(session2);
```
gavinr, line 92:
```suggestion
              console.log('Session from ArcGIS REST JS:', session2);
```

demos/jsapi-integration/index.html
```diff
@@ -62,14 +66,12 @@
           // JavaScript Identity Manager
           esriId.registerToken(session.toCredential());
 
-          // display the title of its basemap
-          console.debug(response);
-          console.debug('baseMap title:', response.baseMap.title);
+          console.debug(webMapItem);
```
gavinr, line 69:
```suggestion
          console.debug('Web map:', webMapItem);
```

gavinr (approved):
Looks good. A few small cleanup items.

demos/jsapi-integration/index.html
```diff
@@ -79,11 +81,17 @@
...
-            .then(cred => {
-              const session2 = arcgisRest.UserSession.fromCredential(cred);
-            });
+          // Once the web map loads the JS API should be "aware" of the ArcGIS.com portal.
+          // Using the findCredential and findServerInfo methods we can create an ArcGISIdentityManager
+          // for use in REST JS from the JS API
+          webmap.watch("loaded", function (loaded) {
+            if(loaded) {
+              const credential = esriId.findCredential("https://arcgis.com/sharing/rest/");
+              const serverInfo = esriId.findServerInfo("https://arcgis.com/sharing/rest/");
```
noahmulfinger, line 90: Two things for this demo:
1. Should we explain why loading specifically a web map is required vs any other item? (I assume because it ensures portal and server credentials)
2. The README is fairly out of date at this point and the code snippet doesn't match this new fromCredential logic.

noahmulfinger (approved):
Looks good to me. One comment about the demo readme that could be done in follow-up.

PR #967, opened 2022-03-21
chore: demo cleanup

patrickarlt:
This fixes all demos and sets their version to `beta`. It is worth noting that this will install the dependencies from NPM NOT the local workspace. Dependencies in the local workspace only install IF their version matches what is required in `package.json`. The `beta` tag resolves on the NPM respository and gives that back NOT the local version. To use local dependencies you need to configure the version in the package.json to the exact version.

This should generally be resolved when we release 4.0.0 and we can set all the dependencies to `^4.0.0` which should install them locally.


gavinr (approved):
looks good

tomwayson (approved):
LGTM

PR #969, opened 2022-03-22
chore: continue cleanup and deprecation of @esri/arcgis-rest-auth

patrickarlt:
This replaces almost all usage of `UserSession` and `ApplicationSession` in the docs and examples. This also removes the Vue example because I can't figure out how to make it work in our older version of Vue and also adds copies of the static methods on `ArcGISIdentityManager` to the deprecated `UserSession` object.


demos/batch-geocoder-node/batch-geocode.js
```diff
@@ -76,7 +76,7 @@ const mapResults = (results) =>
 
 // IMPLEMENTATION!
 
-// Instantiate an ApplicationSession to run Geocoding service
+// Instantiate an ApplicationCredentialsManager to run Geocoding service
 const session = ApplicationCredentialsManager.fromCredentials({
```
noahmulfinger, line 80: We still refer to the created objects as `session`s all over. I assume we would want to switch to terminology like `manager` or similar?

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1703,10 +1703,111 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+UserSession.fromCredential = function (
+  ...args: Parameters<typeof ArcGISIdentityManager.fromCredential>
+) {
+  console.log(
+    "DEPRECATED:, 'UserSession.fromCredential' is deprecated. Use 'ArcGISIdentityManager.fromCredential' instead."
+  );
+
+  console.warn(
+    "WARNING:, 'UserSession.fromCredential' now requires a `ServerInfo` object from the JS API as a second parameter."
+  );
```
noahmulfinger, line 1795: Looks like this warning will always appear even if the user passed in a ServerInfo object.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1703,10 +1703,111 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
   return new ArcGISIdentityManager(options);
 }
+
+/**
+ * @deprecated - Use {@linkcode ArcGISIdentityManager.beginOAuth2}.
+ */ /* istanbul ignore next */
+UserSession.beginOAuth2 = function (
+  ...args: Parameters<typeof ArcGISIdentityManager.beginOAuth2>
+) {
+  console.log(
```
noahmulfinger, line 1720: Should these deprecation messages be warnings instead of standard logs?

demos/batch-geocoder-node/batch-geocode.js
patrickarlt: I'll probably handle that in a follow up PR. I don't think it is as pressing though.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1703,10 +1703,112 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
...
+  ...args: Parameters<typeof ArcGISIdentityManager.completeOAuth2>
+) {
+  console.warn(
+    "DEPRECATED:, 'UserSession.completeOAuth2' is deprecated. Use 'ArcGISIdentityManager.completeOAuth2' instead."
+  );
+  if (args.length <= 1) {
+    console.warn(
+      "WARNING:, 'UserSession.completeOAuth2' is now async and returns a promise the resolves to an instance of `ArcGISIdentityManager`."
+    );
+  }
```
noahmulfinger, line 1740: @patrickarlt  this check for argument was added to the completeOAuth2 function instead of the fromCredential function.

PR #973, opened 2022-04-07
Only throw auth error on code 498/499

patrickarlt:
Thinking through the final auth changes there might be weird case where `ArcGISAuthError` was thrown but it wouldn't really make sense to retry the request. 

This PR changes things so that `ArcGISAuthError` ONLY fires on 498/499 errors which are the only codes the JS API pays attention to.

Then the automatic retry logic only applies on `ArcGISAuthError` and ignores the code.


gavinr (approved):
Looks good to me.

PR #987, opened 2022-05-04
update ApplicationCredentialsManager example

jf990:
user feedback the example on https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-request/ApplicationCredentialsManager does not work.


patrickarlt (approved):
LGTM!

PR #989, opened 2022-05-10
feat: peer and dev deps keep ranges

dbouwman:
- Update `devDependencies` and `peerDependencies` to use `^` ranges for in-repo packages
- change release to use `--deps.bump=inherit` and `--deps.release=inherit` which will leave the `^` ranges in the deps intact.

**NOTE**: There is a bug in the backing release package where a `major` release will _not_ maintain the `^`. There is a PR being discussed to address this, and given we don't do breaking changes often, this seems to be a reasonable solution in the near-term.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
     "prettify": "prettier --write --parser typescript --tab-width 2 --use-tabs false \"packages/**/+(src|test)/**/*.ts\"",
     "typedoc": "typedoc --options ./typedoc.json",
     "prettify:packages-jsons": "prettier-package-json --write ./**/*/package.json",
-    "release:dry": "multi-semantic-release --dry-run --debug --deps.release=inherit --ignore-private-packages",
-    "release": "multi-semantic-release  --deps.release=inherid --ignore-private-packages"
+    "release:dry": "multi-semantic-release --dry-run --deps-bump=inherit --deps.release=inherit --ignore-private-packages --debug ",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -34,8 +34,8 @@
     "prettify": "prettier --write --parser typescript --tab-width 2 --use-tabs false \"packages/**/+(src|test)/**/*.ts\"",
     "typedoc": "typedoc --options ./typedoc.json",
     "prettify:packages-jsons": "prettier-package-json --write ./**/*/package.json",
-    "release:dry": "multi-semantic-release --dry-run --debug --deps.release=inherit --ignore-private-packages",
-    "release": "multi-semantic-release  --deps.release=inherid --ignore-private-packages"
+    "release:dry": "multi-semantic-release --dry-run --deps-bump=inherit --deps.release=inherit --ignore-private-packages --debug ",
+    "release": "multi-semantic-release --deps-bump=inherit --deps.release=inherit --ignore-private-packages"
```
tomwayson, line 38: Does `deps-bump` work? From https://github.com/dhoulb/multi-semantic-release#cli it looks like it should be  `deps.bump` like `deps.release`?

dbouwman: oh fml - thanks for the sharp eyes!

dbouwman: Updated

tomwayson (approved):
thanks @dbouwman

PR #996, opened 2022-05-23
fix: add bbox

drspacemanphd:
Add bbox to list of options for generic search


dbouwman:
Closing as this can be accomplished via `searchOptions.params.bbox`

PR #1014, opened 2022-08-24
Add 'id' property to IUser type definition

anumysore5:
Update user.ts to add 'id' property to IUser type definition


patrickarlt (approved):
LGTM. Thanks @anumysore5!

Issue #995, opened 2022-05-23
4.x redirectUri does not support query param


qlqllu:
When we call `beginOAuth2` with the redirectUri that has query param, after call `completeOAuth2`, it reports `invalid redirect uri` error. After dig around, we find the token endpoint needs the redirect uri to be the same with the redirct uri passed in the `beginOAuth2`, but [this line](https://github.com/Esri/arcgis-rest-js/blob/%40esri/arcgis-rest-request%404.0.3/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L628) remove the query params, which causes the above error I think.

patrickarlt:
As far as I can tell the spec does allow for URLs with query params so we need to respect that and that means this is a bug.

PR #1015, opened 2022-08-25
fix: #995 and #1006, issues with redirect urls

### LINKED ISSUE #1006: 4.x ArcGISIdentityManager does not have its redirectUri property set
I don't think this is exactly the same issue as #995 (apologies if it is). The other issue seems to have more to do with calling `completeOAuth2` with query parameters on the `redirect_uri`, but in this case `completeOAuth2` otherwise works fine. 

After `completeOAuth2` returns a valid `ArcGISIdentityManager`, it does not have its `redirect_uri` property set (and it is marked as a read-only property in the TypeScript definition file, so it cannot be manually set without having to ignore the type error).

Then, when a token expires (or it is invalidated), it does automatically call the token endpoint for a `exchange_refresh_token`, but it doesn't have a `redirect_uri` so the endpoint returns a 400 error (`redirect_uri not specified`). I'm not sure if this is intended design or if we are supposed to be exchanging a token manually.

It's also still possible to set the `redirect_uri` property manually (and ignoring the type errors from modifying a readonly property), which does work and automatically calls the token endpoint with that `redirect_uri` if the token is ever found to be invalid.

patrickarlt:
This fixes both #995 and #1006.

The root cause of #995 was assuming that query params could not be included in oAuth redirect URIs, when in fact they can be. Any query params in the redirect URI are expected to be placed back on the redirect URI in addition to any oauth query params like `code` and `error`. To resolve this you now must pass `redirectUri` in `completeOAuth2()` like you did in 3.x. This was still documented that was but some demo code committed it. I left the existing code path in for people who leave out `redirectUri`.

To resolve #1006 we now properly set `redirectUri` on the instance of `ArcGISIdentityManager` in `beginOAuth2()`, `completeOauth2()` and `fromToken()` when it is provided. `canRefresh()` now also checks for `redirectUri` since if it isn't defined you cannot exchange refresh tokens.

This also fixes a bug where if you did pass an invalid redirect URI or client id to `completeOauth2()` in a popup it would not register it as an error.

The changes to the browser oauth demo are just prettier + adding the redirect URL to the popup page.

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -27,11 +27,34 @@ import { NODEJS_DEFAULT_REFERER_HEADER } from "./index.js";
  * Options for {@linkcode ArcGISIdentityManager.fromToken}.
  */
 export interface IFromTokenOptions {
+  /**
+   * The token you want to create the {@linkcode ArcGISIdentityManager} instance with.
+   */
   token: string;
+  /**
+   * Date when this token will expire.
+   */
   tokenExpires?: Date;
+  /**
+   * The portal that the token was generated from. Defaults to `https://www.arcgis.com/sharing/rest`. Required if you are not using the default portal.
+   */
   portal?: string;
+  /**
+   * If the token is for a specific instance of ArcGIS Server, set `portal` to `null` or `undefined` and set `server` the URL of the ArcGIS Server.
+   */
   server?: string;
+  /**
+   * Optionally set the username. Recommended if available.
+   */
   username?: string;
+  /**
+   * Optional client ID. Used for refreshing expired tokens.
+   */
+  clientId?: string;
+  /**
+   * Optional set a valid redirect URL for the registered client ID. Used internally to refresh expired tokens.
+   */
+  redirectUri?: string;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


dbouwman (approved):
I'm no guru on how oauth is implemented in this library, but this seems like a good fix

tomwayson (approved):
Nice work @patrickarlt. 

Should I start an issue for v5 deprecations and include "support for calling `completeOAuth2()` w/o a `redirectUri`"?

gavinr (approved):

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -435,7 +458,7 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
       if (popup) {
         // If we are authenticating a popup we need to return a Promise that will resolve to an ArcGISIdentityManager later.
         return new Promise((resolve, reject) => {
-          // Add an event listener to listen for when a user calls `ArcGISIdentityManager.completeOAuth2()` in the popup.
+          // Add an event listener to listen for when a user calls `ArcGISIdentityManager.completeOAuth2()()` in the popup.
```
patrickarlt, line 461:
```suggestion
          // Add an event listener to listen for when a user calls `ArcGISIdentityManager.completeOAuth2()` in the popup.
```

PR #1017, opened 2022-09-01
fix(arcgis-rest-portal): add supported search fields

gavinr:
(in progress) Copying the changes from #928 into the new v4 (current) branch

### DIFF
--- a/packages/arcgis-rest-portal/src/util/search.ts
+++ b/packages/arcgis-rest-portal/src/util/search.ts
@@ -14,6 +14,22 @@ export interface ISearchOptions extends IRequestOptions, IPagingParams {
 
   /** Describes whether order returns in ascending or descending order. The default is ascending. */
   sortOrder?: "asc" | "desc";
+
+  /** A comma-separated list of fields to count. The maximum count fields allowed per request is three. Supported count fields are tags, type, access, contentstatus, and categories. */
+  countFields?: string;
+
+  /** The maximum number of field values to count for each `countFields`. The default value is 10, and the maximum number allowed is 200. */
+  countSize?: number;
+
+  /** Structured filtering is accomplished by specifying a field name followed by a colon and the term you are searching for with double quotation marks. It allows the passing in of application-level filters based on the context. Use an exact keyword match of the expected value for the specified field. Partially matching the filter keyword will not return meaningful results. */
+  filter?: string;
+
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-portal/src/util/search.ts
```diff
@@ -14,6 +14,22 @@ export interface ISearchOptions extends IRequestOptions, IPagingParams {
...
+  countFields?: string;
+
+  /** The maximum number of field values to count for each `countFields`. The default value is 10, and the maximum number allowed is 200. */
+  countSize?: number;
+
+  /** Structured filtering is accomplished by specifying a field name followed by a colon and the term you are searching for with double quotation marks. It allows the passing in of application-level filters based on the context. Use an exact keyword match of the expected value for the specified field. Partially matching the filter keyword will not return meaningful results. */
+  filter?: string;
+
+  /** A JSON array or comma-separated list of up to eight org content categories to search items. */
+  categories?: string;
```
patrickarlt, line 28: @gavinr do we want to type this as an actually json array or comma separated string or all 3?
```suggestion
  categories?: string | Array<string>;
```

patrickarlt (changes requested):

packages/arcgis-rest-portal/src/util/search.ts
gavinr: Yes I think that's a good suggestion. But then we need to make sure we cast the string array back to a string, per the doc:

image

https://developers.arcgis.com/rest/users-groups-and-items/search.htm#GUID-1C625F8A-4A12-45BB-B537-74C82315759A

I'll make that modification now.

gavinr: In the case that a user passes in an `Array<string>` (ex: `["/Region/US","/Forest"]`), which of these strings should we convert it to? 

- `categories=/Region/US,/Forest`
- `categories: ["/Region/US","/Forest"]`

I guess it does not matter that much ... ? I'm not sure what typically REST-JS would do in other cases similar to this situation.

gavinr: Updated!

patrickarlt (commented):
LGTM!

patrickarlt (approved):
LGTM!

PR #1018, opened 2022-09-01
Job Task in Arcgis-Rest-Request

marcysilverman:
- Creates a new Job Class in arcgis-rest-request for async job tasks
- Creates a job class that allows you submit job, retrieve job info, retrieve results, as well as individual result properties.
- Also has an event emitter that allows users to check the status based on the jobStatus of the job task.


packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,249 @@
...
+  //   //uses key value pairs for the results obj to get each result property to do their own request
+  //   // Promise.all will do all the individual requests and wait until their all done
+  //   // return a object with all the individual result property
+  // };
+
+  cancelJob() {
+    return request(this.jobUrl + "/cancel", {
+      authentication: this.authentication,
+      params: { jobId: this.jobId, returnMessages: false }
+    }).then((response) => this.emitter.emit("cancelled", response));
```
patrickarlt, line 221: @marcysilverman this will return the results of `this.emitter.emit` to the next promise. This should be
```suggestion
    }).then((response) => {
      this.emitter.emit("cancelled", response));
      return response;
    }
```

The test for this should probably also check that the response is passed through.

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,249 @@
...
+  static cancelJobRequest: string;
+
+  emitter: any;
+
+  private didUserEnableMonitoring: any;
+  private setIntervalHandler: any;
+
+  params: any;
+
+  constructor(options: any) {
```
patrickarlt, line 36: @marcysilverman  we should probably make this a separate interface for when people do want to construct a `Job` manually, like `IJobOptions`.

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  get isMonitoring() {
+    return !!this.setIntervalHandler;
+  }
+
+  fromExistingJob() {
+    return request(this.jobUrl, {
+      authentication: this.authentication
+    });
+  }
```
patrickarlt, line 58: @marcysilverman this should be a static method on `Job` you would use it like so:

```js
// user starts a job
const job = await submitJob(/* some options */)

// save the job info to local storage so we can look up the info later:
localStorage.set("activeJob", JSON.stringify(
  {
    id: job.id,
    url: job.url
  }
));

// later when the user refreshes the page, check for an active job and create it again:
const {id, url} = JSON.parse(localStorage.get("activeJob"))

const job = await Job.fromJobId({
  id,
  url
})
```

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  emitter: any;
+
+  private didUserEnableMonitoring: any;
+  private setIntervalHandler: any;
+
+  params: any;
+
+  constructor(options: any) {
+    this.params = options.requestOptions;
```
patrickarlt, line 37: This isn't used anywhere in the class so it probably isn't necessary.
```suggestion
```

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,249 @@
...
+    this.url = options.url; //user passes in the initial endpoint
+    this.jobId = options.jobId; //saved from the response from the static create job
+    this.pollingRate = options.pollingRate || 5000;
+    this.emitter = mitt(); //interval between each polling request
+    this.jobUrl = this.url.replace("submitJob", `jobs/${this.jobId}`);
```
patrickarlt, line 42: This could probally be converted to a `get` like `isMonitoring()`

demos/job/index.ts
```diff
@@ -0,0 +1,278 @@
...
+      useHierarchy: true,
+      uturnAtJunctions: "esriNFSBNoBacktrack"
+    },
+    populate_directions: true,
+    directions_language: "en",
+    populate_route_lines: false,
+    default_date: "1608051600000"
+  },
+  authentication:
+  "AAPK59244520da284b68bc5d426a1c22d08fp9HQo6v2E2xWsNgC39us1eQQvnrCLxBEYUAI_qR6xaHgU-S7osfaONVFd0YEi9dH",
```
patrickarlt, line 263: @marcysilverman please delete the API key we should be careful about commiting them to repos. I would follow some of the other examples in demo and load a config file or use https://www.npmjs.com/package/dotenv with a gitignored config file so users can set their own API key.

It would also be good to document this int he readme.

patrickarlt (changes requested):
@marcysilverman A few things:

* Can you restore `packages/arcgis-rest-portal/test/groups/add-users.test.ts` it looks like it was accidentally deleted.
* Update your demo readme to contain info more specific to the job
* It looks like `getAllResults()` is commented out. Do you want to implement it?
* It looks like your missing `job.toJSON()`, `job.serialize()` and `Job.deserialize()` methods, can you implement those?
* I think we should use `job.id` and `job.url` instead of `job.jobId` and `job.jobUrl`

I also don't see any of the handling of URls that we agreed upon:

1. `job.submitJob()` should handle the URL both with and without `/submitJob` the URL should be cleaned up and saved so that `job.url` is the version WITHOUT `/submitJob`.
2. When you need to make requests get job info, results, ect... you can concatenate `job.id` to `job.url`

demos/job/readme.md
```diff
@@ -0,0 +1,13 @@
+# Job Class Demo
```
patrickarlt, line 1: @marcysilverman I would add a note about the `.env` file here and how to set it up and that you need PAYG enabled on your devleoper account to use it.

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,271 @@
...
+      startMonitoring: this.isMonitoring,
+      pollingRate: this.pollingRate
+    }
+  }
+
+  serialize() {
+    return JSON.stringify(this)
+  }
+
+  static deserialize(serializeString: string) {
```
patrickarlt, line 175: @marcysilverman we can discuss this tomorrow but I think this needs a second param for `authentication`, users should serialize and deserialize jobs and authentication separately.

packages/arcgis-rest-request/src/job.ts
```diff
@@ -0,0 +1,271 @@
...
+import { IRequestOptions, JOB_STATUSES } from "./index.js";
+import mitt from "mitt";
+
+interface IJobOptions extends IRequestOptions {
+  id?: string;
+  url?: string;
+  params?: any;
+  startMonitoring?: boolean;
+  pollingRate?: number;
+  token?: string;
```
patrickarlt, line 12: This shouldn't take a `token` it should take `authentication`

packages/arcgis-rest-request/test/job.test.ts
```diff
@@ -0,0 +1,627 @@
...
+          .then((res) => {
+            done();
+          }, function (error) {
+            expect(error).toEqual({
+              jobId: "j4fa1db2338f042a19eb68856afabc27e",
+              jobStatus: "esriJobTimeOut"
+            });
+            done();
+          }
+          );
```
patrickarlt, line 217: @marcysilverman lets go over this portion tomorrow you probally want to `catch` here instead of using 2 arguments to `then`

patrickarlt (changes requested):
@marcysilverman I think this will be good code wise after our review tomorrow:

1. Cleanup some tests a bit
2. Cleanup the creation methods to allow passing `authentication`
3. Fix merge conflicts and test
4. Remaining .5% coverage
5. Doc strings for the API ref

I have a branch `add-job-support` pushed up that I think we can work off of.

patrickarlt:
Closing this in favor of #1020.

PR #1019, opened 2022-09-09
Add issue templates

gavinr:
Adds issue templates so that when someone clicks "new issue" in GitHub they will be presented with options something like this:
image


tomwayson (approved):
thanks @gavinr

PR #1024, opened 2022-09-21
bug: create/update items convert extent to string frm bbox

### LINKED ISSUE #1023: processParams not converting extent to string
### Describe the bug

`addItem` and `updateItem` both expect `extent` to be a flattened string instead of a bbox ( `1, 2, 3, 4` vs `[[1, 2], [3, 4]]` ). Apparently around two years ago we made a change to `processParams` ( #779 ) for catalog to allow array of arrays through. At that time it was decided to do that for all the cases so as to not make an explicit reference to a property. Unfortunately that has broken `extent`. What ends up happening is in the actual request the bbox is split into two different extent params, the create/update call succeeds, however the item ends up with an extent which is simply an empty array.

### Reproduction

Will be provided

### Logs

_No response_

### System Info

```shell
v3.4.3
```

### Additional Information

_No response_

benstoltz:
Resolves #1023

### DIFF
--- a/packages/arcgis-rest-portal/src/items/create.ts
+++ b/packages/arcgis-rest-portal/src/items/create.ts
@@ -101,6 +101,9 @@ export function createItemInFolder(
     };
 
     // convert extent, if present, into a string from bbox
+    // processParams was previously doing this sort of work,
+    // however now we need to let array of arrays through
+    // Thus for extents we need to move this logic here
     if (requestOptions.params.extent && isBBox(requestOptions.params.extent)) {
       requestOptions.params.extent = bboxToString(requestOptions.params.extent);
     }
--- a/packages/arcgis-rest-portal/src/items/create.ts
+++ b/packages/arcgis-rest-portal/src/items/create.ts
@@ -10,7 +10,9 @@ import {
   IUpdateItemResponse,
   ICreateUpdateItemOptions,
   serializeItem,
-  determineOwner
+  determineOwner,
+  isBBox,
+  bboxToString
 } from "./helpers";
 
 export interface ICreateFolderOptions extends ICreateUpdateItemOptions {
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-portal/src/items/create.ts
```diff
@@ -98,6 +100,11 @@ export function createItemInFolder(
       ...serializeItem(requestOptions.item)
     };
 
+    // convert extent, if present, into a string from bbox
```
dbouwman, line 103: I'd have done this a little higher so it's clear we're working with the `item.extent`... this is functionally correct though

packages/arcgis-rest-portal/src/items/helpers.ts
```diff
@@ -290,3 +290,26 @@ export function determineOwner(requestOptions: any): Promise<string> {
...
+};
+
+/**
+ * Given a Bbox, convert it to a string. Some api endpoints expect a string
+ *
+ * @param {BBox} extent
+ * @return {*}  {string}
+ */
+ export function bboxToString (extent: number[][]): string {
+  return extent.map((a) => a.join(", ")).join(", ");
```
dbouwman, line 314: while I'm sure this works, it seems that `[[1,2],[3,4]].join(',')` returns `1,2,3,4` and I think that's the old implementation in process params.

dbouwman (approved):
This is good with me - maybe add a comment where we're doing `.params.extent` just to inform the next dev what's going on.

packages/arcgis-rest-portal/src/items/create.ts
benstoltz: Problem with that is doing it higher means manipulating an IItemAdd, which expects that extent is a `number[][]` which means doing the whole `as unknown as number[][]`.

packages/arcgis-rest-portal/src/items/helpers.ts
benstoltz: Ah, I was going off the docs which show the output like: `1, 2, 3, 4` (`extent=<xmin>, <ymin>, <xmax>, <ymax>` ; see [addItem](https://developers.arcgis.com/rest/users-groups-and-items/add-item.htm))

patrickarlt (approved):
LGTM!

PR #1027, opened 2022-10-04
Added Params Preprocessor for GP Mutilvalue Inputs

marcysilverman:
In `static submitJob`, the params passed in will be processed to check if there is a top level array, if so, `JSON.stringify(Array)`. 

Any MultiValue inputs can only be read as a JSON array. We do this before hitting the REST endpoint. Doubled checked with Mark Torrey with an example of his that this does in fact work.

### DIFF
--- a/packages/arcgis-rest-request/src/job.ts
+++ b/packages/arcgis-rest-request/src/job.ts
@@ -45,6 +45,11 @@ export interface ISubmitJobOptions {
   params: any;
 
   /**
+   * Processes arrays to JSON strings for Geoprocessing services. See “GPMultiValue” in {@link https://developers.arcgis.com/rest/services-reference/enterprise/gp-data-types.htm}
+   */
+  processedParams: any;
+
+  /**
    * The base URL of the job without `/submitJob` or a trailing job id.
    */
   url: string;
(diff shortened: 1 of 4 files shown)


packages/arcgis-rest-request/src/job.ts
```diff
@@ -188,10 +188,19 @@ export class Job {
...
 
+    /**
+     * Processes arrays to JSON strings for Geoprocessing services. See “GPMultiValue” in {@link https://developers.arcgis.com/rest/services-reference/enterprise/gp-data-types.htm}
+     */
+    const processedParams = Object.keys(params).reduce((newParams:any, key)=> {
+    const value = params[key]
+    const type = value.constructor.name;
+    newParams[key] = type === "Array" ? JSON.stringify(value) : value;
+    return newParams;
+    }, {});
```
patrickarlt, line 199: I would move this logic into a separate function like `processJobParams` and write tests specifically for this behavior. Right now this is getting covered but we are not verifying the results with a test.

patrickarlt (changes requested):
@marcysilverman looks good but I have a note about a test.

packages/arcgis-rest-request/src/job.ts
```diff
@@ -169,7 +169,20 @@ export class Job {
...
+  static processedParamsFunc(params:any) {
+    const processedParams = Object.keys(params).reduce((newParams:any, key)=> {
+      const value = params[key]
+      const type = value.constructor.name;
+      newParams[key] = type === "Array" ? JSON.stringify(value) : value;
+      return newParams;
+      }, {});
+      
+      return processedParams
+  }
```
patrickarlt, line 185: @marcysilverman can you move this to a separate file in `utils`? I'm a little distrustful of calling this like `this.processedParamsFunc`? Probably something like `utils/process-job-params.ts` and export it as `processJobParams`

Issue #1033, opened 2022-10-21
Unable to upload item data with addItemData()


haoliangyu:
### Describe the bug

I am trying to use the `addItemData` function in node 16 to upload file data to an already created item. It doesn't throw any error but the file is actually not uploaded. When I try to download the item file, it shows a `Item does not have a file.` error page.

This bug happens at rest-js 4.x but not in 2.x or 3.x.

### Reproduction

You can reproduce the problem by running the repository https://github.com/haoliangyu/arcgis-rest-js-node-16-bug. You can run `volta run yarn` and `volta run node index.js` to  reproduce the problem.

### Logs

```shell
There is no error log.
```

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 581.50 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.volta/tools/image/node/16.18.0/bin/node
    Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 8.19.2 - ~/.volta/tools/image/npm/8.19.2/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Safari: 15.6.1
  npmPackages:
    @esri/arcgis-rest-request: ^4.0.0 => 4.1.1
```

haoliangyu:
At rest-js 2.x and 3.x, there is another problem with the `addItemData` function. It may cause the `socket hang up` error in the update request. I have seen a few issues related to the `form-data` library and nodejs Readable stream: https://github.com/form-data/form-data/issues/520, https://github.com/form-data/form-data/issues/503, https://github.com/form-data/form-data/issues/508. But since rest-js 2.x and 3.x are no longer supported, I only submit the issue for 4.x.

PR #1034, opened 2022-10-29
fix(arcgis-rest-request): node ReadStream encoding bug at 3.x

haoliangyu:
This PR addresses the issue https://github.com/Esri/arcgis-rest-js/issues/1033 at rest-js v3. The bug is related to the [form-data](https://github.com/form-data/form-data) library in the node environment. It requires to use an unofficial option to specify the stream length so that the streaming request can end properly (see https://github.com/Esri/arcgis-rest-js/issues/1033#issuecomment-1292016958).

I add a `dataSize` option to the `addItemData()` function so that the user can give a number for the stream data length.

PS: I suspect other functions accepting node ReadStream can have the same issue, but I don't have the availability to verify and fix each API.

PS2: I assume that no more than one stream is used in one API call.

PS3: I have to update the npm version to v7 in CI so that `npm install` won't throw an error.

### DIFF
--- a/.github/workflows/pre-release-tests.yml
+++ b/.github/workflows/pre-release-tests.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
+      - name: Update to NPM 7 for workspace support
+        run: npm install -g npm@7
+
       - name: Install Dependencies
         run: npm install
 
--- a/.github/workflows/pull-request-tests.yml
+++ b/.github/workflows/pull-request-tests.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
+      - name: Update to NPM 7 for workspace support
+        run: npm install -g npm@7
+
       - name: Install Dependencies
         run: npm install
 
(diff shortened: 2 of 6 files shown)


haoliangyu:
The CI failed at the Chrome test stage because that required 100% test coverage for the code. But this change is for node environment. Is there a way to bypass the test coverage check for a certain block of code?

tomwayson (approved):
Nice work @haoliangyu 

I _think_ when this is merged it will trigger a 3.x patch release, but I'm not certain. Let's wait until @dbouwman confirms before merging.

dbouwman:
For rest-js v3, we did not have semantic-release, so you actually want to PR this to `master` and then we manually run the release like we did in the olden times. 

[This PR](https://github.com/Esri/arcgis-rest-js/pull/1024) is the last time we did a 3.x release. 

I can run the scripts if you want. Just lmk via teams once this is merged.

tomwayson:
Thanks @dbouwman, yeah, I just noticed that we don't have  [a .releaserc.json](https://github.com/Esri/hub-components/blob/f46db66efbd29edeee6cfff771b405a4341d1be0/.releaserc.json) file that I believe would be required to run the release on [maintenance branches](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#maintenance-branches).

I've updated the base, and will merge and run the scripts.

haoliangyu:
@tomwayson Do you have any availability to release this at v3.5.1?

tomwayson:
@haoliangyu this has been published as v.3.5.1

PR #1038, opened 2022-11-18
fix: add error parameter to IEditFeatureResult

samMatenaer:
the applyEdit results include an error object when success === false

see the spec here: https://developers.arcgis.com/rest/services-reference/enterprise/edit-result-object.htm

### DIFF
--- a/packages/arcgis-rest-feature-service/src/helpers.ts
+++ b/packages/arcgis-rest-feature-service/src/helpers.ts
@@ -50,24 +50,23 @@ export interface ISharedQueryOptions extends IGetLayerOptions {
 /**
  * Add, update and delete features result Object.
  */
-export type IEditFeatureResult = {
+export interface IEditFeatureResult {
   objectId: number;
   globalId?: string;
-  success: true;
-} | {
-  objectId: number;
-  globalId?: string;
-  success: false;
-  error: {
+  success: boolean;
+  /**
+   * Error is optional and is only returned when `success` is `false`.
+   */
+  error?: {
     code: number;
     description: string;
-  };
+  }
 }
 
 /**
  * `globalId` always returned with attachments via apply edits.
  */
-export type IApplyEditsAttachmentResult = IEditFeatureResult & {
+export interface IApplyEditsAttachmentResult extends IEditFeatureResult {
   globalId: string;
 }
 
--- a/packages/arcgis-rest-feature-service/src/helpers.ts
+++ b/packages/arcgis-rest-feature-service/src/helpers.ts
@@ -50,16 +50,24 @@ export interface ISharedQueryOptions extends IGetLayerOptions {
 /**
  * Add, update and delete features result Object.
  */
-export interface IEditFeatureResult {
+export type IEditFeatureResult = {
   objectId: number;
   globalId?: string;
-  success: boolean;
+  success: true;
+} | {
+  objectId: number;
+  globalId?: string;
+  success: false;
+  error: {
+    code: number;
+    description: string;
+  };
 }
 
 /**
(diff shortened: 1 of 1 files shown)


gavinr:
Thanks Sam - adding to the type (` | { ... error: { ... }}`) makes sense to me.

Regarding switching the interfaces to types: was that intentional and required for this PR?

samMatenaer:
@gavinr it was the only way i could figure out how to make a conditional type. i wasn't sure if there was a way to do it with interfaces. then, by changing the first type to a `type`, i had an error come up that required changing the inherited interfaces to `type` as well

patrickarlt (commented):
@gavinr @samMatenaer I think this would be best as

```ts
/**
 * Add, update and delete features result Object.
 */
export interface IEditFeatureResult {
  objectId: number;
  globalId?: string;
  success: boolean;
  /*
   * Error is optional and is only returned when `success` is `false`.
   */
  error? : {
    code: number;
    description: string;
  }
} 

/**
 * `globalId` always returned with attachments via apply edits.
 */
export interface IApplyEditsAttachmentResult extends IEditFeatureResult {
  globalId: string;
}
```

You can see this work in the [TypeScript playground](https://www.typescriptlang.org/play?#code/PQKhFgCgAIWhBAJogNNArgB0QQwC4Cm0OAdotIgQDYGHQBmB+6ATgQM7RvvpV7QB5AEYArAgGM8AOiixgUAgA9MAexb8AliUIt6OcUQCSAUUQa8AMSZ5WBAEode-AN6zoK0RLyHEALmgk6AC2QgQsANxuAOZUHjhUPgD8-ux4LFpRkTDQPOIG7Oz+QioqNKRZ0NCgbrDQxiwsatAanCqYeBoqJPHEZM2tJFQAnly0rCQE5ADuABYEJNAABrn57Iv9S3pU7ASLMtlybmGNLInQ-q4H0OIqlP6BIWEVlRQc4untnSQpaRnPAL5Qf7QKBQUAQGBwRYxOIJRDreJTHBDThsGwsCbTcwzYh4PD6GZBeZ4TgANw0OGImEww2gk3M7H2h0gSlU6ma2jCegM0EM8Gpw1MDPgeIJRO0Dh4fDpikIZE4JjMlmstklTmgl0qMKE8R8P3SJEyQNBkBuJFSdKV7DVfH8ivMVmYbBt-AAvBq3B4xJI9dAAIwoaKxHVw-wAIj9YcD2RWHEK0DS6AIxsgYOAORmKl45BmOFJRFIdIaTVC4hw6B2zX4VH0AGtONDg7r4VAzRb6XgRfjxITiS67fyaUMhSSu2K+45pe7Ne5PD6-P7o5VYwV-Ink5B-kA)

I also don't see any evidence that `globalId` is always returned for apply edits w/ attachments. I think that `useGlobalId` has to be true for it to return.

COV-GIS:
`IApplyEditsAttachmentResult` is specifically typed for `attachments` with `applyEdits` which requires global ids. See `Using globalIds and splits in edits` section here: https://developers.arcgis.com/rest/services-reference/enterprise/apply-edits-feature-service-.htm and should be left as is.

https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-feature-service/src/helpers.ts#L59-L78

`IEditFeatureResult` is generic for the individual add/update/delete and apply edits endpoints.

Instead of "typing" a union perhaps two interfaces `IEditFeatureResult` and `IEditFeatureResultError` and use like

```typescript
(IEditFeatureResult | IEditFeatureResultError)[];
```

for all instances it's used in the feature service CRUD methods. Just an idea.

gavinr (approved):

PR #1041, opened 2022-11-29
types(feature-service) add/update attachment interfaces for apply edits

COV-GIS:
Add add/update attachment interfaces for apply edits.

Looking back at `applyEdits` not sure why I didn't include these at the time @patrickarlt. Here they are if you want to add them.


patrickarlt (approved):
LGTM. @gavinr would you also take a look at this and merge it?

gavinr (approved):

Issue #1045, opened 2022-12-16
`solveRoute` throws error when used with `rawResponse`


wsvoja:
### Describe the bug

Hi,

we are using the function `solveRoute` in conjuction with the `rawResponse=true`.
We observed, that this throws an error and looking closer, we found that `solveRoute` calls `cleanResponse` in the `then` section after the request was sent.

https://github.com/Esri/arcgis-rest-js/blob/3.x/packages/arcgis-rest-routing/src/solveRoute.ts#L136

The `json()` method wasn't yet called on this response object, and therefore the property `route` doesn't exist in the response object.

It is accessed however in https://github.com/Esri/arcgis-rest-js/blob/3.x/packages/arcgis-rest-routing/src/solveRoute.ts#L161 which throws the error.

This behavior seems to be applicable for at least version 3.X.X and 4.X.X.

### Reproduction

1. create an [ISolveRouteOptions](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-routing/ISolveRouteOptions) object
2. Add `rawResponse: true` during the creation
3. Call [solveRoute](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-routing/solveRoute)
4. Add a `.catch()` block after the call
5. `solveRoute` will not resolve, it will enter the catch block

### System Info

```shell
@esri/arcgis-rest-routing: "^3.5.1"
```

wsvoja:
Hi, we saw, that not only `solveRoute` has this pattern of applying `cleanResponse` after the response was retrieved.
This at least applies to matrix routing [originDestinationMatrix](https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-routing/src/originDestinationMatrix.ts#L169) and [bulkGeocode](https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-geocoding/src/bulk.ts#L104)
In case of `bulkGeocode` the post processing is not a separate function. It is located in the `then` block directly.

This structure causes someone to receive a different json response, when using `rawResponse` and call `json` on it, than without the usage of `rawResponse`.

Suggestion: Monkey patch `json` when rawResponse is requested:
```
  // ...
  .then((response) => {
    if (options.rawResponse) {
      const origJsonMethod = response.json
      response.json = () => {
        return origJsonMethod.call(response).then(cleanResponse);
      }
      return response;
    }
    // ...
```

[...]

PR #1046, opened 2022-12-19
fix(arcgis-rest-routing): solveRoute rawResponse

gavinr:
fixed issue with solveRoute failing when using `rawResponse: true`

#1045

### DIFF
--- a/packages/arcgis-rest-routing/src/solveRoute.ts
+++ b/packages/arcgis-rest-routing/src/solveRoute.ts
@@ -133,7 +133,12 @@ export function solveRoute(
   }
 
   return request(`${cleanUrl(options.endpoint)}/solve`, options).then(
-    cleanResponse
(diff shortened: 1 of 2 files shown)


patrickarlt (approved):
@gavinr this looks good but I think we also need to figure out if we want to do this in general for #1045. Do you want to merge this or continue the discussion in #1045?

gavinr:
@patrickarlt I checked those other two cases, and they are actually already correctly working: https://github.com/Esri/arcgis-rest-js/issues/1045#issuecomment-1424823159.

I did [add a unit test](https://github.com/Esri/arcgis-rest-js/pull/1046/commits/f65a64d35221cd2da42805fb039b9625212deee2) to cover rawResponse with originDestinationMatrix. 

Unless I hear any objections I will go ahead and merge this.

PR #1048, opened 2022-12-19
chore: default to Node v18

gavinr:
Node v18 is now LTS, so this updates the automated docs and release Actions to use that version.

Also updated to the latest version GitHub actions.


gavinr:
This is blocked until we can resolve #1052.

Issue #1058, opened 2022-12-29
Cannot set referer when getting token from `.signIn()`


gavinr:
### Describe the bug

When a making a request to `.getToken()` with an `ArcGISIdentityManager` instance that was created with `.signIn()`, the referer param is not able to be set.

### Reproduction

Here is the basic reproduction where you can see the referer set to the `window.location.orgin`: https://codepen.io/gavinr/pen/YzjqZmX?editors=0010

![image](https://user-images.githubusercontent.com/209355/209999944-304d5567-f746-4992-8a31-56aee6981209.png)

I'm not sure how the referer _should_ be passed in - via the `signIn()` function or the `getToken()` function or something else?
1. via the `signIn()` function: https://codepen.io/gavinr/pen/BaPKRBW?editors=0010
    ![image](https://user-images.githubusercontent.com/209355/210000749-3e05bcf2-5c75-4283-af53-5fa8956fcc94.png)

2. via the `getToken()` function: https://codepen.io/gavinr/pen/LYBNyPw?editors=0010
    ![image](https://user-images.githubusercontent.com/209355/210001341-6d16931d-5727-4824-a135-022eee886c5e.png)

Note that this same code can be run in NodeJS (which is actually the more standard use case for this behavior, but I have included replication cases in the web because it's easier to share), and the referer will [always be set to `NODEJS_DEFAULT_REFERER_HEADER`](https://github.com/Esri/arcgis-rest-js/blob/367a49c56daa07e09665e43454a80ae6600534f5/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L1579-L1586), which is [hardcoded to  `@esri/arcgis-rest-js`](https://github.com/Esri/arcgis-rest-js/blob/367a49c56daa07e09665e43454a80ae6600534f5/packages/arcgis-rest-request/src/request.ts#L15). (See more info in "Additional Information" below)

### System Info

```shell
"@esri/arcgis-rest-request": "^4.1.1"
```

### Additional Information

The request goes out [here](https://github.com/Esri/arcgis-rest-js/blob/367a49c56daa07e09665e43454a80ae6600534f5/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L1599-L1602), the `referer` param cannot be set - it just uses the `window.location.origin`  if in the browser or the value of `NODEJS_DEFAULT_REFERER_HEADER` (which is hardcoded to ) if in Node:

[...]

PR #1059, opened 2022-12-29
Set Referer

gavinr:
One possible solution (with unit test) to #1058


patrickarlt:
@gavinr I think this approach is fine. You should add `referer` to `ISignInOptions` too though https://github.com/Esri/arcgis-rest-js/blob/367a49c56daa07e09665e43454a80ae6600534f5/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L63

gavinr:
Thanks! I added `referer` to `ISignInOptions`.

patrickarlt (approved):
LGTM!

PR #1061, opened 2023-01-04
84/advanced mapping configuration

juliannaeapicella:

### DIFF
--- a/packages/arcgis-rest-portal/src/users/get-user-properties.ts
+++ b/packages/arcgis-rest-portal/src/users/get-user-properties.ts
@@ -1,0 +1,27 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+import { getPortalUrl } from "../util/get-portal-url";
+import { IGetUserOptions } from "./get-user";
+
+export interface IGetUserPropertiesResponse {
+  /**
+   * user MapViewer configuration
+   */
+  mapViewer: string;
+}
+
+/**
+ * Helper that returns the properties attribute for a user.
+ *
+ * @param IGetUserOptions - options to pass through in the request
+ * @returns User properties object
+ */
+export function getUserProperties(requestOptions: IGetUserOptions): Promise<IGetUserPropertiesResponse> {
+  const username =
+    requestOptions.username || requestOptions.authentication.username;
+  const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/properties`;
+
+  return request(url, requestOptions);
+}
--- a/packages/arcgis-rest-portal/src/users/get-user-properties.ts
+++ b/packages/arcgis-rest-portal/src/users/get-user-properties.ts
@@ -1,7 +1,7 @@
 /* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { IRequestOptions, request } from "@esri/arcgis-rest-request";
+import { IRequestOptions, IUserRequestOptions, request } from "@esri/arcgis-rest-request";
 import { getPortalUrl } from "../util/get-portal-url";
 
 export interface IUserProperties {
@@ -24,7 +24,7 @@ export interface IUserProperties {
  * @param IGetUserPropertiesOptions - options to pass through in the request
  * @returns User properties object
  */
-export async function getUserProperties(username: string, requestOptions?: IRequestOptions): Promise<IUserProperties> {
+export async function getUserProperties(username: string, requestOptions: IRequestOptions): Promise<IUserProperties> {
   const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/properties`;
   const response = await request(url, { httpMethod: 'GET', ...requestOptions });
   if (!response.mapViewer) {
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,27 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+import { getPortalUrl } from "../util/get-portal-url";
+import { IGetUserOptions } from "./get-user";
+
+export interface IGetUserPropertiesResponse {
```
dbouwman, line 8: Is this the only property returned in the response?

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * user MapViewer configuration
+   */
+  mapViewer: string;
```
rweber-esri, line 12: Add missing property we discussed to this interface

packages/arcgis-rest-portal/src/users/get-user-properties.ts
dbouwman: fwiw the API returns `{ properties: {...} }` and there can be other configuration stuff in there beyond `mapViewer`. And for that prop, it seems like `.mapViewer` is only defined if it's set to "classic"... 
To me, that feels like it's "leaking" implementation details up to consumers... so it would be better (imho) to always return a value for `mapViewer` and have it be `"classic" | "latest"`

dbouwman: oh interesting... once you set `mapViewer` to "classic", and then re-set it back, it returns `modern` vs being undefined. So let's use that vs `"latest"`

rweber-esri (approved):
Looks good to me. A couple pieces of minor feedback. Approved once addressed. Don't forget to create a new feature branch off the `main` branch and cherry-pick the commits from this branch onto that branch and open another PR against `main`.

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,44 @@
...
+  mapViewer: "classic" | "modern"
+}
+
+/**
+ * Helper that returns the properties attribute for a user.
+ *
+ * @param IGetUserPropertiesOptions - options to pass through in the request
+ * @returns User properties object
+ */
+export function getUserProperties(requestOptions: IGetUserPropertiesOptions): Promise<IGetUserPropertiesResponse> {
```
dbouwman, line 38: Can we use two params vs creating another interface? `username: string, requestOptions: IRequestOptions`? This lib has a ton of single-use interfaces already...

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,44 @@
...
+  const username =
+    requestOptions.username || requestOptions.authentication.username;
+  const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/properties`;
+
+  return request(url, requestOptions);
```
dbouwman, line 43: Suggest processing the response so it will always returns `.mapViewer` with either `"classic" | "modern"` even if the API does not return anything for `.mapViewer`

packages/arcgis-rest-portal/test/mocks/users/user-properties.ts
```diff
@@ -0,0 +1,6 @@
+import { IGetUserPropertiesResponse } from "../../../src/users/get-user-properties";
+
+export const UserPropertiesResponse: IGetUserPropertiesResponse = {
+  landingPage: { url: 'http://www.esri.com' },
```
dbouwman, line 4: The response has a top level `.properties`, below which the `.mapViewer` and `.landingPage` props are nested. I'm fine if we want to strip that out in the `IGetUserPropertiesResponse` structure, but for fetchMock purposes, it should be structured the way the API returns the information

image

dbouwman:
Before we add this to rest-js we should also built out a `setUserProperties` function, which POSTs to `/sharing/rest/community/users/:username/setProperties` - that way we're providing full functionality for this API end point.

juliannaeapicella:
> Before we add this to rest-js we should also built out a `setUserProperties` function, which POSTs to `/sharing/rest/community/users/:username/setProperties` - that way we're providing full functionality for this API end point.

@dbouwman are there any requirements for this kind of function? e.g. does it require authentication?

dbouwman:
@juliannaeapicella re: `landingPage` if we can determine what the default value is, then yes. 
for `setUserProperties` - yes - that will require auth, so use `IUserRequestOptions` which makes `session` a required prop.

patrickarlt:
@dbouwman @juliannaeapicella is there public facing documentation for the `username/properties` endpoint? This doesn't appear in the public REST API doc https://developers.arcgis.com/rest/users-groups-and-items/remove-user-resource.htm

juliannaeapicella:
> @dbouwman @juliannaeapicella is there public facing documentation for the `username/properties` endpoint? This doesn't appear in the public REST API doc https://developers.arcgis.com/rest/users-groups-and-items/remove-user-resource.htm

@patrickarlt No, there is currently no documentation. We discovered this endpoint by observing the network tab in AGO.

juliannaeapicella:
Closing pull request in favor of updated branch here: #1068

PR #1066, opened 2023-01-09
chore: automated unit test timing issue

gavinr:
The unit tests on ubuntu-14 are fairly consistently running into the "off by a few milliseconds" error that is predicted here:
https://github.com/Esri/arcgis-rest-js/blob/367a49c56daa07e09665e43454a80ae6600534f5/packages/arcgis-rest-request/test/ArcGISIdentityManager.test.ts#L1164

Examples:

- image
- image

In general we can get it to work if we keep re-running the unit test, but that's quite time consuming. This PR loosens the unit test so that it will pass more often so we don't have to re-run it as much.


patrickarlt (approved):
LGTM!

PR #1067, opened 2023-01-11
feat(arcgis-rest-portal): add user properties functions

juliannaeapicella:


rweber-esri (approved):
Thanks for making those changes. This is looking good to me. I copied this code into hub-components and verified that both `getUserProperties` and `setUserProperties` work as I'd expect. May want to wait for an approval from @dbouwman or @patrickarlt since they contributed feedback, too.

rweber-esri:
If @dbouwman and @patrickarlt don't have additional feedback, I'd say this is good to merge.

PR #1068, opened 2023-01-11
feat(arcgis-rest-portal): add user properties functions

juliannaeapicella:

### DIFF
--- a/packages/arcgis-rest-portal/src/users/get-user-properties.ts
+++ b/packages/arcgis-rest-portal/src/users/get-user-properties.ts
@@ -1,7 +1,8 @@
 /* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { IRequestOptions, request } from "@esri/arcgis-rest-request";
+import { request } from "@esri/arcgis-rest-request";
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
 import { getPortalUrl } from "../util/get-portal-url";
 
 export interface IUserProperties {
@@ -21,10 +22,10 @@ export interface IUserProperties {
 /**
  * Fetches the properties for a user
  * @param username The user whose properties to fetch
- * @param requestOptions An IRequestOptions object
+ * @param requestOptions An IUserRequestOptions object
  * @returns a promise that resolves an IUserProperties object
  */
-export async function getUserProperties(username: string, requestOptions: IRequestOptions): Promise<IUserProperties> {
+export async function getUserProperties(username: string, requestOptions: IUserRequestOptions): Promise<IUserProperties> {
   const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/properties`;
   const response = await request(url, { httpMethod: 'GET', ...requestOptions });
   if (!response.properties.mapViewer) {
--- a/packages/arcgis-rest-portal/src/users/get-user-properties.ts
+++ b/packages/arcgis-rest-portal/src/users/get-user-properties.ts
@@ -2,7 +2,6 @@
  * Apache-2.0 */
 
 import { IRequestOptions, request } from "@esri/arcgis-rest-request";
-import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
 import { getPortalUrl } from "../util/get-portal-url";
 
 export interface IUserProperties {
@@ -32,29 +31,4 @@ export async function getUserProperties(username: string, requestOptions: IReque
     response.properties.mapViewer = "modern";
   }
   return response.properties;
-}
-
-/**
- * Updates the properties for a user
- * @param username The user whose properties to update
- * @param properties IUserProperties object with properties to update
- * @param requestOptions An IRequestOptions object
- * @returns a promise that resolves to { success: boolean }
- */
-export async function setUserProperties(username: string, properties: IUserProperties, requestOptions: IRequestOptions): Promise<{ success: boolean }> {
-  const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/setProperties`;
-  const options: IRequestOptions = {
-    httpMethod: 'POST',
-    params: { properties },
-    ...requestOptions
-  };
-  try {
-    const response = await request(url, options)
-    if (!response.success) {
-      throw new Error("Success was false");
-    }
-    return response;
-  } catch(e) {
-    throw new Error(`Failed to set user properties: ${e.message}`);
-  }
 }
--- a/packages/arcgis-rest-portal/src/users/get-user-properties.ts
+++ b/packages/arcgis-rest-portal/src/users/get-user-properties.ts
@@ -1,0 +1,54 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions, request } from "@esri/arcgis-rest-request";
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
+import { getPortalUrl } from "../util/get-portal-url";
+
+export interface IUserProperties {
+  /**
+   * user landing page configuration
+   */
+  landingPage: {
+    url: string
+  },
+  /**
+   * user MapViewer configuration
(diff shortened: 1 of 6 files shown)


packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function getUserProperties(username: string, requestOptions: IRequestOptions): Promise<IUserProperties> {
+  const url = `${getPortalUrl(requestOptions)}/community/users/${encodeURIComponent(username)}/properties`;
+  const response = await request(url, { httpMethod: 'GET', ...requestOptions });
+  if (!response.properties.mapViewer) {
+    response.properties.mapViewer = "modern";
+  }
+  return response.properties;
+}
+
+export async function setUserProperties(properties: IUserProperties, requestOptions: IUserRequestOptions): Promise<void> {
```
rweber-esri, line 37: Maybe add a comment/docs for this function. Additionally, it seems a little strange to me that `setUserProperties` would reside in a file named `get-user-properties`. Maybe split this fn out into `set-user-properties.ts`? Also, we specifying that this function returns `Promise<void>` but it actually returns `Promise<{ success: boolean }>`.

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+  /**
+   * user MapViewer configuration
+   */
+  mapViewer: "classic" | "modern",
+  [ key: string ]: unknown
+}
+
+/**
+ * Helper that returns the properties attribute for a user.
```
rweber-esri, line 23: I'd consider updating this documentation to be something like:

```ts
/**
 * Fetches the properties for a user
 * @param username The user whose properties to fetch
 * @param requestOptions A IUserRequestOptions object
 * @returns a promise that resolves an IUserProperties object
 */
```

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param IGetUserPropertiesOptions - options to pass through in the request
+ * @returns User properties object
+ */
+export async function getUserProperties(username: string, requestOptions: IRequestOptions): Promise<IUserProperties> {
```
rweber-esri, line 28: Should `requestOptions: IRequestOptions` be `requestOptions: IUserRequestOptions` here?

juliannaeapicella: @rweber-esri IUserRequestOptions is used because setUserProperties only sets the authenticated user's properties, but currently getUserProperties is configured so you can get any user's properties. I'm not sure if it should be more specific like the post method.

rweber-esri: > ... setUserProperties only sets the authenticated user's properties

See my updated comment above, are you sure that org admins can't update the properties of other users that belong to the org admin's organization?

rweber-esri: I verified that org admins can successfully update the properties of other users in their org. We should refactor this per my ^ suggestions.

packages/arcgis-rest-portal/src/users/set-user-properties.ts
```diff
@@ -0,0 +1,32 @@
...
+    ...requestOptions
+  };
+  try {
+    const response = await request(url, options)
+    if (!response.success) {
+      throw new Error("Success was false");
+    }
+    return response;
+  } catch(e) {
+    const error = e as Error;
```
rweber-esri, line 29: May not need to cast if we add the type to the arg in the catch statement.

```ts
} catch (e: Error) {
  throw new Error(`Failed to set user properties: ${error.message}`);
}
```

juliannaeapicella: @rweber-esri For whatever reason, TypeScript throws an error when trying to explicitly type the variable for a catch clause: https://typescript.tv/best-practices/error-ts1196-catch-clause-variable-type-annotation/. Although it seems a type guard might be more standard practice than casting it?

rweber-esri: I observe that too when I pull this branch down. This is fine to leave as-is

packages/arcgis-rest-portal/src/users/set-user-properties.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { IRequestOptions, request } from "@esri/arcgis-rest-request";
+import { getPortalUrl } from "../util/get-portal-url";
+import { IUserProperties } from "./get-user-properties";
+
+/**
+ * Updates the properties for a user
+ * @param username The user whose properties to update
+ * @param properties IUserProperties object with properties to update
+ * @param requestOptions An IRequestOptions object
```
rweber-esri, line 12: IIRC, @dbouwman suggested this be `IUserRequestOptions` instead of `IRequestOptions`.

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,34 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions, request } from "@esri/arcgis-rest-request";
```
rweber-esri, line 4: IUserRequestOptions in this file too

rweber-esri (approved):
Thanks for making those changes. This is looking good to me. I copied this code into hub-components and verified that both `getUserProperties` and `setUserProperties` work as I'd expect. May want to wait for an approval from @dbouwman or @patrickarlt since they contributed feedback, too.

packages/arcgis-rest-portal/src/users/get-user-properties.ts
```diff
@@ -0,0 +1,35 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { request } from "@esri/arcgis-rest-request";
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
+import { getPortalUrl } from "../util/get-portal-url";
+
+export interface IUserProperties {
```
rweber-esri, line 8: I think we also need to add an re-export statement in [index.ts](https://github.com/Esri/arcgis-rest-js/blob/84/advanced-mapping-configuration-v3/packages/arcgis-rest-portal/src/index.ts#L39) for both of these files so these can be imported properly.

packages/arcgis-rest-portal/test/users/set-user-properties.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        mapViewer: "modern"
+      }
+
+      setUserProperties(session.username, properties, {
+        authentication: session
+      })
+        .then(() => {
+          expect(fetchMock.called()).toEqual(true);
+          const [url, options]: [string, RequestInit] = fetchMock.lastCall();
+          console.log("options:", options);
```
rweber-esri, line 46: Remove debug statement

rweber-esri:
If @dbouwman and @patrickarlt don't have additional feedback, I'd say this is good to merge.

dbouwman (approved):
This looks good to me - I assume we'll PR the same code to `main` so it gets into the 4.x releases as well. Once this is merged, I can run a 3.x release (it's finicky). Ping me

rweber-esri:
> This looks good to me - I assume we'll PR the same code to `main` so it gets into the 4.x releases as well. Once this is merged, I can run a 3.x release (it's finicky). Ping me

Thanks @dbouwman 

@juliannaeapicella has ported this work to a [feature branch](https://github.com/Esri/arcgis-rest-js/tree/84/advanced-mapping-configuration) cut off `main`, and submitted [PR-1067](https://github.com/Esri/arcgis-rest-js/pull/1067). I believe that's ready to merge too. @juliannaeapicella coordinate with @dbouwman once both of these merge so we can make sure this work is released under 3.x and 4.x.

Issue #1070, opened 2023-01-19
OAuth2 flow fails when using hash routing in single-page applications


philnagel:
### Describe the bug

When using has routing in single-page applications (in my case, `vue-router` in `VueJS` with the [hash history mode](https://router.vuejs.org/guide/essentials/history-mode.html#hash-mode) enabled), the PKCE-based OAuth2 flow fails. I think this is caused by the fact that in this case, the authentication page that runs `completeOAuth2()` will have a URL with both a hash and a search component. This requires special handling to correctly parse this type of URL.

Specifically, this error is returned even when everything else should enable a succesful authentication:

```
Uncaught (in promise) ArcGISAuthError: no-auth-state: No authentication state was found, call `ArcGISIdentityManager.beginOAuth2(...)` to start the authentication process.
```

### Reproduction

I've created a repository with the necessary code to reproduce this issue in `VueJS` [here](https://github.com/philnagel/oauth-bug).
The readme of that repo should outline the steps needed to reproduce - let me know of any questions or need for clarification.

The demo shows that the auth flow works fine when using HTML5 history mode, which does not leverage `#` in the route URL. By only changing the router history mode, the flow fails when all else remains equal.

### Logs

```shell
Uncaught (in promise) ArcGISAuthError: no-auth-state: No authentication state was found, call `ArcGISIdentityManager.beginOAuth2(...)` to start the authentication process.
```

### System Info

```shell
@esri/arcgis-rest-request: 4.1.1

  System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 19.38 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.volta/tools/image/node/16.16.0/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 8.11.0 - ~/.volta/tools/image/node/16.16.0/bin/npm
  Browsers:
    Brave Browser: 105.1.43.93
    Chrome: 109.0.5414.87
    Firefox: 108.0.2
    Safari: 16.2
```

patrickarlt:
Closing see https://github.com/Esri/arcgis-rest-js/pull/1071#issuecomment-1425085829

PR #1071, opened 2023-01-19
fix(arcgis-rest-request): fix issue with OAuth2 flow when using hash routing

philnagel:
fixes Esri/arcgis-rest-js#1070

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -542,11 +542,20 @@ export class ArcGISIdentityManager implements IAuthenticationManager {
     const stateStorageKey = `ARCGIS_REST_JS_AUTH_STATE_${clientId}`;
     const stateId = win.localStorage.getItem(stateStorageKey);
 
+    // if redirectUri has hash already, we need special treatment
+    const redirectHasHash =
+      new URL(redirectUri).hash && win.location.toString().includes(redirectUri)
+        ? true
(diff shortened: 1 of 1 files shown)


patrickarlt:
@philnagel thanks for the fix and the reproduction however oAuth 2.0 redirect URIs cannot contain fragments (hashes). From the spec:

> The redirection endpoint URI MUST be an absolute URI as defined by
   [[RFC3986] Section 4.3](https://www.rfc-editor.org/rfc/rfc3986#section-4.3).  The endpoint URI MAY include an
   "application/x-www-form-urlencoded" formatted (per [Appendix B](https://www.rfc-editor.org/rfc/rfc6749#appendix-B)) query
   component ([[RFC3986] Section 3.4](https://www.rfc-editor.org/rfc/rfc3986#section-3.4)), which MUST be retained when adding
   additional query parameters.  The endpoint URI MUST NOT include a
   fragment component.
> 
>https://www.rfc-editor.org/rfc/rfc6749#section-3.1.2

The last sentence is the deal breaker here:

> The endpoint URI MUST NOT include a fragment component.

If you want to use PKCE in your app and still use hash-based routing in Vue you should have the page defined in the redirect URI not be handled by Vue at all. Hosting a static [`authenticate.html` like in the oAuth 2.0 demo](https://github.com/Esri/arcgis-rest-js/blob/main/demos/oauth2-browser/authenticate.html) is enough to suffice but you can't have Vue handle 100% of this and use hash-based routing.

PR #1086, opened 2023-03-05
chore: update advanced oAuth demo for dev summit

patrickarlt:
This updates the advanced oAuth demo for dev summit.


noahmulfinger (approved):
@patrickarlt LGTM I had two small updates. I also think there should be more explanation about what the buttons in app.html are supposed to do. But not a blocking issue.

PR #1102, opened 2023-04-25
fix(arcgis-rest-request): geometry type esriGeometryMultiPatch

### LINKED ISSUE #832: ILayerDefinition "geometryType" field can be "esriGeometryMultiPatch"
I don't expect the library to support consuming the binary MultiPatch encoding natively, but it's possible for [a layer definition](https://services.arcgis.com/V6ZHFr6zdgNZuVG0/ArcGIS/rest/services/LABuildings_3D/FeatureServer/0) to come back with the subject value for the `geometryType` field.  The typings for the `ILayerDefinition` interface should reflect this.

(For such layers, it's possible to specify `multipatchOption: "xyFootprint"` and get back `esriGeometryPolygon` features instead, which are easier to work with.)

gavinr:
fixes #832

### DIFF
--- a/packages/arcgis-rest-request/src/types/geometry.ts
+++ b/packages/arcgis-rest-request/src/types/geometry.ts
@@ -175,7 +175,8 @@ export type GeometryType =
   | "esriGeometryMultipoint"
   | "esriGeometryPolyline"
   | "esriGeometryPolygon"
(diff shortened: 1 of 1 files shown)


patrickarlt (approved):
LGTM!

PR #1104, opened 2023-05-26
docs: update contributor docs

patrickarlt:

PR #1107, opened 2023-06-07
fix: api reference links

COV-GIS:
And update the repo about website url to https://developers.arcgis.com/arcgis-rest-js/ as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,9 @@ request(url).then((response) => {
 
 ### API Reference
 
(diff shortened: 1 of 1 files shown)


gavinr:
Thanks for the contribution! This change is already in progress under https://github.com/Esri/arcgis-rest-js/pull/1104, so I'll close this one and we'll be sure it's updated there. Thanks!

Issue #1119, opened 2023-08-31
Hub needs ability to search for users outside the currently authenticated user's organization


rweber-esri:
### Describe the problem

The Hub Discussions feature has the concept of "channels", which is an access control mechanism governing what user's can participate in discussions. Channels currently have an `access` property that can be one of `public|org|private`, and a `groups` property which is an array of ArcGIS Online group IDs that can additionally grant participation privileges to users (groups can be outside the currently auth'd user's organization). User's permitted to participate in a discussion can at-mention other users that are also permitted to participate in a discussion (given the target user's profile visibility is set such that other users, internal, or external to their org can view their record). Given the above, Hub needs the ability to search for users that may be outside the currently authenticated user's organization.

Today, both V3 & V4 of `arcgis-rest-js` only support searching for users from the `/portals/self/users/search` AGO API endpoint, but it seems that endpoint only searches within the currently authenticated user's organization. 

V3:
* [searchUsers](https://github.com/Esri/arcgis-rest-js/blob/97322cabd6ad07bf0b5303b9719d2c085bde3c54/packages/arcgis-rest-portal/src/users/search-users.ts#L36)
* [genericSearch](https://github.com/Esri/arcgis-rest-js/blob/97322cabd6ad07bf0b5303b9719d2c085bde3c54/packages/arcgis-rest-portal/src/util/generic-search.ts#L84)

V4:
* [searchUsers](https://github.com/Esri/arcgis-rest-js/blob/e18091a63755d25acf4cab1c819886b9b06c5d78/packages/arcgis-rest-portal/src/users/search-users.ts#L35)
* [genericSearch](https://github.com/Esri/arcgis-rest-js/blob/e18091a63755d25acf4cab1c819886b9b06c5d78/packages/arcgis-rest-portal/src/util/generic-search.ts#L86)

The [/community/users](https://developers.arcgis.com/rest/users-groups-and-items/user-search.htm) endpoint seems to be able to access user records _outside_ the currently authenticated user's organization.

### Describe the proposed solution

I believe we should either:

1. Add an entirely new search method, e.g. `searchCommunityUsers`, that would pass a new `searchType` of `communityUser` (or similar) to `genericSearch` to route the request to the `/community/users` endpoint.

```ts
// existing method
export function searchUsers(
  search: IUserSearchOptions | SearchQueryBuilder
): Promise<ISearchResult<IUser>> {
  return genericSearch<IUser>(search, "user");
}

// new method added
[...]

PR #1120, opened 2023-09-01
feat(arcgis-rest-portal): add searchCommunityUsers method to allow fo…

rweber-esri:
[1119](https://github.com/Esri/arcgis-rest-js/issues/1119)

- Refactors `genericSearch` to support calling [/community/users](https://developers.arcgis.com/rest/users-groups-and-items/user-search.htm) endpoint when `searchType` is `communityUser`
- Adds `searchCommunityUsers` method that calls through to `genericSearch`, passing a `searchType` of `communityUser`

V3

### DIFF
--- a/packages/arcgis-rest-portal/src/users/search-users.ts
+++ b/packages/arcgis-rest-portal/src/users/search-users.ts
@@ -43,7 +43,7 @@ export function searchUsers(
  * searchCommunityUsers({ q: 'tommy', authentication })
  *   .then(response) // response.total => 355
  * ```
- * Search a all portals for users.
+ * Search all portals for users.
  *
  * @param search - A RequestOptions object to pass through to the endpoint.
  * @returns A Promise that will resolve with the data from the response.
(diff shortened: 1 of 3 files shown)


tomwayson (commented):
Thanks for this @rweber-esri 

My only concern is what base branch this should be against. I am guessing that since you want this in hub, and hub is still on 3.x, that you want it to go into a 3.x release. I know `main` is for 4.x. I forget what the intent of the `master` branch is, but I recall at some point we were using it for 3.x, maybe while 4.x was under development? I see that there are commits to the `3.x` branch that are more recent than the latest commit on master.

rweber-esri:
> My only concern is what base branch this should be against. I am guessing that since you want this in hub, and hub is still on 3.x, that you want it to go into a 3.x release. I know `main` is for 4.x. I forget what the intent of the `master` branch is, but I recall at some point we were using it for 3.x, maybe while 4.x was under development? I see that there are commits to the `3.x` branch that are more recent than the latest commit on master.

Thanks for the input @tomwayson, you are correct, I want the changes on this branch published under `3.x`. @dbouwman @gavinr @patrickarlt can you verify what the correct branch is for publishing changes under v3 (`master` vs `3.x`)?

dbouwman (approved):

tomwayson:
This is now released as 3.7.0

Issue #1117, opened 2023-08-16
Update REST JS places documentation to reflect pagination changes


sherylgiovanni:
### Description

There has been a couple of changes to the places-service relating to pagination which need updating in the REST JS documentation.

Theses pagination changes are:
- Removed property `maxResultsExceeded`
- Renamed `links` object to `pagination` 
- Renamed `next` and `previous` to `nextUrl` and `previousUrl`
- `nextUrl` and `previousUrl` are now full Urls rather than Url fragments - `links.base` removed

Old format
```
{
  "results": [/*snip*/],
  "maxResultsExceeded": true,
  "links": {
    "base": "https://places-api.arcgis.com",
    "previous": "/arcgis/rest/services/places-service/v1/places/near-point?x=-3.18718337&y=55.95203695&radius=1000.00000000&searchText=&f=pjson&offset=0&pageSize=10",
    "next": "/arcgis/rest/services/places-service/v1/places/near-point?x=-3.18718337&y=55.95203695&radius=1000.00000000&searchText=&f=pjson&offset=20&pageSize=10"
  }
}
```

New format:

```
{
  "results": [/*snip*/],
  "pagination": {
    "previousUrl": "https://placesdev-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point?x=-3.18718337&y=55.95203695&radius=1000.00000000&searchText=&f=pjson&offset=0&pageSize=10",
    "nextUrl": "https://placesdev-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point?x=-3.18718337&y=55.95203695&radius=1000.00000000&searchText=&f=pjson&offset=20&pageSize=10"
  }
}
```

### Changes needed
These are the pages that still have the old attributes `links` and `maxResultsExceeded` and they need to be updated accordingly:
- [IFindPlacesNearPointResponse](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-places/IFindPlacesNearPointResponse)
- [IFindPlacesWithinExtentResponse](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-places/IFindPlacesWithinExtentResponse)
- [findPlacesNearPoint](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-places/findPlacesNearPoint/)
[...]

PR #1133, opened 2023-10-09
Update documentation for places pagination

patrickarlt:
RE: #1117

### DIFF
--- a/packages/arcgis-rest-places/src/openapi-types.ts
+++ b/packages/arcgis-rest-places/src/openapi-types.ts
@@ -888,7 +888,7 @@ export interface components {
       /**
        * @description The census block Id of the place (US only).
        *
-       * @example 060710081003002
+       * @example 60710081003002
        */
       censusBlockId?: string | null;
     };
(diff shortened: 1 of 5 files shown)


packages/arcgis-rest-places/src/openapi-types.ts
```diff
@@ -1263,20 +1389,18 @@ export interface operations {
...
+   *
+   * **Note**: Query parameters are case-sensitive.
    */
   nearPointGet: {
     parameters: {
       query: {
-        /** The x coordinate, or longitude, to search from, in WGS84 decimal degrees. */
-        x: components["schemas"]["XCoord"];
-        /** The y coordinate, or latitude, to search from, in WGS84 decimal degrees. */
-        y: components["schemas"]["YCoord"];
```
gavinr-maps, line 1279: I think `x` and `y` are required here - see https://developers.arcgis.com/rest/places/near-point-get/#query-parameters ... unless `x` and `y` are included elsewhere and I'm just not understanding the `openapi-types.ts` correctly

patrickarlt:
@gavinr should be good now.

gavinr-maps (approved):
Overall looks good. I was doing some final searching and did find one additional place: in `demos\node-places-api\nearPoint.mock.json` it's still using the old format - but I'll update all those demo files in a follow-up PR (#1134).

PR #1135, opened 2023-10-13
chore: change postal and postalExt to string

### LINKED ISSUE #1094: Geocoding postal code type
### Description

Jim Barry brought to my attention that the postal code parameter should be wrapped in quotes as it is an alphanumeric value. Issues arise when using a postal code starting with zero as demonstrated in these CodePen examples ([ex1](https://codepen.io/cyatteau/pen/OJoYRNr), [ex2](https://codepen.io/cyatteau/pen/BaOeRLj)). 

Several places state or use a _number_ instead of a _string_, which need to be addressed:
- [ ] The [API Reference](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-geocoding/IGeocodeOptions#postal) states that the postal code value should be a _number_.
- [ ] The example given [here](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-geocoding/geocode).
- [ ] The tutorial [here](https://developers.arcgis.com/arcgis-rest-js/geocode-and-search/search-for-an-address/).

### Link to page

https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-geocoding/IGeocodeOptions#postal

BrunoCaimar:
Fix #1094

### DIFF
--- a/packages/arcgis-rest-geocoding/test/bulk.test.ts
+++ b/packages/arcgis-rest-geocoding/test/bulk.test.ts
@@ -159,6 +159,51 @@ describe("geocode", () => {
       });
   });
 
+  it("should send a bulk geocoding request with postal params correctly", (done) => {
+    fetchMock.once("*", GeocodeAddresses);
+    const addresses = [
+      {
+        OBJECTID: 1,
+        address: "380 New York St. Redlands",
+        postal: 92373
+      },
+      {
+        OBJECTID: 2,
+        address: "1205 Williston Rd",
+        postal: "05403"
+      }
+    ];
+
+    bulkGeocode({
+      addresses,
+      endpoint:
+        "https://customer.gov/arcgis/rest/services/CompositeGeocoder/GeocodeServer/",
+      params: {
+        outSR: 4326,
+        forStorage: true
+      }
+    })
+      // tslint:disable-next-line
+      .then((response) => {
+        expect(fetchMock.called()).toEqual(true);
+        const [url, options] = fetchMock.lastCall("*");
+        expect(url).toEqual(
+          "https://customer.gov/arcgis/rest/services/CompositeGeocoder/GeocodeServer/geocodeAddresses"
+        );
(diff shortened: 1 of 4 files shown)


BrunoCaimar:
Changed to accept string or number. Also, added tests to cover the change.

packages/arcgis-rest-geocoding/test/bulk.test.ts
```diff
@@ -159,6 +159,51 @@ describe("geocode", () => {
...
+        expect(url).toEqual(
+          "https://customer.gov/arcgis/rest/services/CompositeGeocoder/GeocodeServer/geocodeAddresses"
+        );
+        expect(options.method).toBe("POST");
+        expect(options.body).toContain("f=json");
+        expect(options.body).toContain("outSR=4326");
+        expect(options.body).toContain("forStorage=true");
+        expect(options.body).toContain(
+          `addresses=${encodeURIComponent(
+            '{"records":[{"attributes":{"OBJECTID":1,"address":"380 New York St. Redlands","postal":92373}},{"attributes":{"OBJECTID":2,"address":"1205 Williston Rd","postal":"05403"}}]}'
```
gavinr-maps, line 199: @patrickarlt for cases of batch geocoding, is it ok that we could be sending the `postal` as both a string and number? Or should we be making that consistent within ArcGIS REST JS before it makes the rest call?

gavinr-maps (commented):
Overall looks good. Thanks @BrunoCaimar!

I had one question that I wanted to ask @patrickarlt what he thinks:

packages/arcgis-rest-geocoding/test/bulk.test.ts
patrickarlt: I don't think so, the server should be able to sort it out. Lets merge this.

gavinr-maps (approved):

PR #1136, opened 2023-10-16
feat: user app resources; tests use spyOn

dbouwman:
This adds functions for working with user app resources. These functions are copied from hub.js, along with the tests.

### NOTE
The tests included use `spyOn` to create jasmine spies for `request`. This does not work in this project!


dbouwman:
Main question here is: Can we use spies in here or must we mock all the calls? Hub is already using these fns, but to land in here we'd need to spend time changing the tests... which is hard to fit into my task list

dbouwman:
Fundamentally, we can't use spies in rest-js b/c it's using typescript 3.9+, which marks module properties as read only, precluding the use of jasmine spies. In order to merge this, all the tests needs to be re-worked to use fetchMock. At this time none of us have bandwidth to tackle this. Closing.

PR #1153, opened 2024-04-18
Remove GH pages build process

patrickarlt:
@dbouwman @tomwayson @gavinr As discussed in our REST JS meeting this removes the GH pages build process from `master` so we quit overwriting the 4.x typedoc output with 3.x builds.


gavinr-maps (approved):
Looks good to me.

PR #1170, opened 2024-06-12
Allow arcgis rest places package to use dev endpoints

MonikaKanduri4525:
- Added an endpoint param for places method in arcgis-rest-places package which allows to override the built in URL.
- Added test cases 

Pending:
comments for TypeDoc

### DIFF
--- a/packages/arcgis-rest-places/test/findPlacesNearPoint.test.ts
+++ b/packages/arcgis-rest-places/test/findPlacesNearPoint.test.ts
@@ -78,18 +78,18 @@ describe("findPlacesNearPoint()", () => {
   it("verify endpoint", async () => {
     fetchMock.mock("*", placeNearPointMockNoMoreResults);
 
-    const response = await findPlacesNearPoint({
+    await findPlacesNearPoint({
       x: -3.1883,
       y: 55.9533,
       radius: 10,
       endpoint:
-        "https://placesdev-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point",
+        "https://places-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point",
       authentication: ApiKeyManager.fromKey("MOCK_KEY")
     });
 
     const [url, options] = fetchMock.lastCall("*") as MockCall;
     expect(url).toEqual(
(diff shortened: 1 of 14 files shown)


patrickarlt (changes requested):
@MonikaKanduri4525 this looks great could you reset the demo URL and work on the typedoc comments based on my example?

I'm also going to ask @gavinr to review this as another contributor.

packages/arcgis-rest-places/test/findPlacesNearPoint.test.ts
```diff
@@ -74,4 +74,22 @@ describe("findPlacesNearPoint()", () => {
...
+
+  it("verify endpoint", async () => {
+    fetchMock.mock("*", placeNearPointMockNoMoreResults);
+
+    const response = await findPlacesNearPoint({
+      x: -3.1883,
+      y: 55.9533,
+      radius: 10,
+      endpoint:
+        "https://placesdev-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point",
```
gavinr-maps, line 86: Similar to Patrick's comment above, I think all these example URLs in the test spec files should be changed to 
```suggestion
        "https://places-api.arcgis.com/arcgis/rest/services/places-service/v1/places/near-point",
```

packages/arcgis-rest-places/test/findPlacesNearPoint.test.ts
```diff
@@ -74,4 +74,22 @@ describe("findPlacesNearPoint()", () => {
     expect(nextPage.nextPage).toBeUndefined();
     expect(url).toContain("token=MOCK_KEY");
   });
+
+  it("verify endpoint", async () => {
+    fetchMock.mock("*", placeNearPointMockNoMoreResults);
+
+    const response = await findPlacesNearPoint({
```
gavinr-maps, line 81: In these tests files, since we're not using the `response` variable, I think you can leave that off, like this:
```suggestion
    await findPlacesNearPoint({
```

gavinr-maps (changes requested):
Overall looks good! In addition to Patrick's comments, just a few small things:

MonikaKanduri4525:
@patrickarlt @gavinr-maps - Thanks for the inputs, addressed the feedback here [161a784](https://github.com/Esri/arcgis-rest-js/pull/1170/commits/161a7840e01ba70436ba4245c5c11c53d6a3d362)

patrickarlt (approved):
LGTM I'll merge once tests pass.

PR #1176, opened 2024-08-22
Fix: Replace Skypack with esm.run

sherylgiovanni:
With Skypack being no longer supported, we decided to migrate to esm.run which hopefully will have more long-term support. We updated all documentation referencing Skypack to reference esm.run now.


patrickarlt (approved):
LGTM!

PR #1188, opened 2025-01-09
chore: npm serve command

gavinr-maps:
Just for convenience, adds an npm command (`npm run serve`) to run the script that is already here in the repo.


patrickarlt (approved):
LGTM

PR #1190, opened 2025-01-14
Add support for the new Elevation service via a new `arcgis-rest-elevation` package

sherylgiovanni:
cc @patrickarlt @gavinr-maps


gavinr-maps (changes requested):
I think we probably want to include a `README.md` within the folder `packages\arcgis-rest-elevation`, similar to the other packages.

packages/arcgis-rest-elevation/src/index.ts
```diff
@@ -0,0 +1,12 @@
+/* Copyright (c) 2023 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { findElevationAtPoint } from "./findElevationAtPoint.js";
+import { findElevationAtManyPoints } from "./findElevationAtManyPoints.js";
```
gavinr-maps, line 5: Are these lines required?

patrickarlt (changes requested):
@sherylgiovanni Overall this looks good but I do have a few small comments.

packages/arcgis-rest-elevation/src/findElevationAtManyPoints.ts
```diff
@@ -0,0 +1,83 @@
...
+ * });
+ *
+ * console.log(results)
+ * ```
+ */
+export function findElevationAtManyPoints(
+  requestOptions: IFindElevationAtManyPointsOptions
+): Promise<IFindElevationAtManyPointsResponse> {
+  return (
+    request(`${baseUrl}/elevation/at-many-points`, {
```
gavinr-maps, line 69: I think you forgot to use `appendCustomParams` here. I think you should create an `options` variable using `appendCustomParams` similar to how you're doing it in line 60 of `findElevationAtPoint.ts`.

patrickarlt (approved):
LGTM

PR #1193, opened 2025-02-12
Developer Credentials v2

patrickarlt:
This PR updates the `@esri/arcgis-rest-developer-credentials` package to work with the new version of API keys released last year. This is a breaking change that will be released as a 2.0.0. Woth updates to the `createApiKey()` and `updateApiKey()`.

* Removes the `deleteApiKey()` and `deleteOAuthApp()` which can be replaced by `removeItem()` from the portal package.
* Removes most checks for a specific type of item since the new API key actually looks very similar to an OAuth 2.0 app. It will be up to users to manage this.
* Removes the `Privileges` enum. Since the new API keys can have privilege that a user has we can't encode every privilege.

This also has a matching PR in the samples repo https://github.com/Esri/arcgis-rest-js-samples/pull/8


gavinr-maps (approved):
Looks good to me.

PR #1194, opened 2025-02-20
chore: add manual release option

patrickarlt:
Fixes our manual release job to actually work.


gavinr-maps (approved):
Moving the changes from https://github.com/Esri/arcgis-rest-js/blob/main/.github/workflows/release.yml over into this manual-release file. Looks good.

PR #1196, opened 2025-02-26
Add invalidateApiKey(), generate API key bug fix, doc updates

patrickarlt:
Adds a few things I would like in for Dev and Tech Summit:

* `invalidateApiKey()` this isn't publically documented and has some quirks so it is good to wrap.
* Fix an issue with generating API key tokens. We only need to pass `client_secret` here.
* Clean up some code examples in the docs.


gavinr-maps (approved):
looks good

Issue #1201, opened 2025-03-17
Add "Developer Quick Start" instructions to CONTRIBUTING.md


gavinr-maps:
### Describe the problem

Since we have [moved the demos/samples to a separate repo](https://github.com/Esri/arcgis-rest-js/pull/1192), there is now no quick way to run a quick code test manually. For example, if I want to verify a bug.

### Describe the proposed solution

I propose adding a new section to [CONTRIBUTING.md](https://github.com/Esri/arcgis-rest-js/blob/main/CONTRIBUTING.md) called "Developer Quick Start". This is not a quick start for people who are using ArcGIS REST JS - that is located here: https://developers.arcgis.com/arcgis-rest-js/get-started/ ... this is a "Quick Start" for developers making changes or testing ArcGIS REST JS. This section would contain, roughly:

1. In a terminal, run `npm run dev:bundled`
2. In a separate terminal, run `npm run serve`
3. Open https://localhost:8080/sample.html

... and we would add `sample.html` to this repo containing the following:

<details>

<summary>sample.html</summary>

```html
  <!DOCTYPE html>
  <html lang="en">
    <head>
      <meta charset="utf-8" />
      <title>Sample</title>
    </head>

    <body>
      <!-- Map package imports to URLS starting with @esri/... which will get served from the local build of the modules -->
      <script type="importmap">
        {
          "imports": {
            "@esri/arcgis-rest-request": "/@esri/arcgis-rest-request/dist/bundled/request.esm.js",
            "@esri/arcgis-rest-portal": "/@esri/arcgis-rest-portal/dist/bundled/portal.esm.js"
          }
        }
      </script>

      <script type="module">
        import { searchItems } from "@esri/arcgis-rest-portal";

        let element = document.createElement("pre");
        document.body.appendChild(element);

        searchItems("water").then((response) => {
          element.textContent = JSON.stringify(response, null, 2); // false
        });
      </script>
    </body>
  </html>
```

</details>

### Alternatives considered

Developers can download the arcgis-rest-samples repo and use that for quick testing, but sometimes setting up a second repo is too much work for small/quick verifications.

patrickarlt:
Closing this as done.

PR #1203, opened 2025-03-17
"Developer quick start" in Contributing.md instructions

gavinr-maps:
#1201


patrickarlt (approved):
@gavinr-maps this looks good but do you also want to add a sample based on the [`node-typescript-es-modules`](https://github.com/Esri/arcgis-rest-js-samples/tree/main/samples/node-typescript-es-modules) demo? This would also let us go some basic TypeScript work while only adding `ts-node` (or `tsx` which I'm coming to prefer)

Also do you want to move these to a "samples" or "workspace" folder if we are going have more then one of these?

gavinr-maps:
@patrickarlt that's updated, thanks!

Issue #1209, opened 2025-03-25
IUserRequestOptions.authentication should be an ArcGISIdentityManager


tomwayson:
### Describe the bug

[In 3.x `IUserRequestOptions.authentication` was a `UserSession`](https://esri.github.io/arcgis-rest-js/api/auth/IUserRequestOptions/#authentication), so in 4.x I would expect `IUserRequestOptions.authentication` to be an `ArcGISIdentityManager`, but it is an `IAuthenticatedRequestOptions`. 

I'm currently migrating from 3.x, and all of my code that uses `IUserRequestOptions` is broken b/c it expects `authentication` to have properties that were on `UserSession` (`token`, etc).

I don't believe that it was our intent to make `IUserRequestOptions` more generic, b/c it now has the exact same signature as `IAuthenticatedRequestOptions`, so this looks like a bug to me.

### Reproduction

n/a

### Logs

```shell

```

### System Info

```shell
4.3.0
```

tomwayson:
I see that `authentication` was changed to `IAuthenticationManager` in #1186.

I was able to get past my problem for now by going back to v4.2.3 (and the corresponding portal and feature-service packages).

Not sure whether that makes this a bug or not. Feels like it should be since my code works on 4.2.3 but would break if I upgrade to 4.3.0.

patrickarlt:
@tomwayson with the release of the new API key implementation last year API Keys can now perform any portal operations that they are scoped for so `IUserRequestOptions.authentication` should work for anything that that the `getToken()` and `getUser()` methods defined by `IAuthenticationManager`

I think your running into this because you are upgrading from 3.x. In 4.x you should generally rely on  the `.getToken()` method rather then the `token` getter if you need the token. We don't really use `.token` in the code anywhere. https://github.com/search?q=repo%3AEsri%2Farcgis-rest-js%20.token&type=code

I do agree there are 2 things that are weird:

1. `IUserRequestOptions` and `IAuthenticatedRequestOptions` are the same now which I missed in https://github.com/Esri/arcgis-rest-js/pull/1186.
2. Having `IAuthenticationManager` (the interface that defines a manager) and `AuthenticationManagerBase` (our shared implementation subclass) is a little confusing.

Is your code actually broken or is it just failing to compile and throwing type errors?

tomwayson:
Thanks @patrickarlt

> Is your code actually broken or is it just failing to compile and throwing type errors?

[...]

PR #1212, opened 2025-04-07
fix: add token property to IAuthenticationManager

patrickarlt:
This should resolve the issues in https://github.com/Esri/arcgis-rest-js/issues/1209 by adding the `token` getter back to `IAuthenticationManager`.`IUserAuthe

### DIFF
--- a/packages/arcgis-rest-demographics/test/getGeography.test.ts
+++ b/packages/arcgis-rest-demographics/test/getGeography.test.ts
@@ -5,6 +5,7 @@ import fetchMock from "fetch-mock";
 import { getGeography } from "../src/getGeography.js";
 
 const MOCK_AUTH = {
+  token: "token",
   getToken() {
     return Promise.resolve("token");
   },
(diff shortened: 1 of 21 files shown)


tomwayson (approved):
Thanks @patrickarlt 

This LGTM, and once merged should enable Hub to upgrade to the latest version of rest-js packages w/o a major re-write of `session.token` -> `await session.getToken()`.

The only thing I wonder about is that I don't believe this addresses either of the "weird" issues that you bring up in https://github.com/Esri/arcgis-rest-js/issues/1209#issuecomment-2759890162

Maybe you plan to address those at a later juncture.

tomwayson:
@patrickarlt 

I've pulled this down and tried it in Hub.js, but I'm still seeing build errors.

I've pushed a small fix in [c0f1e22](https://github.com/Esri/arcgis-rest-js/pull/1212/commits/c0f1e2201d20222ec7932565dfff72431cd88dbd) for one of the errors I was seeing.

Even after resolving that we still have several build errors. At least a few of these errors could end up being runtime errors, b/c our code does use at least one other property beyond `token` (`tokenExpires`) that does not exist on other authentication types.

At the end of the day, I still feel like `IUserRequestOptions` should have `authentication: ArcGISIdentityManager`, but if that has negative implications, or is not something that we're all on board w/, I have opened #1224 as a minimally invasive, and purely additive way to address #1209. If we merge that, I can do a find and replace of `IUserRequestOptions` w/ `IIdentityRequestOptions` in Hub.js.

patrickarlt:
closing this in favor of https://github.com/Esri/arcgis-rest-js/pull/1225

PR #1214, opened 2025-04-09
fix: account for authorizedCrossOriginNoCorsDomains

dbouwman:
Running tests locally things time out b/c `master` is on node 14, and on Apple silicon that means chrome runs in rosetta which is like 10,000x slower... so a lot of the tests time out 🤦🏼‍♂️

Will add much more detail once I've worked out the gaps in test coverage

### DIFF
--- a/packages/arcgis-rest-auth/test/UserSession.test.ts
+++ b/packages/arcgis-rest-auth/test/UserSession.test.ts
@@ -47,7 +47,7 @@ describe("UserSession", () => {
     expect(session2.refreshTokenTTL).toEqual(1440);
   });
 
-  describe(".getToken()", () => {
+  fdescribe(".getToken()", () => {
     it("should return unexpired tokens for trusted arcgis.com domains", (done) => {
       const session = new UserSession({
         clientId: "id",
@@ -262,6 +262,14 @@ describe("UserSession", () => {
         { repeat: 2, method: "POST" }
       );
 
+      // Needed since we now cache parts of portal/self in request to enable no-cors domains
+      fetchMock.getOnce(
+        "https://www.arcgis.com/sharing/rest/portals/self?f=json&token=new",
+        {
+          fakePortalResponse: true,
+        }
+      );
+
       Promise.all([
         session.getToken("https://www.arcgis.com/sharing/rest/portals/self"),
         session.getToken(
@@ -332,6 +340,14 @@ describe("UserSession", () => {
         token: "serverToken",
         expires: TOMORROW,
       });
+
+      // Needed since we now cache parts of portal/self in request to enable no-cors domains
+      fetchMock.getOnce(
+        "https://gis.city.gov/sharing/rest/portals/self?f=json&token=token",
+        {
+          fakePortalResponse: true,
+        }
+      );
 
       session
         .getToken(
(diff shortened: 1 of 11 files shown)


packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -262,6 +263,14 @@ describe("UserSession", () => {
...
       );
 
+      // // Needed since we now cache parts of portal/self in request to enable no-cors domains
+      // fetchMock.getOnce(
+      //   "https://www.arcgis.com/sharing/rest/portals/self?f=json&token=new",
+      //   {
+      //     fakePortalResponse: true,
+      //   }
+      // );
+
```
tomwayson, line 273: can we delete this?

packages/arcgis-rest-auth/test/UserSession.test.ts
```diff
@@ -451,6 +468,14 @@ describe("UserSession", () => {
...
       });
 
+      // // Needed since we now cache parts of portal/self in request to enable no-cors domains
+      // fetchMock.getOnce(
+      //   "https://gis.city.gov/sharing/rest/portals/self?f=json&token=token",
+      //   {
+      //     fakePortalResponse: true,
+      //   }
+      // );
+
```
tomwayson, line 478: can we delete this?

packages/arcgis-rest-request/src/request.ts
```diff
@@ -257,6 +264,25 @@ export function request(
...
+  // Check if a no-cors request is required. Just because it's a no-cors domain
+  // does not mean we need to make a no-cors request. Internally we track requests
+  // and only send them once an hour.
+  // Additionally we never need to send a no-cors request for same-origin requests
+
+  // default to false, for nodejs
+  let sameOrigin = false;
+  // if we are in a browser, check if the url is same origin
+  /* istanbul ignore else */
+  if (typeof window !== "undefined") {
```
tomwayson, line 281: we have another place where we check for the existence of `window` and also have to ignore coverage:

https://github.com/Esri/hub.js/blob/8c4f49b1ea5e32092f7459fa4549929ecbe3b298/packages/common/src/urls/_get-location.ts#L8

Maybe it's best to add a `isBrowser` util that we can stub for tests. If that's too much of a hassle here, maybe when we add th at to 4.x.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -268,25 +294,39 @@ export function request(
...
-         */
-        err.url = url;
-        err.options = options;
-        /**
-         * if an attempt is made to talk to an unfederated server
-         * first try the request anonymously. if a 'token required'
-         * error is thrown, throw the UNFEDERATED error then.
-         */
-        originalAuthError = err;
+  // TODO: Feels like  we should start w/ the noCors request if needed, THEN do the getToken
```
tomwayson, line 297: is this still a TODO?

packages/arcgis-rest-request/src/utils/isSameOrigin.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Used to determine if we need to do any additional cross-origin
+ * handling for the request.
+ * @param url
+ * @param win - optional window object to use for origin comparison
+ *             (useful for testing)
+ * @returns
+ */
+export function isSameOrigin(url: string, win?: Window | undefined): boolean {
+  /* istanbul ignore next */
+  if (!win && !window) {
```
tomwayson, line 12: another potential use for `isBrowser()`

tomwayson (approved):
as we discussed, I think the global name should be `ARCGIS_REST_JS_NO_CORS`, a couple of other minor comments, but all in all a good solution to a sticky probelm.

PR #1217, opened 2025-04-12
feat(arcgis-rest-portal): require `start` and `num` params

### LINKED ISSUE #853: [9.2 API Breaking Change] getUserContent requires pagination properties
Pagination change to `/sharing/rest/content/user/username`

This is a breaking change in the REST API and is not backward compatible.   Before this change in 9.2, the API could fetch all the items in a folder.  In folders with lots of items, the request was timing out.  Now, pagination is forced; the default is 1000 items per page.

Currently the `getUserContent` function has `start` and `num` as optional - we should make them required to ensure REST-JS is consistent w/ the Sharing API.

Sharing API release is slated for end of June 2021

arcenter:
Resolves #853.

This PR introduces a runtime check to ensure that the `getUserContent()` function aligns with the breaking change in the ArcGIS REST API (v9.2), which now requires pagination parameters. The parameters `start` and `num` are now required to run the function, removing defaults.

#### Changes
- Added a runtime check that checks whether `start` and `num` parameters are present in `content.ts`. An error is thrown if either is missing.
- Removed previous unit test in `content.test.ts` that verified defaulting of `start` and `num`.
- Added unit test in `content.test.ts` to verify the function throws when `start` and `num` parameters are absent.

#### Testing
- Ran `npm test` to confirm all tests pass.

### DIFF
--- a/packages/arcgis-rest-portal/src/items/content.ts
+++ b/packages/arcgis-rest-portal/src/items/content.ts
@@ -43,12 +43,14 @@ export interface IUserContentResponse extends IPagedResponse {
 export const getUserContent = (
   requestOptions: IUserContentRequestOptions
 ): Promise<IUserContentResponse> => {
-  const {
-    folderId: folder,
-    start = 1,
-    num = 10,
-    authentication
-  } = requestOptions;
+  const { folderId: folder, start, num, authentication } = requestOptions;
+
+  if (typeof start !== "number" || typeof num !== "number") {
+    return Promise.reject(
+      new Error("Both 'start' and 'num' are required for getUserContent.")
+    );
+  }
+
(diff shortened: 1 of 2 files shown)


gavinr-maps:
@arcenter thank you for the PR. After reviewing your code, I realized that the current code does already have defaults. So we may not need this PR, please see my comment here: https://github.com/Esri/arcgis-rest-js/issues/853#issuecomment-2805032322 ... will come back here and update the PR once we decide on that. Thank you again for the effort of this PR!

patrickarlt:
There are already defaults here. https://github.com/Esri/arcgis-rest-js/issues/853#issuecomment-2805032322

PR #1218, opened 2025-04-15
fix: search categories

gavinr-maps:
#685


patrickarlt (approved):
LGTM!

gavinr-maps:
I think I need to fix the unit tests - will do that ASAP and then merge.

PR #1219, opened 2025-04-15
Add node 22 to tests

gavinr-maps:
Node 22 is LTS right now.

https://nodejs.org/en/about/previous-releases

image


patrickarlt (approved):
LGTM!

PR #1224, opened 2025-04-30
feat(arcgis-rest-request): add IIdentityRequestOptions

### LINKED ISSUE #1209: IUserRequestOptions.authentication should be an ArcGISIdentityManager

tomwayson:
This is a minimally invasive change that will allow me to resolve #1209, though I would prefer we go w/ #1225

### DIFF
--- a/packages/arcgis-rest-request/src/authenticated-request-options.ts
+++ b/packages/arcgis-rest-request/src/authenticated-request-options.ts
@@ -1,24 +1,33 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { ApplicationCredentialsManager } from "./ApplicationCredentialsManager.js";
 import { ArcGISIdentityManager } from "./ArcGISIdentityManager.js";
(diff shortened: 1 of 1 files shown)


patrickarlt:
Closing this in favor if https://github.com/Esri/arcgis-rest-js/pull/1225

PR #1223, opened 2025-04-29
fix: postMessage must send a Credential that is usable by JSSDK

dbouwman:
In the 3.x -> 4.x transition, the structure returned as part of the postMessage auth flow changed. It needs to be shaped like a [Credential](https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-Credential.html) that the JS SDK can use to instantiate an IdentityMgr

Have verified this in Hub

PR #1227, opened 2025-05-09
Revert "fix: postMessage must send a Credential that is usable by JSSDK"

tomwayson:
Reverts Esri/arcgis-rest-js#1223

Turns out that currently some apps (dashboard) expect `messsage.data.credential` to be the shape of `idManager.toCredential()` , and others (survey 123) expect it to be the shape of `idManager.toJSON()`.

Also, in #1223, we only updated the function on parent side that sends the message, but we did **not** update [the function that an embedded app would use to create an ArcGISIdentityManager instance from the message](https://github.com/Esri/arcgis-rest-js/blob/d93bf0a8801b34a8125d988947f8128b5eaa2f4d/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L907) - that's still expecting the shape of `toJSON()`. 

Furthermore, we can't just update that to `ArcGISIdentityManager.fromCredential(message.data.credential)`, b/c that now requires a second argument of w/ server info - so I'm not sure what is the best way forward.

For now I propose that we revert the "fix" and at least apps that expect `toJSON()` will work while we come up w/ a more robust fix.

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -1317,7 +1317,7 @@ export class ArcGISIdentityManager
       if (isValidOrigin && isValidType) {
         let msg = {};
         if (isTokenValid) {
-          const credential = this.toCredential();
+          const credential = this.toJSON();
           msg = {
             type: "arcgis:auth:credential",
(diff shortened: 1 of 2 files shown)


tomwayson:
Closing in favor of #1228

PR #1228, opened 2025-05-09
fix(arcgis-rest-request): postMessage auth works w/ credential

tomwayson:
## Background

The Hub team is in the process of migrating from ArcGIS REST JS 3.x to 4.x and we discovered a change in the payload that is sent/expected when [passing authentication from a parent app to an embedded app via postMessage](https://esri.github.io/arcgis-rest-js/guides/embedded-apps/). For us, this manifested as a failure to pass authentication from the Hub to an embedded Dashboard, but surprisingly not to an embedded Survey123 application.
 
Furthermore, we consider this change in payload to be a bug, b/c the old payload, an [ICredential](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-request/ICredential/), was suitable for initiating authentication w/ either ArcGIS REST JS or the Maps SDK for JavaScript, whereas the new payload, an [IArcGISIdentityManagerOptions](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-request/IArcGISIdentityManagerOptions/), would need to be transformed to work with the Maps SDK for JavaScript.

We suspect that some ArcGIS apps either have not upgraded to 4.x, or have made accommodations to handle both the old and new payload formats.

## Proposal

We propose:
- The parent app _always_ sends the  original`ICredential` format that is compatible w/ the Maps SDK (i.e. can be passed directly to [esriId.registerToken()](https://developers.arcgis.com/javascript/latest/api-reference/esri-identity-IdentityManager.html#registerToken)), REST JS 3.x, and (as of this PR) REST JS 4.x
- A client app using 4.x will (as of this PR) expect the `ICredential` payload, but also fallback to handling the  `IArcGISIdentityManagerOptions` format that would have been sent by a 4.x parent that has not yet updated to the latest version.

## Status

We thought we addressed this in #1223, we missed a couple of things, which are now included in this PR:

1. [this line](https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/src/UserSession.ts#L1108)
2. Also, we only updated the function on parent side that sends the message (`createPostMessageHandler`), but we did **not** update [the function that an embedded app would use to create an ArcGISIdentityManager instance from the message (`parentMessageHandler`)](https://github.com/Esri/arcgis-rest-js/blob/d93bf0a8801b34a8125d988947f8128b5eaa2f4d/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts#L907) - that's still expecting the shape of `toJSON()`. 

I have confirmed that this worked in our app for at least dashboards and surveys:

image

image

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -904,7 +904,13 @@ export class ArcGISIdentityManager
    */
   private static parentMessageHandler(event: any): ArcGISIdentityManager {
     if (event.data.type === "arcgis:auth:credential") {
-      return new ArcGISIdentityManager(event.data.credential);
+      const credential = event.data.credential as ICredential;
+      const serverInfo = {
+        hasPortal: true,
+        hasServer: false,
+        server: credential.server
+      } as IServerInfo;
+      return ArcGISIdentityManager.fromCredential(credential, serverInfo);
     }
     if (event.data.type === "arcgis:auth:error") {
       const err = new Error(event.data.error.message);
@@ -1318,10 +1324,10 @@ export class ArcGISIdentityManager
         let msg = {};
         if (isTokenValid) {
           const credential = this.toCredential();
-          msg = {
-            type: "arcgis:auth:credential",
-            credential
-          };
+          // the following line allows us to conform to our spec without changing other depended-on functionality
+          // https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-auth/post-message-auth-spec.md#arcgisauthcredential
+          credential.server = credential.server.replace("/sharing/rest", "");
+          msg = { type: "arcgis:auth:credential", credential };
         } else {
           msg = {
             type: "arcgis:auth:error",
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -25,6 +25,16 @@ import { NODEJS_DEFAULT_REFERER_HEADER } from "./index.js";
 import { AuthenticationManagerBase } from "./AuthenticationManagerBase.js";
 
 /**
+ * distinguish between an ICredential and IArcGISIdentityManagerOptions
+ */
+function isCredential(credential: any): credential is ICredential {
+  return (
+    typeof credential.userId === "string" ||
+    typeof credential.expires === "number"
+  );
+}
+
+/**
  * Options for {@linkcode ArcGISIdentityManager.fromToken}.
  */
 export interface IFromTokenOptions {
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -904,7 +904,13 @@ export class ArcGISIdentityManager
...
   private static parentMessageHandler(event: any): ArcGISIdentityManager {
     if (event.data.type === "arcgis:auth:credential") {
-      return new ArcGISIdentityManager(event.data.credential);
+      const credential = event.data.credential as ICredential;
+      const serverInfo = {
+        hasPortal: true,
+        hasServer: false,
+        server: credential.server
+      } as IServerInfo;
+      return ArcGISIdentityManager.fromCredential(credential, serverInfo);
```
tomwayson, line 913: @patrickarlt it seems to me that the `serverInfo` argument to `fromCredential` could be optional b/c it's only used `if (serverInfo.hasServer)` - seems like we could change that to `if (serverInfo && serverInfo.hasServer) {` and make it optional.

B/c I wanted this fix to be minimally invasive, I left that out of this and just stubbed in a dummy serverInfo here, but if you agree, I could add that in a future PR.

dbouwman (approved):
This looks good!

tomwayson:
We have released this fix as `@esri/arcgis-rest-request@4.5.1-alpha.1` and deployed a version of the Hub application to QA that can be used to test it by adding `?uiVersion=PR-14801` to the URL of any Hub site in QA. See [the "Instructions for Testing" in the PR ](https://github.com/ArcGIS/opendata-ui/pull/14801)for examples.

tomwayson:
FYI - further evidence that this was likely not an intentional change, [the 4.x docs still show `ICredential` payload](https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-request/post-message-auth-spec.md).

dasa:
This looks good in my testing. 

These are the objects `fromParent` returns by version:

3.8.0

```json
{
  "username": "dpaddock",
  "token": "xyzzy...",
  "tokenExpires": "2025-05-15T21:52:16.152Z",
  "portal": "https://www.arcgis.com/sharing/rest",
  "ssl": true,
  "tokenDuration": 20160,
  "refreshTokenTTL": 20160
}
```

4.5.0

```json
{
  "token": "xyzzy...",
  "portal": "https://www.arcgis.com/sharing/rest",
  "ssl": true,
  "tokenDuration": 20160,
  "server": "https://www.arcgis.com"
}
```

4.5.0-alpha.1

```json
{
  "username": "dpaddock",
  "token": "xyzzy...",
  "tokenExpires": "2025-05-15T21:55:26.980Z",
  "portal": "https://www.arcgis.com/sharing/rest",
  "ssl": true,
  "tokenDuration": 20160
}
```

4.5.0-alpha.1 is almost the same as 3.8.0 except it doesn't include `refreshTokenTTL`.

PR #1232, opened 2025-06-25
Improve auth manager serialization

patrickarlt:
This PR adds a consistent `serialize()` and `deserialize()` methods to all auth managers as well as a `toJSON()` method. This also adds doc and examples for all methods. This also adds a `type` param to the serialized managers so you can tell them apart.

### DIFF
--- a/packages/arcgis-rest-request/src/ApiKeyManager.ts
+++ b/packages/arcgis-rest-request/src/ApiKeyManager.ts
@@ -33,7 +33,10 @@ export class ApiKeyManager
    */
   public readonly portal: string = "https://www.arcgis.com/sharing/rest";
 
-  private key: string;
+  /**
+   * The original API Key used to create this instance.
+   */
+  private readonly key: string;
 
   /**
    * The preferred method for creating an instance of `ApiKeyManager`.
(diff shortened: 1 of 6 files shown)


packages/arcgis-rest-request/src/ApiKeyManager.ts
```diff
@@ -52,11 +55,77 @@ export class ApiKeyManager
...
+   * import { ApiKeyManager } from '@esri/arcgis-rest-request';
+   *
+   * const apiKey = new ApiKeyManager.fromKey("...")
+   *
+   * localStorage.setItem("apiKey", apiKey.serialize());
+   * ```
+   * @returns {string} The serialized JSON string.
+   */
+  serialize() {
+    return JSON.stringify(this);
```
tomwayson, line 106: I think you meant for this to be 
```suggestion
    return JSON.stringify(this.toJSON());
```

packages/arcgis-rest-request/src/ApiKeyManager.ts
```diff
@@ -33,7 +33,10 @@ export class ApiKeyManager
    */
   public readonly portal: string = "https://www.arcgis.com/sharing/rest";
 
-  private key: string;
+  /**
+   * The original API Key used to create this instance.
+   */
+  private readonly key: string;
```
tomwayson, line 39: did you mean for this to be `public`?

packages/arcgis-rest-request/src/ApplicationCredentialsManager.ts
```diff
@@ -63,12 +64,12 @@ export class ApplicationCredentialsManager
...
-  private clientSecret: string;
-  private token: string;
-  private expires: Date;
-  private duration: number;
+  public readonly portal: string;
+  public readonly token: string;
+  public readonly clientId: string;
+  public readonly clientSecret: string;
+  public readonly expires: Date;
+  public readonly duration: number;
```
tomwayson, line 72: I'm wondering why you're making all of these `public`, especially since this PR makes it possible for consumers to get these values via  `const { token, clientId, clientSecret, expires, duration } = manager.toJSON()`

packages/arcgis-rest-request/src/ApplicationCredentialsManager.ts
```diff
@@ -63,12 +64,12 @@ export class ApplicationCredentialsManager
   extends AuthenticationManagerBase
   implements IAuthenticationManager
 {
-  public portal: string;
-  private clientId: string;
-  private clientSecret: string;
-  private token: string;
-  private expires: Date;
-  private duration: number;
+  public readonly portal: string;
```
tomwayson, line 67: I think this is a breaking change, no?

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -1205,9 +1237,25 @@ export class ArcGISIdentityManager
...
+   *
+   * // later, you can retrieve the manager from localStorage
+   * const serializedString = localStorage.getItem("arcgis-identity-manager");
+   * const manager = ArcGISIdentityManager.deserialize(serializedString);
+   * ```
+   *
+   * @returns The serialized JSON string.
+   */
   public serialize() {
     return JSON.stringify(this);
```
tomwayson, line 1256: I know this pre-dates your PR, but this should also probably have been:
```suggestion
    return JSON.stringify(this.toJSON());
```

tomwayson (commented):
All in all I think it's good to standardize the `toJSON()` and de/serialization functions. I left a few comments.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: This actually isn't necessary. `JSON.stringify()` calls this internally.

> If the value has a toJSON() method, it's responsible to define what data will be serialized. Instead of the object being serialized, the value returned by the toJSON() method when called will be serialized.
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

packages/arcgis-rest-request/src/ApiKeyManager.ts
patrickarlt: See my comment https://github.com/Esri/arcgis-rest-js/pull/1232#discussion_r2169485927

packages/arcgis-rest-request/src/ApiKeyManager.ts
patrickarlt: Since we have the public `token` property now and that is consistent across auth managers I think we should promote that as the public way to get and use a token. We could deprecate `key` and move entirely to `token` everywhere if we wanted.

packages/arcgis-rest-request/src/ApplicationCredentialsManager.ts
patrickarlt: I think it is nice to be able to get them without calling `toJSON()` since they will get documented as public properties in the doc.

packages/arcgis-rest-request/src/ApplicationCredentialsManager.ts
patrickarlt: Technically yes but if anyone is changing the `portal` property on the fly they probably have other problems. I would view this a bug/omission from ths original PR.

PR #1236, opened 2025-07-25
feat: support no cors requests

dbouwman:
PR to `alpha` so we get a release that will allow Hub to setup a built for verification in the no-cors test environment.


dbouwman:
bah. had this and the pr to `main` open in different windows, and merged the wrong one. multi-tasking--

I'm pretty sure this will all be fine, but an alpha release could have been nice. I'll get 4.7.0 into Hub and run our e2e's to double check

PR #1237, opened 2025-07-25
Add queryAllFeatures()

KarshitShah:
Added `queryAllFeatures()` to retrieve all the features.

### DIFF
--- a/packages/arcgis-rest-feature-service/src/query.ts
+++ b/packages/arcgis-rest-feature-service/src/query.ts
@@ -88,6 +88,61 @@ export interface IQueryFeaturesResponse extends IFeatureSet {
   exceededTransferLimit?: boolean;
 }
 
+/**
+ * query all features request options. See [REST Documentation](https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm) for more information.
+ */
+export interface IQueryAllFeaturesOptions extends ISharedQueryOptions {
+  objectIds?: number[];
+  relationParam?: string;
+  // NOTE: either time=1199145600000 or time=1199145600000, 1230768000000
+  time?: number | number[];
+  distance?: number;
+  units?: Units;
+  /**
+   * Attribute fields to include in the response. Defaults to "*"
+   */
+  outFields?: "*" | string[];
+  returnGeometry?: boolean;
+  maxAllowableOffset?: number;
+  geometryPrecision?: number;
+  // NOTE: either WKID or ISpatialReference
+  inSR?: string | ISpatialReference;
+  outSR?: string | ISpatialReference;
+  gdbVersion?: string;
+  orderByFields?: string;
+  groupByFieldsForStatistics?: string;
+  outStatistics?: IStatisticDefinition[];
+  returnZ?: boolean;
+  returnM?: boolean;
+  multipatchOption?: "xyFootprint";
+  resultOffset?: number;
+  resultRecordCount?: number;
+  // TODO: IQuantizationParameters?
+  quantizationParameters?: any;
+  resultType?: "none" | "standard" | "tile";
+  // to do: convert from Date() to epoch time internally
+  historicMoment?: number;
+  returnTrueCurves?: false;
+  sqlFormat?: "none" | "standard" | "native";
+  returnExceededLimitFeatures?: true;
+  /**
+   * Response format. Defaults to "json"
+   * NOTE: for "pbf" you must also supply `rawResponse: true`
+   * and parse the response yourself using `response.arrayBuffer()`
+   */
+  f?: "json" | "geojson" | "pbf";
+  /**
+   * someday...
+   *
+   * If 'true' the query will be preceded by a metadata check to gather info about coded value domains and result values will be decoded. If a fieldset is provided it will be used to decode values and no internal metadata request will be issued.
+   */
+  // decodeValues?: boolean | IField[];
+}
+
+export interface IQueryAllFeaturesResponse extends IFeatureSet {
+  exceededTransferLimit?: true;
+}
+
 export interface IQueryResponse {
   count?: number;
   extent?: IExtent;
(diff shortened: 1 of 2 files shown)


packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+ *   .then(result)
+ * ```
+ *
+ * @param requestOptions - Options for the request
+ * @returns A Promise that will resolve with the query response.
+ */
+export async function queryAllFeatures(
+  requestOptions: IQueryFeaturesOptions
+): Promise<IQueryFeaturesResponse> {
+  const pageSize = 2000;
```
patrickarlt: Thes size of the pages might vary between services. To find the size you should make| request to the service URL `http://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/3` and look at the `maxRecordCount` property.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+        "multipatchOption",
+        "resultOffset",
+        "resultRecordCount",
+        "quantizationParameters",
+        "returnCentroid",
+        "resultType",
+        "historicMoment",
+        "returnTrueCurves",
+        "sqlFormat",
+        "returnExceededLimitFeatures",
```
patrickarlt: This should always be forced to `true` so the response includes the `exceededTransferLimit` property.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+      }
+    );
+    const response: IQueryFeaturesResponse = await request(
+      `${cleanUrl(requestOptions.url)}/query`,
+      queryOptions
+    );
+
+    allFeatures = allFeatures.concat(response.features);
+
+    if (response.features.length < pageSize) {
```
patrickarlt: Instead of calculating this based on page size the officlai way to do this is via the `exceededTransferLimit` property on the response which you can see in these examples:

https://services.arcgis.com/xOi1kZaI0eWDREZv/arcgis/rest/services/Alternative_Fueling_Stations/FeatureServer/0/query?outFields=*&where=1=1&f=json


https://services.arcgis.com/xOi1kZaI0eWDREZv/arcgis/rest/services/Alternative_Fueling_Stations/FeatureServer/0/query?outFields=*&where=1=1&f=geojson

This also means that the `f` param should be forced to be `json` or `geojson` so that this property can be included.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+        "returnGeometry",
+        "maxAllowableOffset",
+        "geometryPrecision",
+        "inSR",
+        "outSR",
+        "gdbVersion",
+        "returnDistinctValues",
+        "returnIdsOnly",
+        "returnCountOnly",
+        "returnExtentOnly",
```
patrickarlt: These properties should not be allowed because they will alter the shape of the response to not support pagination.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+        "orderByFields",
+        "groupByFieldsForStatistics",
+        "outStatistics",
+        "returnZ",
+        "returnM",
+        "multipatchOption",
+        "resultOffset",
+        "resultRecordCount",
+        "quantizationParameters",
+        "returnCentroid",
```
patrickarlt: This properties should not be allowed because they will alter the shape of the response to not support pagination. This probably should mean that this new `queryAllFeatures()` function should have a separate interface for he params.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+        "returnCountOnly",
+        "returnExtentOnly",
+        "orderByFields",
+        "groupByFieldsForStatistics",
+        "outStatistics",
+        "returnZ",
+        "returnM",
+        "multipatchOption",
+        "resultOffset",
+        "resultRecordCount",
```
patrickarlt, line 333: since you are allowing the user to pass a custom value for `resultRecordCount` here you should also validate that the user supplied `resultRecordCount` is not higher then the services `maxRecordCount` and if it is use the `maxRecordCount`.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +204,106 @@ export function queryFeatures(
...
+      hasMore = false;
+    } else {
+      offset += pageSize;
+    }
+  }
+
+  return {
+    ...requestOptions,
+    features: allFeatures
+  };
```
patrickarlt: In order to keep the response formats consistent I think that this should reture a structure based on the values of the `f` parameter.

https://services.arcgis.com/xOi1kZaI0eWDREZv/arcgis/rest/services/Alternative_Fueling_Stations/FeatureServer/0/query?outFields=*&where=1=1&f=json

https://services.arcgis.com/xOi1kZaI0eWDREZv/arcgis/rest/services/Alternative_Fueling_Stations/FeatureServer/0/query?outFields=*&where=1=1&f=geojson

This means that if the user specifies `f=geojson` they will get a valid GeoJSON Feature Collection or if `f=json` a valid Esri formatted Feature Collection.

patrickarlt (changes requested):
@KarshitShah this looks good overall bet these are some edge cases in the REST API that I would like to be handled which I've detailed in my comment.

packages/arcgis-rest-feature-service/src/query.ts
patrickarlt: Full lost of parameters that should be disallowed because they change the shape of the response:

`returnDistinctValues`
`returnIdsOnly`
`returnCountOnly`
`returnExtentOnly`
`returnCentroid`

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -88,6 +88,61 @@ export interface IQueryFeaturesResponse extends IFeatureSet {
...
+  historicMoment?: number;
+  returnTrueCurves?: false;
+  sqlFormat?: "none" | "standard" | "native";
+  returnExceededLimitFeatures?: true;
+  /**
+   * Response format. Defaults to "json"
+   * NOTE: for "pbf" you must also supply `rawResponse: true`
+   * and parse the response yourself using `response.arrayBuffer()`
+   */
+  f?: "json" | "geojson" | "pbf";
```
patrickarlt: PBF won't support returning the `exceededTransferLimit`

```suggestion
  f?: "json" | "geojson";
```

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -204,3 +259,119 @@ export function queryFeatures(
...
+        "returnM",
+        "multipatchOption",
+        "resultOffset",
+        "resultRecordCount",
+        "quantizationParameters",
+        "resultType",
+        "historicMoment",
+        "returnTrueCurves",
+        "sqlFormat",
+        "returnExceededLimitFeatures",
```
patrickarlt:
```suggestion
```

I would remove this and always default it to `true` because we need this property in the response.

patrickarlt (commented):
@KarshitShah this is looking good. I have a few final comment before merging this for yo too take a look at.

patrickarlt:
closing done in https://github.com/Esri/arcgis-rest-js/pull/1246

PR #1240, opened 2025-07-28
fix(arcgis-rest-request): isSameOrigin returns false when url arg is null

rweber-esri:
* Fixes a JS error that is observed when `isSameOrigin` is called with a `null` value for it's `url` arg.

Context: Survey views in the Hub application appear to have regressed when [PR-15058](https://github.com/ArcGIS/opendata-ui/commit/63ee661c585c2d6f49606e5af6c6d67dbd6f173a) merged which bumped our `@esri/arcgis-rest-request` dependency from `4.5.1` to `4.7.1`. When this occurs, a JavaScript error is raised from `isSameOrigin` because a value of `null` is being passed as `url`, which is referenced from the survey's Stakeholder Feature Service View. This worked with V3 of RestJS and with `4.5.1`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33540,7 +33540,7 @@
     },
     "packages/arcgis-rest-request": {
       "name": "@esri/arcgis-rest-request",
-      "version": "4.6.0",
+      "version": "4.7.1",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


tomwayson (approved):
good catch @rweber-esri

PR #1241, opened 2025-08-08
Add typings for IAddItem, GetItemResources, and ILayerDefinition

gowin20:
Adding some types and interfaces to support the Maplibre ArcGIS project.

Originally drafted here: https://github.com/ArcGIS/maplibre-arcgis/blob/main/src/%40types/typings.d.ts

### DIFF
--- a/packages/arcgis-rest-feature-service/src/helpers.ts
+++ b/packages/arcgis-rest-feature-service/src/helpers.ts
@@ -743,7 +743,7 @@ export interface ILayerDefinition extends IHasZM {
   /** Boolean value indicating whether the layer supports the Calculate REST operation when updating features. */
   supportsCalculate?: boolean;
   /** Boolean value indicating whether the layer supports exceedsLimit in a query operation. */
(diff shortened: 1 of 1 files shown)


packages/arcgis-rest-feature-service/src/helpers.ts
```diff
@@ -741,6 +742,8 @@ export interface ILayerDefinition extends IHasZM {
   supportsAttachmentsByUploadId?: boolean;
   /** Boolean value indicating whether the layer supports the Calculate REST operation when updating features. */
   supportsCalculate?: boolean;
+  /** Boolean value indicating whether the layer supports exceedsLimit in a query operation. */
+  supportsExceedsLimitStatistics: boolean;
```
patrickarlt: Needs to be optional since this might not exist.

```suggestion
  supportsExceedsLimitStatistics?: boolean;
```

patrickarlt (changes requested):

Issue #1234, opened 2025-07-09
Add "get all features" method to arcgis-rest-feature-service


gowin20:
### Describe the problem

There is currently no method to retrieve all features from a feature service. There is also no pagination support for feature requests, as outlined in #1206.

While this was not previously realistic, recent updates to ArcGIS feature services now make this possible. See the JS Maps SDK's recently added "snapshot mode", for example.

### Describe the proposed solution

I propose a new method called `getAllFeatures` that queries a service URL and returns all of the features matching the query, rather than just 2,000. It would do this by making an initial request to the `exceedsLimit` statistic, and if the layer is within a reasonable limit, paginate through all features and return them. This would be useful in many applications and ensure best practices are used when many features are requested by a user.

gavinr-maps:
Thanks for logging this @gowin20! I think it's a common request.

This request is similar to #690. But I guess in #690 I was talking about the portal API instead of feature services. 

For Feature Services, there are prior discussions on this topic [here](https://github.com/Esri/geoservices-js/issues/35), [here](https://github.com/Esri/featureservice2trigger/issues/3), [here](https://github.com/Esri/arcgis-rest-js/pull/126#issuecomment-368526898), [here](https://gavinr.com/arcgis-javascript-query-all-features/), etc.

I've created a package that uses ArcGIS REST JS to do this: 
- [query-all-features](https://github.com/gavinr/query-all-features#query-all-features)
- demo: [codepen](https://codepen.io/gavinr/pen/ExQYegd?editors=0010)

So maybe it would be as easy as bringing that into ArcGIS REST JS?

But I'm not sure if we want to - I think ArcGIS REST JS has historically tried to stay close wrapping the API endpoints and not adding much functionality on top of it. But maybe this would be a good exception since it's a common request. I would look to @patrickarlt for more input.

gowin20:
@gavinr-maps thanks for the useful resources! @patrickarlt and I have been talking about this pretty extensively in the context of the new Maplibre ArcGIS plugin. We've got some pretty solid logic written already that uses the `exceedsLimit` statistic and requests all features as GeoJSON. Here's the code, I just gave you repo access: https://github.com/ArcGIS/maplibre-arcgis/blob/44ee78a914b3b49e26e64b00c3d654daa84a0df7/src/FeatureLayer.ts#L135

[...]

PR #1247, opened 2025-08-15
feat(arcgis-rest-feature-service): add support for f=geojson to queryAllFeatures()

patrickarlt:
This adds support for using `f=geojson` in `queryAllFeatures()` which I missed while reviewing the [original PR](https://github.com/Esri/arcgis-rest-js/issues/1246).

https://github.com/Esri/arcgis-rest-js/issues/1234

### DIFF
--- a/packages/arcgis-rest-feature-service/src/query.ts
+++ b/packages/arcgis-rest-feature-service/src/query.ts
@@ -367,7 +367,7 @@ export async function queryAllFeatures(
 
     const returnedCount = response.features.length;
 
-    // check if the response has exceededTransferLimit handles both thes stantard json and geojson responses
+    // check if the response has exceededTransferLimit handles both the standard json and geojson responses
     const exceededTransferLimit =
       response.exceededTransferLimit ||
       (response as any).properties?.exceededTransferLimit;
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -17,6 +17,7 @@ import {
   ISharedQueryOptions,
   IStatisticDefinition
 } from "./helpers.js";
+import { inspect } from "util";
```
gavinr-maps: Is this `inspect` necessary? I don't see it being used.

```suggestion
```

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -367,8 +368,13 @@ export async function queryAllFeatures(
 
     const returnedCount = response.features.length;
 
+    // check if the response has exceededTransferLimit handles both thes stantard json and geojson responses
```
gavinr-maps:
```suggestion
    // check if the response has exceededTransferLimit handles both the standard json and geojson responses
```

gavinr-maps (approved):
Overall looks good. A few small questions here:

PR #1249, opened 2025-08-20
Various fixes for sessions

patrickarlt:
This PR makes a number of changes to the newly released basemap session package:

* Change `startAutoRefresh()`/`stopAutoRefresh()` to `enableAutoRefresh()`/`disableAutoRefresh()` to avoid overuse of `start*` methods.
* Change `startExpirationChecking()`/`stopExpirationChecking()` to `enableExpirationChecking()`/`disableExpirationChecking()` to avoid overuse of `start*` methods.
* Throw errors if `duration` is under 30 seconds or over 12 hours.
* When the `expires` event fires it now disables expiration checking so that is cannot fire more then once per valid session by default. Refreshing a session now enables expiration checking by default.
* Adds a `secondsUntilExpiration` and `millisecondsUntilExpiration` getters
* `autoRefresh` is now `false` by default to allow users to opt into it.
* Calling `.refreshCredentials()` multiple times in a row now only makes one request for a new session.

### DIFF
--- a/packages/arcgis-rest-basemap-sessions/src/BaseSession.ts
+++ b/packages/arcgis-rest-basemap-sessions/src/BaseSession.ts
@@ -360,14 +360,6 @@ export abstract class BaseSession implements IAuthenticationManager {
 
     const now = new Date();
     const millisecondsLeft = this.endTime.getTime() - now.getTime();
-    console.log(
-      "now",
-      now,
-      "endTime",
-      this.endTime,
-      "millisecondsLeft",
-      millisecondsLeft
-    );
 
     return millisecondsLeft;
   }


packages/arcgis-rest-basemap-sessions/src/BaseSession.ts
```diff
@@ -312,15 +325,53 @@ export abstract class BaseSession implements IAuthenticationManager {
...
+    console.log(
+      "now",
+      now,
+      "endTime",
+      this.endTime,
+      "millisecondsLeft",
+      millisecondsLeft
+    );
+
+    return millisecondsLeft;
```
alaframboise, line 364: If you are worried about this being inaccurate because it is a snapshot in time, you could just have one getter and return something like this:

```
get timeRemaining(): number {
...
return { 
  atTime: now,
  millisecondsLeft: millisecondsLeft,
  secondsLeft: secondsLeft // arguably required, but handy
}
```

MarkTorrey (approved):

packages/arcgis-rest-basemap-sessions/src/BaseSession.ts
```diff
@@ -292,6 +295,16 @@ export abstract class BaseSession implements IAuthenticationManager {
     },
     SessionClass: new (params: IBasemapSessionParams) => T
   ): Promise<T> {
+    if (duration < 30) {
```
alaframboise: The min duration value for the REST API is 10 seconds. Probably should match that here.

https://mary0002-session-tokens-test-preview.gha.afd.arcgis.com/rest/basemap-styles/sessions-start-get/#durationseconds

alaframboise (changes requested):
Overall looks good. Few Suggestions. 

I'm still questioning why `expirationCheckInterval` is even settable and how this is helpful in the constructor. Especially if there's no event fired on each interval check. If you fired an event on for each interval, e.g., `checkExpiration` then I understand how it could be used on a client. 

And do we need to run an interval timer now that we have `secondsUntilExpiration()`? I can get the time remaining whenever I want...

@MarkTorrey and I both felt this should have been an internal setter, unless we are missing the use case as to why a client app would want to set this value and what the benefits are. Are we trying to let the developer control the internal timer interval to match their time check interval on the client app (if they have one)? Are we doing this for performance reasons? 

We must be missing something here. Let us know.

patrickarlt:
> I'm still questioning why `expirationCheckInterval` is even settable and how this is helpful in the constructor

@alaframboise `expirationCheckInterval` is a read only property ([declared here](https://github.com/Esri/arcgis-rest-js/pull/1249/files#diff-2bd296203a0baed7c5692af810fdd2f1ef1efa80af82befd7275e2cb4b79d938R166)) that needs to be set in the constructor so we can access it when when we need to start/stop expiration checking. Users can't set it when starting a session of creating a class manually and it scales with the `duration` property to make sure we aren't checking too often. So I guess I can just make this private.

packages/arcgis-rest-basemap-sessions/src/BaseSession.ts
```diff
@@ -312,15 +325,53 @@ export abstract class BaseSession implements IAuthenticationManager {
...
+   * Returns the number of milliseconds until the session token is no longer valid. If the session is expired, it will return 0.
+   */
+  get millisecondsUntilExpiration(): number {
+    if (this.isExpired) {
+      return 0;
+    }
+
+    const now = new Date();
+    const millisecondsLeft = this.endTime.getTime() - now.getTime();
+    console.log(
```
alaframboise: @patrickarlt  I would remove this logging, as it fills up the console. Unless there's a reason to leave it on that I'm missing. Are we trying to remind developers a timer is running and they should turn it off?

alaframboise (commented):
Consider unlogging timer messages.

alaframboise (approved):
Looks good! Tested with the dev guide session sample and passed all tests and use cases. Let me know when the package is republished.

Issue #934, opened 2021-10-27
Update Item doesn't support uploading thumbnail images


ssylvia:
[According to docs](https://developers.arcgis.com/rest/users-groups-and-items/update-item.htm), you can add an image file to the `thumbnail` param that should be uploaded as a thumbnail image.

This does not work:
```ts
updateItem({
  authentication: userSession,
  item: {
    ...otherItemDetails,
    thumbnail: MyBlob
  }
}),
```

However, if I use the standard request, it works:

```ts
request('https://www.arcgis.com/sharing/rest/content/users/jsmith/items/2b4c7a08699146f0b270ab609ec4aa91/update',{
  authentication: userSession,
  params: {
    ...otherItemDetails,
    thumbnail: MyBlob
  }
})
```

patrickarlt:
Closing since this is resolved.

Issue #333, opened 2018-09-21
Decorate all items with full thumbnail URL


patrickarlt:
When developing applications I am constantly doing:

```js
item.thumbnail = item.thumbnail
  ? `${portal}/content/items/${item.id}/info/${item.thumbnail}?${token}`
  : null;
```

since this is standard across everything so we should be able to do it any time we return an item from anywhere in REST JS and it would save all of us a lot of headache probably something more like this would handle `token` being optional for public items.

```js
let thumbnailURL =  item.thumbnail
  ? `${portal}/content/items/${item.id}/info/${item.thumbnail}`
  : null;

if(thumbnailURL && item.access !== "public") {
   thumbnailURL += `?${token}`;
}

item.thumbnail = thumbnailURL


```

jgravois:
you probably already saw it, but we're discussing the exact same thing for item resources in #322.

a generic utility method to assist with this would be _awesome_.

tomwayson:
I think @jgravois meant #323

jgravois:
yes. he did.

tomwayson:
See also: https://gist.github.com/tomwayson/230aeeeefea46e574f732639464d5cf8#file-index-html-L118

patrickarlt:
Done.

PR #1250, opened 2025-08-21
feat(arcgis-rest-portal): add support for uploading thumbnail and decorate items will full thumbnail url

KarshitShah:
Addresses #934 and #333

### DIFF
--- a/workspace/sample.ts
+++ b/workspace/sample.ts
@@ -15,7 +15,7 @@
 import { searchItems } from "../packages/arcgis-rest-portal/dist/esm/index.js";
 
 // If you are using node < 14.8, you need to wrap the await in an async IFFE
-// (async function () {
+//(async function () {
 const response = await searchItems("water");
 console.log(`Public item search for "water": ${response.total} total results`);
 response.results.map((itm) => {


packages/arcgis-rest-portal/src/items/update.ts
```diff
@@ -79,6 +79,28 @@ export function updateItem(
       requestOptions.params.extent = bboxToString(requestOptions.params.extent);
     }
 
+    // handle thumbnail separately if it's a Blob or File
+    const thumbnail = requestOptions.params?.thumbnail;
+    if (thumbnail instanceof Blob || thumbnail instanceof File) {
```
patrickarlt: @KarshitShah can you remove this and try passing the upload as a `Blob`? This code looks almost identical to https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-request/src/utils/encode-form-data.ts#L25 it might be worth trying to modify that code instead to handle files or require users to provide this as a blob.

KarshitShah: I’ve included a sample file. I tested updating the thumbnail using updateItem() in its original form and it works without any updates.

packages/arcgis-rest-portal/test/items/update.test.ts
```diff
@@ -215,6 +215,53 @@ describe("search", () => {
...
+        .then((response) => {
+          expect(fetchMock.called(itemUrl)).toBe(true);
+          const [, options] = fetchMock.lastCall(itemUrl)!;
+          const body = options.body as FormData;
+          const entries = Array.from((body as any).entries()) as [
+            string,
+            any
+          ][];
+          const thumbnailEntry = entries.find(([key]) => key === "thumbnail");
+          expect(thumbnailEntry).toBeDefined();
```
patrickarlt: I would change the to check that this is an instance of `Blob` and the data matches the mock to confirm that `updateItem()` sent the blob properly.

packages/arcgis-rest-portal/test/items/helpers.test.ts
```diff
@@ -58,3 +58,12 @@ describe("determineOwner()", () => {
     });
   });
 });
+
+describe("determineOwner()", () => {
```
patrickarlt:
```suggestion
describe("decorateThumbnail()", () => {
```

packages/arcgis-rest-portal/src/items/update.ts
```diff
@@ -79,6 +79,28 @@ export function updateItem(
...
+    //     }
+    //   });
+
+    //   // add thumbnail with filename
+    //   const filename =
+    //     thumbnail instanceof File ? thumbnail.name : "thumbnail.png";
+    //   formData.append("thumbnail", thumbnail, filename);
+    //   (requestOptions as any).body = formData;
+    // }
+
```
patrickarlt:
```suggestion
```

packages/arcgis-rest-portal/src/items/helpers.ts
```diff
@@ -294,3 +294,32 @@ export function isBBox(extent: unknown): boolean {
...
+    ? `${portal}/content/items/${item.id}/info/${item.thumbnail}`
+    : null;
+
+  if (thumbnailURL && item.access !== "public" && token) {
+    thumbnailURL += `?${token}`;
+  }
+
+  return {
+    ...item,
+    thumbnail: thumbnailURL
```
patrickarlt:
```suggestion
    thumbnailUrl: thumbnailURL
```

According to the docs `thumbnailUrl` will be downloaded and used to update item when passed so if you decorate it here you can still pass the results of `getItem()` back to `updateItem()` however all calls to `updateItem()` will then redownload and update the thumbnail.

patrickarlt: One way to resolve this would be to call `decorateThumbnail()` in `updateItem()` compare the result with the provided `thumbnailUrl`. If they match (excluding the `token`) then delete the `thumbnailUrl` from the params.

@dbouwman @tomwayson @gavinr

dbouwman: That seems reasonable @patrickarlt

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -46,7 +47,22 @@ export function getItem(
...
-  return request(url, options);
+
+  return request(url, options).then(async (item: IItem) => {
+    const portal = requestOptions?.portal || getPortalUrl(requestOptions);
+    let token: string | undefined;
+
+    // if authentication is provided and it’s an authentication manager and the item is not public then request a token for secure access
+    if (
+      requestOptions?.authentication &&
+      typeof requestOptions.authentication !== "string" &&
```
dbouwman, line 58: I'm not sure these two lines are needed - because if the item is not public, then auth must have been passed in, otherwise you would get the item back.

tomwayson (commented):
a agree w/ the suggestions from @dbouwman and @patrickarlt

packages/arcgis-rest-portal/src/items/helpers.ts
patrickarlt: @KarshitShah lets implement the logic as I have it defined above

packages/arcgis-rest-portal/src/items/get.ts
patrickarlt: @dbouwman I think this is necessary to make sure that for non-public items we get a valid token to work with for `decorateThumbnail()`

packages/arcgis-rest-portal/src/items/helpers.ts
patrickarlt: @KarshitShah this is a key change that I'm still not seeing in this PR. The full thumbnail URL needs to be set at `thumbnailUrl` so what you can compare the decorated `thumbnail` property vs the user provided `thumbnailUrl` in update item. If they match then you can delete `thumbnailUrl` because it is set to the same value and we can assume that the user doesn't want to update it.

packages/arcgis-rest-portal/src/items/update.ts
```diff
@@ -79,6 +80,29 @@ export function updateItem(
...
+    const paramsThumbnail = requestOptions.params.thumbnail;
+
+    if (
+      typeof decoratedThumbnail.thumbnail === "string" &&
+      typeof paramsThumbnail === "string"
+    ) {
+      if (
+        decoratedThumbnail.thumbnail.split("?")[0] ===
+        paramsThumbnail.split("?")[0]
```
patrickarlt: @KarshitShah this comparison will need to change to reflect https://github.com/Esri/arcgis-rest-js/pull/1250/files#r2334514361

patrickarlt (changes requested):
@KarshitShah This needs a few more changes when you have a chance to look.

patrickarlt:
@KarshitShah here is an example of what we want to be able to support with this:

```js
const item = await getItem("item id")
// will have thumbnail and thumbnailUrl properties
item.description = "New description";

// thumbnail and thumbnailUrl wilbe compared, will be the same
// delete thumbnailUrl so the thumbnail will not be updated
updateItem(item);

//----

const item = await getItem("item id")
// will have thumbnail and thumbnailUrl properties
item.thumbnail = new Blob(["new thumbnail data"]);

// thumbnail is not a string, so it will be updated
updateItem(item);

//----

const item = await getItem("item id")
// will have thumbnail and thumbnailUrl properties
item.thumbnailUrl = "https://new-thumbnail-url.com/image.png";

// thumbnail and thumbnailUrl do not match
// so the thumbnail will be updated and the new thumbnailUrl will be used
updateItem(item);
```

packages/arcgis-rest-portal/test/items/update.test.ts
```diff
@@ -215,6 +215,150 @@ describe("search", () => {
...
+            "https://myorg.maps.arcgis.com/sharing/rest/content/items/5bc/info/thumbnail.png"
+        },
+        authentication: MOCK_USER_SESSION
+      })
+        .then(() => {
+          expect(fetchMock.called(itemUrl)).toBe(true);
+          const [, options] = fetchMock.lastCall(itemUrl)!;
+          const body = options.body as any;
+          let thumbnailValue: any;
+          if (body instanceof FormData) {
```
patrickarlt: @KarshitShah I don't think that this `if` statement is necessary. The tests should be deterministic and this should always produce the same result and code path so the tests should not handle variations like this. Only one of these things should be true for every run of the tests.

packages/arcgis-rest-portal/test/items/update.test.ts
```diff
@@ -215,6 +215,150 @@ describe("search", () => {
...
+          expect(fetchMock.called(itemUrl)).toBe(true);
+          const [, options] = fetchMock.lastCall(itemUrl)!;
+          const body = options.body as any;
+          let thumbnailIsBlob = false;
+          if (body instanceof FormData) {
+            const entries = Array.from((body as any).entries()) as [
+              string,
+              any
+            ][];
+            const thumbEntry = entries.find(([key]) => key === "thumbnail");
```
patrickarlt:
```suggestion
            const thumbEntry = (body as any).get("thumbnail");      
```

`FormData` has a getter that you can use instead of this. https://developer.mozilla.org/en-US/docs/Web/API/FormData/get

packages/arcgis-rest-portal/test/items/update.test.ts
```diff
@@ -215,6 +215,150 @@ describe("search", () => {
...
+          expect(fetchMock.called(itemUrl)).toBe(true);
+          const [, options] = fetchMock.lastCall(itemUrl)!;
+          const body = options.body as any;
+          let thumbnailValue: any;
+          if (body instanceof FormData) {
+            const entries = Array.from((body as any).entries()) as [
+              string,
+              any
+            ][];
+            const thumb = entries.find(([key]) => key === "thumbnailUrl");
```
patrickarlt: See my previous comment to simplify this lookup.

patrickarlt (changes requested):
@KarshitShah I have a few small changes to the tests. I also checked this against a love item on production and it worked great!

Can you also revert the changes to the workspace sample so we can keep a clean environment?

Issue #1251, opened 2025-09-02
Migrate from Karma/Jasmine to Vitest


patrickarlt:
In https://github.com/Esri/arcgis-rest-js/issues/1187 I did some research on what it would take to update to TypeScrpit 5 and determined that we cannot (or that it would be really difficult too) without migrating away from Karma which only supports TypeScript 4.

This issues is to capture the work needed to migrate all of our tests to [Vitest](https://vitest.dev/guide/). I already set this up with the addition of the basemap sessions package in #1239 and it works well.

In general I would like to do a separate PR to migrate each packages test starting with least to most complex:

* [ ] Elevation
* [ ] Places
* [ ] Demographics
* [ ] Geocoding
* [ ] Routing
* [ ] Feature Service
* [ ] Developer Credentials
* [ ] Portal
* [ ] Request
* [ ] [Live Tests](https://github.com/Esri/arcgis-rest-js/blob/main/jasmine.live.json)

The migration mostly involves:

* Remove the package to be migrated from the `jasmine.json` and `karma.conf.cjs` files.
* Adding the imports from `vitest` to the top of each test file.
* Rename from `it()` to `test()`
* Any other minor renaming as necessary
* Migrate anything like spys or clock functions to their `vitest` equivalents.

Once we finish this we will only run one test suite (mostly in Node) via Vitest instead of the same test suite in both Node and browsers. I would like to still test the browser only functions like oAuth authentication using either [browser mode in `vitest`](https://vitest.dev/guide/browser/) or just bu manually invoking playwright in `vitest` directly.

dixonyant:
Completed, moving Live Tests to a separate issue.

PR #1252, opened 2025-09-02
chore: migrate elevation tests to ArcGIS REST JS

patrickarlt:
This PR migrates the elevation tests to Vitest as an example of https://github.com/Esri/arcgis-rest-js/issues/1251

This also makes some changes to reduce console noise in the tests by passing a auth manager instead of a raw token.

### DIFF
--- a/jasmine.json
+++ b/jasmine.json
@@ -3,7 +3,6 @@
   "spec_files": [
     "arcgis-rest-demographics/{src,test}/**/*.test.ts",
     "arcgis-rest-developer-credentials/{src,test}/**/*.test.ts",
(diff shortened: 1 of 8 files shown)


tomwayson (commented):
Love it! Vitest is where we want to go w/ Hub.js too.

So now we'll have 2 code coverage checks on PRs, right? One for the un-migrated Jest/Karma tests and one for Vitest?

patrickarlt:
@tomwayson Yup we will have 2 coverage reports one from Karma and one from Vitest. Thankfully the new PR check combines them both for us https://github.com/Esri/arcgis-rest-js/blob/main/.github/workflows/pull-request-tests.yml#L47-L55

Issue #1253, opened 2025-09-13
`request` displays incorrect warnings when API key or session tokens are passed


gowin20:
### Describe the bug

It looks like `request` is not recognizing the new format for API keys, or for session tokens. It displays a warning when they are passed as strings. I believe it is due to this line: https://github.com/Esri/arcgis-rest-js/blob/d2170ee88312891ab779eb4ca6d084e03f357189/packages/arcgis-rest-request/src/request.ts#L291

### Reproduction

1. Create a new app with REST JS request.
2. Call `request` and pass an API key as a string to the `authentication` field.
3. See console warning about OAuth 2.0 token.

### Logs

```shell

```

### System Info

```shell
"@esri/arcgis-rest-request": "^4.7.2",
```

patrickarlt:
This should be resolved in `@esri/arcgis-rest-request@4.7.3`. @gowin20 will you confirm and close this.

PR #1254, opened 2025-09-13
add token prefixes to oauth conditional

gowin20:
Fix for #1253 

Prevent warnings about OAuth 2.0 tokens showing up when you pass an API key or basemap session token.


patrickarlt (approved):
LGTM!

PR #1256, opened 2025-09-17
Fix #1231 by allowing a custom state object to opt out of originalUrl behavior

### LINKED ISSUE #1231: ArcGIS REST-JS will always set herf as state.originalUrl to pass to rest api when authorizing, which will limit the length of url
### Describe the problem

ArcGIS Experience Builder supports recording app states, where it may result in a very long app url. 
While ArcGIS REST-JS will always set `herf `as `state.originalUrl` to pass to rest api when authorizing, which will limit the length of url. 

See https://devtopia.esri.com/Beijing-R-D-Center/ExperienceBuilder/issues/23772#issuecomment-5575414 

### Describe the proposed solution

We hope rest-js can enhance it to support long URL. 

### Alternatives considered

_No response_

### Additional Information

Experience Builder teams is expecting this issue can be solve in 2025 R3 Online release

User requirement: 
https://desktopqa/SFTriage?sfid=BUG-000174587
https://desktopqa/SFTriage/#/BUG-000176443

patrickarlt:
This PR allows for a workaround to avoid #1231. This allows a custom object to be passed as the `state` property in `.beginOauth2()`. If an object is passed the `originalUrl` will not be encoded into the `state`.

```ts
ArcGISIdentityManager.beginOAuth2({
  cilentId,
  redirectUrl,
  state: {
    // an empty object will work to remove the originalUrl behavior
  }
});
```

Then in order to make `state` a little more useful it is also exposed on the resulting `ArcGISIdentityManager` instance so it can be accessed after the oAuth process is complete.

### DIFF
--- a/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
+++ b/packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
@@ -179,7 +179,7 @@ export interface IOAuth2Options {
   style?: "" | "light" | "dark";
 
   /**
-   * Custom value for oAuth 2.0 state. A random identifier will be generated if this is not passed. You may also pass a plain object and the `id` property will be used as the state value. Of an object is passed all properties will be serialized and made available when the {@linkcode ArcGISIdentityManager.completeOAuth2} method is called. If a string is passed it will be available as the `id` property and an additional `originalUrl` property will be added containing the URL of the page that initiated the oAuth 2.0 process.
+   * Custom value for oAuth 2.0 state. A random identifier will be generated if this is not passed. You may also pass a plain object and the `id` property will be used as the state value. If an object is passed, all properties will be serialized and made available when the {@linkcode ArcGISIdentityManager.completeOAuth2} method is called. If a string is passed, it will be available as the `id` property and an additional `originalUrl` property will be added containing the URL of the page that initiated the oAuth 2.0 process.
    */
   state?: string | IOAuthState;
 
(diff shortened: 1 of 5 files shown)


patrickarlt:
@ke1905 @wei8123 Sorry for the massive delay on this. This should resolve #1231. Once this PR merges it will automatically release `4.7.3` which should have the fix.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -178,9 +179,9 @@ export interface IOAuth2Options {
   style?: "" | "light" | "dark";
 
   /**
-   * Custom value for oAuth 2.0 state. A random identifier will be generated if this is not passed.
+   * Custom value for oAuth 2.0 state. A random identifier will be generated if this is not passed. You may also pass a plain object and the `id` property will be used as the state value. Of an object is passed all properties will be serialized and made available when the {@linkcode ArcGISIdentityManager.completeOAuth2} method is called. If a string is passed it will be available as the `id` property and an additional `originalUrl` property will be added containing the URL of the page that initiated the oAuth 2.0 process.
```
gavinr-maps:
```suggestion
   * Custom value for oAuth 2.0 state. A random identifier will be generated if this is not passed. You may also pass a plain object and the `id` property will be used as the state value. If an object is passed, all properties will be serialized and made available when the {@linkcode ArcGISIdentityManager.completeOAuth2} method is called. If a string is passed, it will be available as the `id` property and an additional `originalUrl` property will be added containing the URL of the page that initiated the oAuth 2.0 process.
```

PR #1257, opened 2025-09-17
chore: migrate arcgis-rest-places package tests to Vitest

dixonyant:
This PR migrates the places package tests to Vitest as part of #1251 

This also makes the same changes as #1252 updating tests to use an auth manager for authentication.

### DIFF
--- a/jasmine.json
+++ b/jasmine.json
@@ -3,7 +3,6 @@
   "spec_files": [
     "arcgis-rest-demographics/{src,test}/**/*.test.ts",
     "arcgis-rest-developer-credentials/{src,test}/**/*.test.ts",
-    "arcgis-rest-elevation/{src,test}/**/*.test.ts",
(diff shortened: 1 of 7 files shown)


packages/arcgis-rest-places/test/utils.test.ts
```diff
@@ -1,9 +1,11 @@
+import { hasNextPage } from "../src/utils.js";
+import { describe, test, expect } from "vitest";
+// unused imports?
 import { ApiKeyManager } from "@esri/arcgis-rest-request";
 import fetchMock from "fetch-mock";
```
dixonyant: the above two imports aren't used in this file, should i remove them?

patrickarlt (approved):
LGTM! Merging! @dixonyant will you update this on the internal tracker?

PR #1259, opened 2025-09-18
chore: migrating tests to Vitest

dixonyant:


packages/arcgis-rest-demographics/test/queryDemographicData.test.live.ts
dixonyant, line 1: not sure what to do with this file, i didn't see it get picked up by Vitest, so I might have to add the .live extension to the include. Don't know if it was a jasmine/karma-only thing with the jasmine.live.json configuration file include. Would appreciate feedback.

dixonyant (commented):
commented on live tests file

packages/arcgis-rest-demographics/test/queryDemographicData.test.live.ts
patrickarlt: @dixonyant the live tests have fallen by the wayside a bit. The original intent was to perform tests against live APIs and scenarios for additional safety without relying on mocks.

To test these locally you would create a `.env` file locally and and then run `npm run test:node:live` from https://github.com/Esri/arcgis-rest-js/blob/main/package.json#L31 which uses a different Jasmine config at https://github.com/Esri/arcgis-rest-js/blob/main/jasmine.live.json

I think we should do something similar once everything else is wrapped up. Since the live tests require credentials to run. Ideally I would like to also work on running these with credentials owned by us (Esri) so users who contribute don't need to set this up.

patrickarlt (approved):
LGTM! I think we can merge this with the changes to live tests and get them running after all the other tests are done.

packages/arcgis-rest-demographics/test/queryDemographicData.test.live.ts
dixonyant: I totally agree, especially after we finish this first pass. I can keep that in mind.

PR #1262, opened 2025-09-19
chore: make the job tests less noisy

patrickarlt:
This PR makes the job tests quieter by disabling polling during the tests to reduce unmatched requests in fetch mock. In a lot of cases I feel like this shouldn't be happening anyway because we have a lot of generic `*` handlers.

This mostly reworks the test to still test the same cases but without the polling that was causing extra requests. This also adds a utility function to create mocks with different ids so we can differentiate requests between test more easily.

### DIFF
--- a/packages/arcgis-rest-request/test/job.test.ts
+++ b/packages/arcgis-rest-request/test/job.test.ts
@@ -12,45 +12,88 @@ import {
   mockAllResults
 } from "./mocks/job-mock-fetches.js";
 
+function createJobMocks(
+  jobId: string = "MOCK_JOB_ID",
+  serviceName: string = "MOCK_SERVICE"
+) {
+  const baseUrl = `https://www.arcgis.com/arcgis/rest/services/${serviceName}/GPServer/Task/`;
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-request/test/job.test.ts
```diff
@@ -453,34 +525,46 @@ describe("Job", () => {
   });
 
   it("it gets the results however we wait for the job to succeeded them show the results", () => {
```
dixonyant, line 527: i'd be interested to know more about what these cases do. The description is a bit hard to understand, but I'm sure i'll learn more how it's set up soon enough.

dixonyant (approved):
Looks good, love the abstraction of variables

packages/arcgis-rest-request/test/job.test.ts
patrickarlt: A lot of these tests are various tests about how the lifecycle of these types of requests work. If you want we can meet and step through this when you get to migrating the tests for the request packages.

PR #1264, opened 2025-09-22
Routing test migration

dixonyant:


packages/arcgis-rest-routing/test/solveRoute.test.ts
```diff
@@ -153,34 +147,28 @@ describe("solveRoute", () => {
...
+    expect(response.routes.features[0].attributes.Name).toEqual(
+      "Location 1 - Location 2"
+    );
   });
 
-  it("should make a simple solveRoute request (array of 3d stops)", (done) => {
+  test("should make a simple solveRoute request (array of 3d stops)", async () => {
     fetchMock.once("*", Solve);
 
     const MOCK_AUTH = {
```
dixonyant, line 174: I know there's a strong case for mocking the auth object per test...what are thoughts between having simple-auth on a per-test-case basis vs pulling it out to the top as we had in one of the previous packages?

packages/arcgis-rest-routing/test/closestFacility.test.ts
```diff
@@ -172,24 +173,22 @@ describe("closestFacility", () => {
...
+  test("should throw an error when a closestFacility request is made without a token", async () => {
     fetchMock.once("*", ClosestFacility);
 
-    closestFacility({
-      incidents,
-      facilities,
-      returnCFRoutes: true
-    })
-      // tslint:disable-next-line
-      .catch((e) => {
```
dixonyant, line 184: Vitest automatically fails tests that return an unhandled rejection or error with async await changes

packages/arcgis-rest-routing/test/closestFacility.test.ts
```diff
@@ -199,42 +198,37 @@ describe("closestFacility", () => {
...
-        expect(options.body).toContain("token=token");
-
-        expect(response.routes.spatialReference.latestWkid).toEqual(4326);
-        expect(response.routes.features[0].attributes.Name).toEqual(
-          "Echo Park Ave & W Sunset Blvd, Los Angeles, California, 90026 - Flint Wash Trail"
-        );
-        done();
-      })
-      .catch((e) => {
-        fail(e);
```
dixonyant, line 233: info: from what I can gather, Vitest automatically fails tests that return an unhandled rejection or error within async functions.

packages/arcgis-rest-routing/test/solveRoute.test.ts
patrickarlt: It would be a lot simpler to have one mock in most cases. Generally we are just testing that the `token` value gets passed through to the URL. There are a bunch of other tests in `request` that make sure the `getToken()` method on the authentication object is called and the resulting value used for the token.

patrickarlt: Once we migrate everything it might be a good idea to see if can centralize basic mocks and fixtures with `test.extend` so all tests can share the same basic API key and identity manager fixtures https://vitest.dev/guide/test-context.html#test-extend

packages/arcgis-rest-routing/test/closestFacility.test.ts
patrickarlt: I'm fairly sure that this is also the default behavior in Jasmine too so this was probably unneccessary.

PR #1265, opened 2025-09-24
chore: migrate arcgis-rest-feature-service package tests to Vitest

dixonyant:
This PR migrates the routing package tests to Vitest as part of https://github.com/Esri/arcgis-rest-js/issues/1251

### DIFF
--- a/packages/arcgis-rest-feature-service/test/attachments.test.ts
+++ b/packages/arcgis-rest-feature-service/test/attachments.test.ts
@@ -1,6 +1,8 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
+import { FormData } from "@esri/arcgis-rest-form-data";
+import { describe, afterEach, test, expect } from "vitest";
 import fetchMock from "fetch-mock";
 import { attachmentFile } from "../../../scripts/test-helpers.js";
 import {
@@ -13,7 +15,6 @@ import {
   deleteAttachments,
   IDeleteAttachmentsOptions
 } from "../src/index.js";
-
 import {
   getAttachmentsResponse,
   addAttachmentResponse,
@@ -21,7 +22,7 @@ import {
   deleteAttachmentsResponse,
   genericInvalidResponse
 } from "./mocks/feature.js";
-import { FormData } from "@esri/arcgis-rest-form-data";
+
 const serviceUrl =
   "https://services.arcgis.com/f8b/arcgis/rest/services/Custom/FeatureServer/0";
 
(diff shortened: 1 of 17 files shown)


packages/arcgis-rest-feature-service/test/attachments.test.ts
```diff
@@ -13,177 +15,129 @@ import {
...
 
 describe("attachment methods", () => {
   afterEach(() => {
     fetchMock.restore();
   });
-  it("should return an array of attachmentInfos for a feature by id", (done) => {
-    const requestOptions = {
+
+  test("should return an array of attachmentInfos for a feature by id", async () => {
+    const requestOptions: IGetAttachmentsOptions = {
```
dixonyant, line 35: Thought it would be better to type cast on declaration for better type-checking and handling changes better.

packages/arcgis-rest-feature-service/test/attachments.test.ts
```diff
@@ -13,177 +15,129 @@ import {
...
-      }
-    } as IAddAttachmentOptions;
+      params: { returnEditMoment: true }
+    };
     fetchMock.once("*", genericInvalidResponse);
-    addAttachment(requestOptions)
-      .then(() => {
-        // nothing to test here forcing error
-        fail();
+    await expect(addAttachment(requestOptions)).rejects.toMatchObject({
```
dixonyant, line 87: cleaned up this logic to just throw the error and use vitest testing objects to test the response for included fields.

packages/arcgis-rest-feature-service/test/decodeValues.test.ts
```diff
@@ -19,49 +19,53 @@ describe("formatCodedValues()", () => {
...
-      })
-      .catch((e) => {
-        fail(e);
-      });
+    });
+
+    expect(response.features[0]).toEqual(cvdFeaturesFormatted[0]);
+  });
+
+  test("should fetch metadata then skip decoding if no cvd codes", async () => {
```
dixonyant, line 53: The reason for adding this test is because there is an if-clause in the decodeValues function that was previously under an instanbul-ignore statement. I think a linter is removing that line or disregarding it after the change from jasmine to vitest. My solutions are either to make a test that covers that clause as above (then remove the ignore comment), or to add "-- preserve" to the ignore line and keep previous behavior.

packages/arcgis-rest-feature-service/test/getServiceAdminInfo.test.ts
```diff
@@ -22,20 +23,19 @@ describe("getServiceAdminInfo: ", () => {
...
 
-    return getServiceAdminInfo(
+    const response = await getServiceAdminInfo(
       "https://servicesqa.arcgis.com/orgid/arcgis/rest/services/mysevice/FeatureServer",
       MOCK_USER_SESSION
-    ).then((result) => {
-      const [url] = fetchMock.lastCall("*");
+    );
 
-      expect(result.foo).toBe("bar", "should return the api response");
```
dixonyant, line 34: vitest deprecates the secondary fail message.

packages/arcgis-rest-feature-service/test/attachments.test.ts
```diff
@@ -13,177 +15,129 @@ import {
...
+    expect(options.method).toBe("POST");
 
-        const params = options.body as FormData;
-        expect(params instanceof FormData).toBeTruthy();
-        // we can introspect FormData in Chrome this way, but not Node.js
-        // more info: https://github.com/form-data/form-data/issues/124
-        if (params.get) {
-          expect(params.get("returnEditMoment")).toEqual("true");
-        }
-        expect(addAttachmentResponse.addAttachmentResult.objectId).toEqual(
```
dixonyant, line 84: this seems to be testing the mock object, not the one returned by the response object. I changed it to test the response object but thought it worth a notice.

packages/arcgis-rest-feature-service/test/attachments.test.ts
```diff
@@ -13,177 +15,129 @@ import {
...
+    expect(fetchMock.called()).toBeTruthy();
+    const [url, options] = fetchMock.lastCall("*");
+    expect(url).toEqual(
+      `${requestOptions.url}/${requestOptions.featureId}/deleteAttachments`
+    );
+    expect(options.body).toContain("attachmentIds=1001");
+    expect(options.body).toContain("returnEditMoment=true");
+    expect(options.method).toBe("POST");
+    expect(response.deleteAttachmentResults[0].objectId).toBe(1001);
+    expect(response.deleteAttachmentResults[0].success).toBe(true);
```
dixonyant, line 141: changed toEqual to toBe for simple types since we don't need to check for deep equality.

packages/arcgis-rest-feature-service/test/getServiceAdminInfo.test.ts
```diff
@@ -22,20 +23,19 @@ describe("getServiceAdminInfo: ", () => {
     portal: "https://myorg.maps.arcgis.com/sharing/rest"
   });
 
-  it("makes request to the admin url", () => {
+  test("should return the api response when making a request to the admin url", async () => {
```
dixonyant, line 26: moved secondary fail message to test description.

packages/arcgis-rest-feature-service/test/decodeValues.test.ts
dixonyant: https://github.com/vitest-dev/vitest/issues/2021

packages/arcgis-rest-feature-service/test/attachments.test.ts
patrickarlt: Good catch!

packages/arcgis-rest-feature-service/test/attachments.test.ts
patrickarlt: I think this a better way to do it. Thanks!

packages/arcgis-rest-feature-service/test/attachments.test.ts
patrickarlt: Looks good!

packages/arcgis-rest-feature-service/test/attachments.test.ts
patrickarlt: Looks good!

packages/arcgis-rest-feature-service/test/decodeValues.test.ts
patrickarlt: This function is really weird looking at it again. I think we could probably refactor it to simplify it in v5 with `structuredClone()` and `Object.keys()`. For now I think that this test looks good though.

packages/arcgis-rest-feature-service/test/getServiceAdminInfo.test.ts
patrickarlt: Looks good.

packages/arcgis-rest-feature-service/test/decodeValues.test.ts
dixonyant: sweet, since this is good i'll remove that ignore from the other file before merging

PR #1267, opened 2025-10-06
Portal test migration chore: migrate and update tests to Vitest

dixonyant:
This is part of the https://github.com/Esri/arcgis-rest-js/issues/1251 test migration work.

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -390,7 +390,7 @@ export function internalRequest(
         // dont append a '?' unless parameters are actually present
         /* istanbul ignore next */
         const urlWithQueryString =
-          queryParams === "" ? url : url + "?" + encodeQueryString(params);
+          queryParams === "" ? url : `${url}?${queryParams}`;
 
         if (
           // This would exceed the maximum length for URLs by 2000 as default or as specified by the consumer and requires POST


packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -108,6 +108,7 @@ export function getItemData(
     ...requestOptions
   };
 
+  /* istanbul ignore next -- @preserve */
```
dixonyant: ~~When karma dropped these tests, we lost coverage on this branch and the throw error branch. I looked into writing tests for these two cases but wanted to open the floor for discussion on an approach to these changes in general, as moving to Vitest means we are gonna have to test in Node, mock browser objects,  or separate the tests for run environments. Do we want make those changes in these PR's?~~

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -118,7 +119,7 @@ export function getItemData(
     const emptyResponseErr = RegExp(
       /The string did not match the expected pattern|(Unexpected end of (JSON input|data at line 1 column 1))/i
     );
-    /* istanbul ignore else */
+    /* istanbul ignore else -- @preserve */
```
dixonyant: ~~The ignores leave a documentation path for cases where we are losing browser coverage.~~

packages/arcgis-rest-portal/test/services/get-unique-service-name.test.ts
```diff
@@ -23,21 +23,19 @@ describe("get-unique-service-name:", () => {
...
     fetchMock.mock("*", { available: true }, { method: "GET" });
 
-    return getUniqueServiceName(
+    const result = await getUniqueServiceName(
       "myService",
       "Feature Service",
       MOCK_USER_SESSION,
       0
-    ).then((result) => {
-      const call = fetchMock.lastCall();
```
dixonyant, line 35: unused variable

packages/arcgis-rest-portal/test/util/scrub-control-chars.test.ts
```diff
@@ -17,9 +18,6 @@ describe("scrubControlChars", function () {
     // ISO 8859 special char
     hasAllControlChars += String.fromCharCode(160);
 
-    expect(scrubControlChars(hasAllControlChars)).toBe(
-      "foobarbaz",
-      "removes all control chars"
```
dixonyant, line 22: removing jasmine secondary success message as it does not exist in vitest.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -388,6 +388,7 @@ export function internalRequest(
         // encode the parameters into the query string
         const queryParams = encodeQueryString(params);
         // dont append a '?' unless parameters are actually present
+        /* istanbul ignore next */
```
dixonyant, line 391: When the portal tests are taken out of the karma environment, the coverage is lost for the true case. I found it a bit hard to isolate, as every GET request i tried would always append "f=json" as a param to the url and I couldn't find a way to trigger the true path. I think there is a test somewhere in the other packages or portal that is covering this case implicitly, but the time to do discovery is already longer then the time it'll take to just finish this and migrate rest-request under the vitest side and reverify then. my solution is temporarily add ignore until we finish converting rest-request.

packages/arcgis-rest-request/test/utils/append-custom-params.test.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * This test file is being added (maybe temporarily?) because the appendCustomParams function
+ * is getting coverage indirectly by tests of methods in other packages that use it,
+ * but doesnt have any unit tests on its own in request.
+ */
+
```
dixonyant, line 6: appendCustomParams doesn't have a unit test in request, but was being covered implicitly by tests in other packages (portal) similar to the url issue. I added this test to fix the coverage in the request. It might be redundant, but i'll revisit it after rest-request is finished converting. I think it's good to have this test explicitly tested here though.

scripts/test-helpers.ts
```diff
@@ -22,6 +22,7 @@ export const YESTERDAY = (function () {
   return now;
 })();
 
+/* istanbul ignore next */
```
dixonyant: ~~Moving coverage out of karma causes this test coverage to begin failing similar to the other istanbul ignore issues. It was being tested indirectly by tests in portal. My test prototypes were basically just mocking the constructor. But we might want to mock a browser for coverage, with vitest browser mocks.~~

packages/arcgis-rest-portal/test/items/get.test.ts
```diff
@@ -49,71 +50,49 @@ describe("get", () => {
...
+    const [url, options] = fetchMock.lastCall("*");
+    expect(url).toEqual(
+      "https://www.arcgis.com/sharing/rest/content/items/3ef/data?f=json"
+    );
+    expect(options.method).toBe("GET");
   });
 
-  it("should return binary item data by id", (done) => {
+  test("should return binary item data by id", async () => {
     // Blob() is only available in the browser
```
dixonyant: ~~I think this test may be (one of) contributing to the loss of coverage on the getItemData branches.~~

packages/arcgis-rest-portal/src/items/get.ts
patrickarlt: @dixonyant I think that is makes the most sense to make these work in Node now. If this proves too complicated of time consuming we could continue to ignore this until v5 when we stop using `node-fetch` and rely solely on built in fetch.

scripts/test-helpers.ts
dixonyant: resolved this issue by just moving test-helpers.ts to the vitest coverage and removing it from karma. I made a test-helpers.karma.ts because karma still needs to provision the helper functions for request tests, so to separate the cross-dependency issue that was making karma give errors, I gave each environment its own helper file temporarily. It should all be resolved once request is finished migrating.

packages/arcgis-rest-portal/src/items/get.ts
dixonyant: removed ignore altogether as test is now running in node

packages/arcgis-rest-portal/src/items/get.ts
dixonyant: reverted ignore as this test has been changed to run in node, by using the Blob type exposed in the test helpers ponyfill used in other tests for File and FormData. When we stop using node-fetch then we can readdress them altogether.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -388,6 +388,7 @@ export function internalRequest(
...
         const urlWithQueryString =
           queryParams === "" ? url : url + "?" + encodeQueryString(params);
```
dixonyant: this encodeQueryString seems to be happening twice, should we just append ...queryParams below instead of making the call again?
```suggestion
          queryParams === "" ? url : `${url}?${queryParams}`;
```
```suggestion
          queryParams === "" ? url : url + "?" + queryParams;
```

scripts/test-helpers.karma.ts
```diff
@@ -0,0 +1,32 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
```
dixonyant, line 1: temporary test helpers for karma without the fileAttachment() method to preserve 1:1 coverage in the respective environments until rest-request is finished migrating to vitest.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -390,7 +390,7 @@ export function internalRequest(
         // dont append a '?' unless parameters are actually present
         /* istanbul ignore next */
         const urlWithQueryString =
-          queryParams === "" ? url : url + "?" + encodeQueryString(params);
+          queryParams === "" ? url : `${url}?${queryParams}`;
```
dixonyant, line 393: encodeQueryString(params) was being called twice instead of just appending the encoded string.

patrickarlt:
@tomwayson @dbouwman can you also help review this since the portal package was mostly written by the Hub team?

packages/arcgis-rest-request/src/request.ts
tomwayson, line 393: I think you just want to ignore the true case then, like this?

```suggestion
        const urlWithQueryString =
queryParams === "" ? /* istanbul ignore next */ url : url + "?" + encodeQueryString(params);
```

vitest.config.js
```diff
@@ -25,6 +25,7 @@ export default defineConfig({
         "packages/arcgis-rest-feature-service/src/**/*.{ts,js}",
         "packages/arcgis-rest-developer-credentials/src/**/*.{ts,js}",
         "packages/arcgis-rest-portal/src/**/*.{ts,js}",
+        "scripts/test-helpers.ts"
```
tomwayson, line 28: shoudl these helpers be included in code coverage?

dixonyant: nice catch, karma seemed to combine provisioning and coverage with it but I realized it doesn't need to be the same in vitest, so i'm removing it from coverage in the request migration changes.

packages/arcgis-rest-request/src/request.ts
dixonyant: Yes. although once request changes are finished, the true case should be covered again, in which case i'll be able to remove the ignore altogether. I noted this for revisit once request is finished.

PR #1270, opened 2025-10-31
Request test migration

dixonyant:


packages/arcgis-rest-request/test/utils/encode-form-data.test.ts
```diff
@@ -1,26 +1,30 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
+import { describe, test, expect } from "vitest";
 import { encodeFormData } from "../../src/utils/encode-form-data.js";
 import {
   requiresFormData,
   processParams
 } from "../../src/utils/process-params.js";
-import { createReadStream } from "fs";
```
dixonyant, line 9: Had to move this to dynamic-environment-load since vitest gets mad trying to pre-initialize 'fs' when building the browser environment.

packages/arcgis-rest-request/test/revoke-token.test.ts
```diff
@@ -1,55 +1,54 @@
-import { revokeToken, appendCustomParams } from "../src/index.js";
```
dixonyant, line 1: unused import

package.json
```diff
@@ -118,7 +115,7 @@
     "typescript": "^4.0.0",
     "ultra-runner": "^3.10.5",
     "vite": "^7.0.5",
-    "vitest": "^3.2.4"
+    "vitest": "^4.0.5"
```
dixonyant, line 119: 😁

package.json
```diff
@@ -18,19 +18,23 @@
...
-    "test:firefox": "karma start --single-run --browsers=Firefox",
-    "test:firefox:ci": "karma start --single-run --browsers=FirefoxHeadless",
-    "test:node": "node --loader ts-node/esm node_modules/jasmine/bin/jasmine.js --config=jasmine.json",
-    "test:node:debug": "node --inspect --inspect-brk --loader  ts-node/esm node_modules/.bin/jasmine --config=jasmine.json",
-    "test:node:live": "node --loader ts-node/esm node_modules/.bin/jasmine --config=jasmine.live.json",
-    "test:node:live:debug": "node  --inspect --inspect-brk --loader ts-node/esm node_modules/.bin/jasmine --config=jasmine.live.json",
-    "test:vitest": "vitest --config=vitest.config.js --run",
+    "test:all": "vitest --config=vitest.config.js --run",
+    "test:browsers": "vitest --config=vitest.config.js --run --project=browser",
+    "test:ci": "vitest --config=vitest.config.js --run --project=node --project=headless",
```
dixonyant, line 25: If we just want to test chrome in ci, we can add --browser=chrome to just run chrome headless in the build  for quicker runtimes. We could also optimize which playwright browsers we install during the ci build as well.

.github/workflows/pull-request-tests.yml
```diff
@@ -29,20 +29,17 @@ jobs:
...
-      - name: Test in Node
-        run: npm run test:node
-
-      - name: Test in Chrome
-        run: npm run test:chrome:ci
-
-      - name: Test with Vitest
-        run: npm run test:vitest
+      - name: Test in CI
+        run: npm run test:ci
```
dixonyant, line 42: test:ci runs tests in node, and headless chrome/firefox/webkit(safari) full suite in each. 

If we want to just run node and chrome that can be done to speed up our build times.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -31,6 +31,7 @@ import { IOAuthState } from "./types/oauthState.js";
 function isCredential(credential: any): credential is ICredential {
   return (
     typeof credential.userId === "string" ||
+    /* istanbul ignore next -- @preserve */
```
dixonyant, line 34: @tomwayson Karma was covering this code but switching to vitest lost coverage. I did some discovery and found https://github.com/Esri/arcgis-rest-js/commit/ae8159b7ad428a4fd3cf4ca665829a9e8e95ddb2 but didn't see explicit coverage added, and the current tests for it were a bit complex to understand on a first pass. Advice?

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -944,7 +945,8 @@ export class ArcGISIdentityManager
             hasServer: false,
             server: credential.server
           } as IServerInfo)
-        : new ArcGISIdentityManager(credential);
+        : /* istanbul ignore next -- @preserve */
+          new ArcGISIdentityManager(credential);
```
dixonyant, line 949: Pt.2 of above comment

packages/arcgis-rest-request/test/ArcGISIdentityManager.test.ts
```diff
@@ -1051,888 +913,801 @@ describe("ArcGISIdentityManager", () => {
...
-          crypto: {
-            getRandomValues: function () {
-              return new Uint8Array(32);
+  describe(".refreshCredentials() browser tests", () => {
+    /**
+     * `beginOAuth2()` and `completeOAuth2()` only work on browsers due to their use of `dispatchEvent()` and `addEventListener()`.
+     *
+     * We COULD make these tests work in Node if we wanted to mock these calls but you actually have to trigger the events.
+     */
+    test("should be a test", () => {
```
dixonyant: vitest gets mad in node when there's no test in this describe block. gonna see if there's a tag to exclude in node env.

dixonyant (commented):
My review and comments

packages/arcgis-rest-request/test/ArcGISIdentityManager.test.ts
dixonyant, line 1: this diff is royally bugged :\

.github/workflows/pre-release-tests.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [18, 20, 22]
```
dixonyant, line 21: Vitest v4 does not like node 18 and we lose access to a crypto.hash function https://medium.com/@bensedrinedonia/how-i-fixed-the-crypto-hash-is-not-a-function-error-in-vite-on-windows-node-js-18-bf494185f952 idk our exact reasoning for supporting the 3 node versions so don't know if we want to implement a fix or just upgrade.

.github/workflows/pre-release-tests.yml
```diff
@@ -29,17 +29,14 @@ jobs:
...
-      - name: Test in Node
-        run: npm run test:node
-
-      - name: Test in Chrome
-        run: npm run test:chrome:ci
-
-      - name: Test with Vitest
-        run: npm run test:vitest
+      - name: Test in CI
+        run: npm run test:ci
```
dixonyant, line 42: test:ci runs tests in node, and headless chrome/firefox/webkit(safari) full suite in each. build time is 5 mins now with this so i wonder if we just run a node test suite and then headless chrome like we did in karma and then have access to the others via the package json. I could also customize the playwright installation to only install a chrome browser to additionally decrease the ci build size and time.

.github/workflows/pre-release-tests.yml
patrickarlt: This list should generally just be whatever versions of Node are:

* Currently in LTS
* Supported by Github actions

So fine with this. Ideally we should update to include 24 and 26 soon and drop 20 but that is outside the scope of this PR.

.github/workflows/pull-request-tests.yml
patrickarlt: I'm fine with this since those are the main browser engines at this point we can wait a little longer for CI.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: These functions are only supposed to be used internally by trusted apps so I think we are ok to ignore these for now. @tomwayson what do yo think?

package.json
patrickarlt: Overall this still seems fast enough and I would rather be through since you already took the time to set this up.

vitest.config.js
```diff
@@ -25,11 +25,12 @@ export default defineConfig({
         "packages/arcgis-rest-feature-service/src/**/*.{ts,js}",
         "packages/arcgis-rest-developer-credentials/src/**/*.{ts,js}",
         "packages/arcgis-rest-portal/src/**/*.{ts,js}",
-        "scripts/test-helpers.ts"
+        "packages/arcgis-rest-request/src/**/*.{ts,js}"
```
patrickarlt: This list could also be cleaned up a bit similar to by below comment.

patrickarlt (commented):
@dixonyant Congrats! I think this looks great. I left a few comment throughout but i think this is otherwise good to do. I would like @dbouwman or @tomwayson to look over the parent auth and no cors tests since those are their areas of expertise and the tests are especially complicated.

* I did note that we could simplify the Vitest project config now that this work is done.
* I would remove the .env.example file and the one live test file. I think that we can revisit the live tests later.

dbouwman:
@dixonyant I'll block off some time today to look at the postMessage and no-cors stuff.

vitest.config.js
dixonyant: I didn't see a below comment, idk if github disappeared it, but I think I'm tracking, I'll move these to a variable and include that instead.

packages/arcgis-rest-request/test/utils/generate-code-challenge.test.ts
```diff
@@ -1,23 +1,31 @@
+import { describe, test, expect } from "vitest";
 import { generateCodeChallenge } from "../../src/utils/generate-code-challenge.js";
-import { isBrowser } from "../../../../scripts/test-helpers.karma.js";
+import { isBrowser, isNode } from "../../../../scripts/test-helpers.js";
 
 describe("generateCodeChallenge()", () => {
+  if (isNode) {
```
dixonyant, line 6: I could do a todo() or a skip() test here to satisfy vitest's no empty describe rule for node, but then it just shows skipped tests warning in the console. Trivial, but liked all green solution better :)

vitest.config.js
dixonyant: Was able to get the same result by just including all the sources in packages and same with coverage.

package.json
patrickarlt: 🎉

.github/workflows/pre-release-tests.yml
patrickarlt: 👍🏻

vitest.config.js
patrickarlt: @dixonyant this is what comment was! 👍🏻

patrickarlt:
@dixonyant I think this is all ready to merge and we can close out the migration epic. Great work on this!

Issue #821, opened 2021-03-31
Update typings for functions that wrap `request()` to return `Response` when using `rawResponse`


thw0rted:
As I commented on #462, the current typings for e.g. `queryFeatures` say that the method returns `Promise<IQueryFeaturesResponse | IQueryResponse>`.  The method actually returns a `Promise<Response>` when the `rawResponse` option is true.

It would be nice if the method could return the correct promise type directly based on the shape of the options object passed, either via implied generics or using a (granted, pretty verbose) overload.  Overloading would be verbose but I don't think especially difficult -- see [this Angular class](https://github.com/angular/angular/blob/master/packages/common/http/src/client.ts) for an example.  Much like this library, the Angular HTTPClient has a `request` method that changes its return type based on an options-object, and multiple helpers that call `request` and also vary the return type based on arguments.

patrickarlt:
> It would be nice if the method could return the correct promise type directly based on the shape of the options object passed, either via implied generics or using a (granted, pretty verbose) overload

[...]

PR #1272, opened 2025-11-24
Update dependencies

patrickarlt:
Now that we are free from Karma and Jasmine (yay @dixonyant!) I've updated the dependencies for TypeScript, TypeDoc, ESLint and Rollup. This fixes problems with peer dependencies and means we can go back to normal `npm install` instead of `npm install --legacy-peer-deps`.

this also makes a bunch of other changes including:

* Resolves a bunch of random linting errors that came up as a result of updating the linting libraries and TypeDoc.
* Add and configure changesets to manage releases.

### DIFF
--- a/packages/arcgis-rest-routing/src/serviceArea.ts
+++ b/packages/arcgis-rest-routing/src/serviceArea.ts
@@ -77,7 +77,7 @@ function getTravelDirection(
  *   .then(response) // => {routes: {features: [{attributes: { ... }, geometry:{ ... }}]}
  * ```
  *
- *  See the [REST API documnetation](https://developers.arcgis.com/rest/network/api-reference/service-area-synchronous-service.htm) for more information.
+ *  See the [REST API documentation](https://developers.arcgis.com/rest/network/api-reference/service-area-synchronous-service.htm) for more information.
  *
  * @param requestOptions Options to pass through to the routing service.
  * @returns A Promise that will resolve with service area polygons for the request.


patrickarlt:
Making this a draft while I figure out TypeDoc issues.

packages/arcgis-rest-routing/src/helpers.ts
```diff
@@ -10,6 +10,10 @@ import {
   IFeatureSet
 } from "@esri/arcgis-rest-request";
 
+export interface IFeatureSetWithGeoJson extends IFeatureSet {
```
dixonyant, line 13: I'm interested as to what this is doing as I may have run into an overlapping type issue with the query library response that also extends this IFeatureSet, returning both geojson and arcgis under the features prop

.changeset/config.json
```diff
@@ -0,0 +1,14 @@
...
+  "$schema": "https://unpkg.com/@changesets/config@3.1.2/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "fixed": [["@esri/arcgis-rest-*"]],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
```
dixonyant, line 11: glory

packages/arcgis-rest-developer-credentials/src/shared/types/apiKeyType.ts
```diff
@@ -52,7 +52,7 @@ export interface ICreateApiKeyOptions
  */
 export interface IGetApiKeyOptions extends Omit<IRequestOptions, "params"> {
   /**
-   * {@linkcode IAuthenticationManager} authentication.
+   * {@linkcode @esri/arcgis-rest-request!IAuthenticationManager} authentication.
```
dixonyant, line 55: wow was this all it needed? :)

.github/workflows/release.yml
```diff
@@ -59,25 +61,28 @@ jobs:
...
 
       - name: Update NPM
         run: npm install -g npm@latest
 
       - name: Install
-        run: npm ci --legacy-peer-deps
-
-      - name: Build
-        run: npm run build
+        run: npm ci
```
dixonyant, line 74: nice :)

packages/arcgis-rest-routing/src/helpers.ts
patrickarlt: The routing methods work a little different and return a `geoJson` property next to a `features` property. I think what you are looking for is [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) I cooked up something in https://github.com/Esri/arcgis-rest-js/issues/821#issuecomment-924148504 that I think we should look at for v5 since I think that changing the types would be a breaking change.

packages/arcgis-rest-developer-credentials/src/shared/types/apiKeyType.ts
patrickarlt: Yup. It was buried in the doc on [declaration references](https://typedoc.org/documents/Declaration_References.html)

packages/arcgis-rest-routing/src/helpers.ts
dixonyant: That was a fantastic read and nice rabbit hole to jump down. Gives a lot of context

packages/arcgis-rest-routing/src/serviceArea.ts
```diff
@@ -80,9 +77,10 @@ function getTravelDirection(
  *   .then(response) // => {routes: {features: [{attributes: { ... }, geometry:{ ... }}]}
  * ```
  *
+ *  See the [REST API documnetation](https://developers.arcgis.com/rest/network/api-reference/service-area-synchronous-service.htm) for more information.
```
noahmulfinger:
```suggestion
 *  See the [REST API documentation](https://developers.arcgis.com/rest/network/api-reference/service-area-synchronous-service.htm) for more information.
```

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
```diff
@@ -352,7 +355,7 @@ export class ArcGISIdentityManager
    *
    * A {@linkcode ArcGISAccessDeniedError} error will be thrown if the user denies the request on the authorization screen.
    *
-   * @browserOnly
```
noahmulfinger, line 355: 1. So was support dropped for `@browserOnly` and `@nodeOnly` tags?\
2. How will these typedoc changes affect our typedoc processing on the developer website?

packages/arcgis-rest-request/src/index.ts
```diff
@@ -46,6 +46,7 @@ export * from "./types/service.js";
 export * from "./types/group.js";
 export * from "./types/user.js";
 export * from "./types/job-statuses.js";
+export * from "./types/oauthState.js";
```
noahmulfinger, line 49: This is just an internal detail and a larger discussion for follow-up, but I see some file names using kebab case and others using camel case. Do we want to settle on a standard for this repo?

package.json
```diff
@@ -60,58 +58,46 @@
   },
   "husky": {
     "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "npm run precommit"
+      "pre-commit": "npm run precommit && npm run typedoc"
```
noahmulfinger: Just wondering if running `typedoc` should be a PR check instead of a commit check. For example when I ran it locally without building, it threw a bunch of errors due to unbuilt packages. If someone is just committing minor chore commits it could be overkill.

RELEASE.md
```diff
@@ -1,10 +1,14 @@
 # Release system
 
-## v4
+## December 2025/January 2026 Updates
 
-As of v4, the ArcGIS REST JS packages are released automatically via Semantic Release.
+In Late 2025 ArcGIS REST JS now uses [changesets](https://github.com/changesets/changesets) to manage releases. This happens automatically using the default [changeset action](https://github.com/changesets/action) in the [release workflow](./.github/workflows/release.yml). When changes are merge to `main` changesets will automatically creates a PR that when merged will trigges a new release of all packages.
```
noahmulfinger:
```suggestion
In late 2025 ArcGIS REST JS now uses [changesets](https://github.com/changesets/changesets) to manage releases. This happens automatically using the default [changeset action](https://github.com/changesets/action) in the [release workflow](./.github/workflows/release.yml). When changes are merged to `main`, changesets will automatically creates a PR that, when merged, will trigger a new release of all packages.
```

RELEASE.md
```diff
@@ -1,10 +1,14 @@
...
+
+As a part of these changes all versions of ArcGIS REST JS packages are how synced to the highest minor version accross all packages. All packages will now maintain the same version number for ease of reference.
```
noahmulfinger:
```suggestion
As a part of these changes, all versions of ArcGIS REST JS packages are now synced to the highest minor version across all packages. All packages will now maintain the same version number for ease of reference.
```

noahmulfinger: Is it the case that "all packages will now maintain the same version number"? Isn't that dependent on the changeset keyword used (`patch` or `minor`)?

noahmulfinger (changes requested):
Typos and a few questions

RELEASE.md
patrickarlt: Yes but they will all be bumped to the same version (patch/minor/major) together. Keeping everything on the same version will make things like the support lifecycles easier

packages/arcgis-rest-request/src/index.ts
patrickarlt: We should, probably for v5.

packages/arcgis-rest-request/src/ArcGISIdentityManager.ts
patrickarlt: We never actually supported these on the developer website. We just moved this info into the comment.

package.json
patrickarlt: Good idea.

noahmulfinger (approved):

PR #1273, opened 2025-11-24
Add trusted publishing configuration

patrickarlt:
This PR sets up trusted publishing for `@esri/arcgis-rest-*` packages. Long lived personal tokens are on the way out after continued supply chain attacks.

This PR switches to use NPM Trusted Publishing according to this documentation:

* https://docs.npmjs.com/trusted-publishers
* https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/
* https://docs.github.com/en/actions/concepts/security/openid-connect

Then we switch to using the GitHub actions provided `GITHUB_TOKEN` instead of the token in the repo secrets for the Git operations that make up the rest of the publishing process according to these docs:

* https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions#github-workflows-release.yml-configuration-for-node-projects
* https://docs.github.com/en/actions/tutorials/authenticate-with-github_token#permissions-for-the-github_token
* https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

This PR will also need to updated to remove the `--legacy-peer-deps` flags if https://github.com/Esri/arcgis-rest-js/pull/1272 merges first.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,7 @@ jobs:
           node-version: 24
 
       - name: Update NPM
-        run: npm install -g npm@11.5.1
+        run: npm install -g npm@latest
 
       - name: Install
         run: npm ci --legacy-peer-deps


.github/workflows/release.yml
```diff
@@ -64,14 +66,10 @@ jobs:
...
+          node-version: 24
 
-      - name: Configure npm
-        run: |
-          echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > .npmrc
-          cat .npmrc
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Update NPM
+        run: npm install -g npm@11.5.1
```
gavinr-maps: Is this intentionally v11.5.1 or should it be `latest`? I see `latest` in [the doc](https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow).

patrickarlt: I was initially uncertain about just automatically adopting latest but I don't think it should be that much of the problem in the long run so I've updated this.

.github/workflows/release.yml
```diff
@@ -24,12 +24,10 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
-    # PRs will run tests on node 14,16 on ubuntu, macos and windows
-    # so for the release, we're just running node 16@ubuntu
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node: [18]
+        node: [24]
```
dixonyant, line 30: will we need to retroactively go through the other yml files and update them do remove legacy-peer-deps and update the node versions to 24 as well?

dixonyant (approved):
LGTM! I left only one comment.

.github/workflows/release.yml
patrickarlt: @dixonyant yes eventually we want to only be on LTS versions of Node in these tests but I wanted to make sure we were up to date here since have to use NPM 11.5.1

patrickarlt:
@dixonyant @gavinr-maps going to merge this and see if we get a release of the request package!

PR #1274, opened 2025-11-24
Fix release workflow tests

patrickarlt:
When I setup trusted publishing I neglected to add the step to install the Playwright browsers so this PR adds them into the release workflow.

This also makes a few other changes:

* Remove the `pre-release-tests.yml` file since it just runs the tests and does nothing. We run the tests in the `release.yml` file.
* Run the release tests on Node 20, 22 and 24 which are our supported Node versions but only run the release on Node 24.


dixonyant (approved):
Nice! Super clean. :)

Issue #1271, opened 2025-11-10
Add support for PBF queries to `queryFeatures()` and `queryAllFeatures()`


patrickarlt:
In order to support https://github.com/Esri/maplibre-arcgis/issues/19 we need to add support for `f=pbf` to both [`queryFeatures()`](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-feature-service/queryFeatures/) and [`queryAllFeatures()`](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-feature-service/queryAllFeatures/). Rowan had kindly supported us porting his work in [`arcgis-pbf-parser`](https://github.com/rowanwins/arcgis-pbf-parser/) into REST JS to get started.

In order to support both GeoJSON and Esri formatted geometries I think that we should implement 3 new values for the `f` param

* `f=pbf` = Returns the compressed PBF so a user can decide how they want to handle it. We should also expose 2 utility functions `pbfToGeoJSON()` and `pbfToArcGIS()` which can also be used internally.
* `f=pbf-as-geojson` - Returns the decompressed PBF as GeoJSON. This is the same behavior as [`arcgis-pbf-parser`](https://github.com/rowanwins/arcgis-pbf-parser/).
* `f=pbf-as-arcgis` - Returns the decompressed PBF as [ArcGIS geometry objects](https://developers.arcgis.com/rest/services-reference/enterprise/geometry-objects/) this can be done via [`@terraformer/arcgis`](https://github.com/terraformer-js/terraformer/tree/main/packages/arcgis)

For the `queryAllFeatures()` function we would only support `f=pbf-as-geojson` and `f=pbf-as-arcgis` since we need to decompress the PBF anyway to understand if we need to load more features.

PR #1283, opened 2025-12-31
Query features with pbf-as-geojson and pbf-as-arcgis format parameters

dixonyant:
### Description:

We need a way to query pbf as geojson or arcgis json objects to speed up load times for larger data sets, such as on-demand feature loading in MapLibre. 

Adding support for f=pbf-as-geojson and f=pbf-as-arcgis to [queryFeatures()](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-feature-service/queryFeatures/) and [queryAllFeatures()](https://developers.arcgis.com/arcgis-rest-js/api-reference/arcgis-rest-feature-service/queryAllFeatures/) for issue #1271.

This PR adds two format query params to queryFeatures and queryAllFeatures: 
```
	const options: IQueryFeaturesOptions | IQueryAllFeaturesOptions = {
		...params,
		f: "pbf-as-geojson" | "pbf-as-arcgis"
	}
```

#### Additions:
- Generated new PBFFeatureCollection decoder file generated from the latest Esri proto spec.
- Added two parsers adapted from Rowan's work in [arcgis-pbf-parser](https://github.com/rowanwins/arcgis-pbf-parser).
	- exported two functions pbfToGeoJson() and pbfToArcGIS()
- Added new query method to parse pbf raw responses and return geojson or arcgis json.
- Tests: added live tests, query and parsers tests, and test helpers for managing geometry precision variance.

**Known/future Issues:** (pbf-as-arcgis)
- Doesn't support Multipoint
- hasn't been tested with 3D or HasZ Point objects
- Doesn't support length property on IFields in pbf-to arcgis (see below)
- esriFieldTypeString (sqlTypeNVarchar) decoded fields do not return a length property based on the current available proto spec.
- More could be done with to pbfToArcGIS to add other geometry types to it and better error management if we choose to do a full suite support for REST JS for pbf-to-arcgis and upgrade to a newly generated proto definition that supports all new fields (like sqlType and length). But the use case for pbf-as-arcgis that is more than read-only seems to not exist for now.

#### Testing distribution:

**LIVE Tests:** test parser and query features against live endpoints
- test query responses
- test errors
- test decoded object shape & equality
- test geometry types
- test geometry precisions

**Local Tests:**
- Query Tests
	- Test query responses through fetch-mock 
	- Test error states and status codes
- Parser Tests
	- Test decoded object shape & equality
	- Test geometry types
	- Test geometry precisions

### DIFF
--- a/.changeset/cute-bikes-check.md
+++ b/.changeset/cute-bikes-check.md
@@ -1,0 +1,9 @@
+---
+"@esri/arcgis-rest-feature-service": minor
+---
+
+Added query params where user can query a pbf request and get back geojson or arcgis features.
+- User can request features from feature service using "f: pbf-as-geojson" or "f: pbf-as-arcgis" in request options
+- Added EsriFeatureCollection proto spec and modular pbf decoder
+- Added pbfToGeoJSON parser and pbfToArcGIS parser
+- Added Live Tests 
--- a/.changeset/little-places-tell.md
+++ b/.changeset/little-places-tell.md
@@ -1,0 +1,6 @@
+---
+"@esri/arcgis-rest-feature-service": minor
+"@esri/arcgis-rest-request": patch
+---
+
+Added pbf types to request RequestFormats types.


packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -150,6 +156,78 @@ export interface IQueryResponse {
...
+          err.message,
+          err.code,
+          err,
+          url,
+          customOptions
+        );
+      }
+      try {
+        const arrayBuffer = await response.arrayBuffer();
+        /* istanbul ignore else --@preserve */
```
dixonyant, line 224: istanbul coverage drops expecting an implicit "else" branch.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
+/* istanbul ignore file --@preserve */
```
dixonyant: A few thoughts for this class:
2. (How) do we want to document/accredit the work from rowan as we are importing it here? 
3. For testing, do we want to just black-box test this with pbf i/o or flesh out the tests fully per method in this class?
4. there was one unimplemented data type (MultiPoint) that rowan said wasn't covered in this pbf-parser https://github.com/rowanwins/arcgis-pbf-parser?tab=readme-ov-file, do we need to implement that with this story?

packages/arcgis-rest-feature-service/src/pbf-parser/PbfFeatureCollection.ts
```diff
@@ -0,0 +1,896 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+/* istanbul ignore file --@preserve */
```
dixonyant: This is a pbf generated class. don't know how we would want to proceed with tests for it.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -282,11 +367,10 @@ export async function queryAllFeatures(
   let offset = 0;
   let hasMore = true;
   let allFeaturesResponse: IQueryAllFeaturesResponse | null = null;
-
   // retrieve the maxRecordCount for the service
+
   const pageSizeResponse = await request(requestOptions.url, {
-    httpMethod: "GET",
-    authentication: requestOptions.authentication
```
dixonyant, line 289: authentication wasn't necessarily needed on feature-service public endpoints, but was being forced in rest-js. removed to match queryFeatures() functionality.

packages/arcgis-rest-feature-service/test/query.test.ts
```diff
@@ -134,6 +140,341 @@ describe("getFeature() and queryFeatures()", () => {
...
+      try {
+        await queryFeatures(docsPbfOptions);
+      } catch (error) {
+        expect(error).toBeInstanceOf(ArcGISAuthError);
+        expect((error as any).message).toBe("498: Invalid token.");
+      }
+    });
+
+    test("(invalid auth) should throw arcgis auth error when service returns 200 with json object containing token required error", async () => {
+      const featureServiceInvalidTokenErrorResponse = {
```
dixonyant: not sure how to force 499 or 500 error states but wanted to handle them as they are covered in request for geojson/arcgis formats, but are bypassed when requesting with raw response, and need to be handled here.

vitest.projects.ts
```diff
@@ -1,11 +1,20 @@
 import { playwright } from "@vitest/browser-playwright";
 
-export const allTestPackages = ["packages/**/*.{test,spec}.?(c|m)[jt]s?(x)"];
+export const testPackages = ["packages/**/*.{test,spec}.?(c|m)[jt]s?(x)"];
+
+export const liveTestPackages = [
+  "packages/**/*.{test,spec}.live.?(c|m)[jt]s?(x)"
+];
+
+export const excludePackages = [
```
dixonyant, line 9: vitest may respect negation patterns but istanbul stops generating reports. Using exclude works.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
patrickarlt: @dixonyant I would add an acknowledgement that the code is adapted from https://github.com/rowanwins/arcgis-pbf-parser?tab=readme-ov-file and add Rowan to the contributors in the `package.json`

patrickarlt: I've also never actually seen a multipoint service either so I think we are fine.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+import { FeatureCollectionPBuffer as EsriPbfBuffer } from "./PbfFeatureCollection.js";
+import Pbf from "pbf";
+
+export default function decode(featureCollectionBuffer: any) {
+  let decodedObject;
+  try {
+    decodedObject = EsriPbfBuffer.read(new Pbf(featureCollectionBuffer));
+  } catch (error) {
+    throw new Error("Could not parse arcgis-pbf buffer: ");
```
patrickarlt: Would be good to display the error here.

```suggestion
    throw new Error(`Could not parse arcgis-pbf buffer: ${error}`);
```

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+  if (!decodedObject.queryResult) {
+    throw new Error("Could not parse arcgis-pbf buffer: ");
```
patrickarlt:
```suggestion
    throw new Error("Could not parse arcgis-pbf format. Missing queryResult key. ");
```

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+  };
+
+  const geometryParser = getGeometryParser(geometryType);
+
+  const featureLen = featureResult.features.length;
+  for (let index = 0; index < featureLen; index++) {
+    const f = featureResult.features[index];
+    out.features.push({
+      type: "Feature",
+      id: getFeatureId(fields, f.attributes, objectIdField),
```
patrickarlt: @dixonyant currently `getFeatureId()` iterates over all the fields again for every feature to find the matching `objectIdField`. This could get expensive when things have lots of fields (20+ fields is not uncommon) and lots of features (full pages) so I think you could avoid this with something like:
```suggestion
  for (let index = 0; index < featureLen; index++) {
    const f = featureResult.features[index];
    out.features.push({
      type: "Feature",
      //deliberatly don't get the id field here
      properties: collectAttributes(fields, f.attributes),
      geometry: f.geometry && geometryParser(f, transform)
    });
  }
  
  // loop over features again and just assign
  feature.id = feature.properties[objectIdField]
```

packages/arcgis-rest-feature-service/src/pbf-parser/PbfFeatureCollection.ts
patrickarlt: I'm fine ignoring this since it is generated and is kind of covered by the other tests.

packages/arcgis-rest-feature-service/src/query.ts
patrickarlt: @dixonyant This still needs to get passed if it is provided in otherwise auth won't get provided to services that require it and will error. What exactly were you seeing as the issue here. `authentication` should be allowed to be `undefined`.

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.live.ts
patrickarlt, line 1: @dixonyant Live tests are| great include here but as we discussed I would make a test case for points, lines and polygons.

packages/arcgis-rest-feature-service/test/query.test.live.ts
```diff
@@ -0,0 +1,364 @@
...
+    fetchMock.restore();
+  });
+
+  describe("queryFeatures()", () => {
+    describe("with pbf-as-geojson", () => {
+      // LIVE TEST: should decode a valid pbf-as-geojson response from public server without api key without fetchmock
+      test("LIVE TEST (valid): should query pbf-as-geojson features from live server and decode into geojson from arrayBuffer", async () => {
+        const testPublicFeatureServer: IQueryFeaturesOptions = {
+          url: "https://services.arcgis.com/P3ePLMYs2RVChkJx/arcgis/rest/services/ACS_Marital_Status_Boundaries/FeatureServer/2",
+          f: "pbf-as-geojson",
```
patrickarlt, line 36: @dixonyant You could also do something similar for these tests to what you did in the PBF to ArcGIS tests here and request an `f=geojson` to compare.he PBF output against the GeoJSON output.

packages/arcgis-rest-feature-service/test/query.test.ts
patrickarlt: This is a fine way to just mock it. I'm not 100% sure off the top of my head how to trigger one. Not sure if @gavinr might know?

packages/arcgis-rest-feature-service/test/query.test.ts
```diff
@@ -183,7 +523,7 @@ describe("queryAllFeatures", () => {
   });
 
   test("fetches multiple pages based on feature count with authentication", async () => {
-    fetchMock.mock(`${serviceUrl}?f=json&token=MOCK_TOKEN`, {
+    fetchMock.mock(`${serviceUrl}?f=json`, {
```
patrickarlt: @dixonyant related to my comment above this still needs to pass as-is.

packages/arcgis-rest-feature-service/test/query.test.ts
```diff
@@ -348,4 +687,398 @@ describe("queryAllFeatures", () => {
...
+        );
+        arrayBufferSet1 = await pbf.arrayBuffer();
+      }
+
+      if (isNode) {
+        const fs = await import("fs");
+        const filePath =
+          "./packages/arcgis-rest-feature-service/test/mocks/PbfResultsSet6Partial.pbf";
+        arrayBufferSet1 = fs.readFileSync(filePath);
+      }
```
patrickarlt: This logic gets repeated a bunch for the different file types. Could you simplify it into a helper method?

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+  const objectIdField = featureResult.objectIdFieldName;
+
+  // Wires up the field keynames
+  const fields = featureResult.fields;
+  for (let index = 0; index < fields.length; index++) {
+    const field = fields[index];
+    field.keyName = getKeyName(field);
+  }
+
+  const out = {
```
gowin20: Do we want to use the `GeoJSON.FeatureCollection` type for this method? https://www.npmjs.com/package/@types/geojson

Right now I'm pretty sure `queryFeatures` returns GeoJSON on certain inputs, but the return type isn't marked as such.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+  };
+}
+
+// * @property {number} esriGeometryTypePoint=0 esriGeometryTypePoint value
+// * @property {number} esriGeometryTypeMultipoint=1 esriGeometryTypeMultipoint value
+// * @property {number} esriGeometryTypePolyline=2 esriGeometryTypePolyline value
+// * @property {number} esriGeometryTypePolygon=3 esriGeometryTypePolygon value
+// * @property {number} esriGeometryTypeMultipatch=4 esriGeometryTypeMultipatch value
+// * @property {number} esriGeometryTypeNone=127 esriGeometryTypeNone value
+function getGeometryParser(featureType: any) {
```
gowin20:
```suggestion
function getGeometryParser(featureType: number) {
```

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+    case 2:
+      return createLine;
+    case 0:
+      return createPoint;
+    default:
+      return createPolygon;
+  }
+}
+
+function createPoint(f: any, transform: any) {
```
gowin20, line 99: What are the types of these inputs? `IFeature`?

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+        transform,
+        startPoint,
+        stopPoint
+      );
+
+      // Check if the ring is clockwise, if so it's an outer ring
+      // If it's counter-clockwise its a hole and so push it to the prev outer ring
+      // This is perhaps a bit naive
+      // see https://github.com/terraformer-js/terraformer/blob/master/packages/arcgis/src/geojson.js
+      // for a fuller example of doing this
```
gowin20, line 177: Does this logic only apply to ArcGIS polygons? Is there separate logic for GeoJSON?

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,270 @@
...
+    pt1 = pt2;
+  }
+  return total >= 0;
+}
+
+function createLinearRing(
+  arr: any,
+  transform: any,
+  startPoint: any,
+  stopPoint: any
```
gowin20: I still think we should use explicit types here. I'm not sure if these methods are for ArcGIS feature JSON, GeoJSON, or both.

package.json
```diff
@@ -69,13 +69,15 @@
     "@rollup/plugin-node-resolve": "^16.0.3",
     "@rollup/plugin-terser": "^0.4.4",
     "@rollup/plugin-typescript": "^12.3.0",
+    "@terraformer/arcgis": "2.1.2",
```
gowin20: Should these dependencies be at the top level `package.json`, or within the `arcgis-rest-feature-service` package instead?

packages/arcgis-rest-request/src/utils/ResponseFormats.ts
```diff
@@ -4,6 +4,9 @@
 export type ResponseFormats =
   | "json"
   | "geojson"
+  | "pbf"
+  | "pbf-as-geojson"
+  | "pbf-as-arcgis"
```
gowin20, line 9: Adding these to `ResponseFormats` means these output options are included in `IParams`, and they will be listed as valid options for `request`, `generateToken`, and several other methods which don't support PBF. I think we should only add these options to the methods that support them -- so `queryFeatures` and `queryAllFeatures` iirc.

packages/arcgis-rest-feature-service/src/query.ts
gowin20: Agreed that we definitely need authentication to be an option to support secure data layers

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -282,11 +366,10 @@ export async function queryAllFeatures(
   let offset = 0;
   let hasMore = true;
   let allFeaturesResponse: IQueryAllFeaturesResponse | null = null;
-
   // retrieve the maxRecordCount for the service
+
   const pageSizeResponse = await request(requestOptions.url, {
```
gowin20, line 387: Since we're already requesting the layer definition here, we could perform an additional check to see if the service supports output as PBF or GeoJSON:

```js
const format = requestOptions.params.f
if (!pageSizeResponse.supportedQueryFormats.includes(format)) throw new Error('This feature service does not support output in the requested format.');
```

This would only be possible in `queryAllFeatures` since it requires this initial URL request before the query.

packages/arcgis-rest-feature-service/src/utils/pbfToArcGIS.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const pbfToArcGIS = (arrayBuffer: ArrayBuffer): IPbfToArcGIS => {
+  // convert pbf array buffer to geojson structure
+  const decoded = pbfToGeoJSON(arrayBuffer);
+  const featureCollection: GeoJSON.FeatureCollection = {
+    type: decoded.featureCollection.type,
+    features: decoded.featureCollection.features as GeoJSON.Feature[]
+  };
+  // send geojson structure to terraformer for conversion to arcgis json
+  const arcgis = geojsonToArcGIS(featureCollection);
```
gowin20: Is it OK that we're relying on terraformer for this behavior? I don't think the package is owned by Esri.

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: Adding this back, I totally misinterpreted previous feedback on this line :D I will give it another go and make sure the tests still pass cases with undefined/defined auth.

packages/arcgis-rest-feature-service/test/query.test.live.ts
dixonyant: I believe I did that in the pbfToGeoJSON.test.live.ts file, is that what you were meaning?

packages/arcgis-rest-feature-service/src/utils/pbfToArcGIS.ts
dixonyant: That is a good question. It was the scope of this issue so we could at lest get basic features from both geojson and arcgis for the mapLibre pbf story, but looking at it now, it seems we may have all the information available to just return a full arcgis object with fields and features that fits an expected arcgis structure, otherwise with terraformer we lose a lot of information. My recommendation is that we spin off another story to adapt the geojson parser to an "native" rest js arcgis parser.

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: do we have any services that don't support pbf? we will only ever make requests to the server that are geojson | json | pbf. the pbf-as are rest-js -only options that direct rest js to do the conversions after the request.

packages/arcgis-rest-request/src/utils/ResponseFormats.ts
dixonyant: Yeah....I ran into this and don't know what to think. Typescript gets all upset otherwise unless we refactor to use generics. I feel like we are misdirecting developers to think that that pbf-as- options are compatible with the base requests and other services, even though it's a rest-js and query feature- only thing. Maybe we create another interface? or, I'll give this another pass over. One thing to note is that we are gonna be refactoring IRequestOptions altogether in a near restjs v5 upgrade and some or all of its properties including the way we do params may be changed.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
dixonyant: this is just a geojson parser, although technically the pbf is a encoded arcgis object per my understanding. My aim for this story was to get the parser at least working and such. I suppose I could do more to make this class more robust. Just don't know if that should be in this story or a follow-up story as this is already large.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
dixonyant: I believe this is esriGeometry polygons, so arcgis. then he transforms them into a geojson structure.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
dixonyant: the way I've ported this class, is after conversion, I typecast the output using utility functions to be used by the rest of the rest js service. I think this class should get another pass over to get things typed internally, I was just trying to get it ported over and working in this story.

package.json
dixonyant: I'll try and load it in feature service

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -150,6 +155,77 @@ export interface IQueryResponse {
...
+            features: decoded.featureCollection.features,
+            exceededTransferLimit: decoded.exceededTransferLimit
+          };
+          return geoJsonFeaturesResponse;
+        }
+      } catch (error) {
+        throw new ArcGISRequestError(
+          "Unable to decode pbf response.",
+          500,
+          response,
```
dixonyant, line 238: @patrickarlt I updated all these errors to have the response in the body instead of the thrown error. Is that what you meant with the ArcgisRequestError change? I handle a server general error above, and then internal pbf parsing failure error here as 500.

packages/arcgis-rest-feature-service/src/utils/pbfToArcGIS.ts
dixonyant: I will make a follow-up issue to examine this further and implement straight arcgis conversion in the parse so we don't lose the extra data converting back to arcgis.

packages/arcgis-rest-request/src/utils/ResponseFormats.ts
dixonyant: I followed up on this issue and this is necessary behavior, (same with geojson) until we implement the v5 upgrade and create overrides for request with generics and more.

package.json
dixonyant: done

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
dixonyant: the way queryFeatures is already written that I could see, it doesnt mark features as geojson anyways. (ie. f=json vs f=geojson) has no typed features, queryAllFeatures just handles whatever object types are on the response from request(): Promise<any> . When we implement overrides in v5 this will become pertinent. I will do a pass over and typecast as much as I can, but that info wont go past the query functions until we implement generics.

dixonyant: made the change for this one.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
dixonyant: this f variable isnt exactly an IFeature, it's IFeature(ish). It has attributes and geometry but also compressed_geometry and other geometric variables on the object. Transform isn't documented anywhere that I could find except in the generated pbf buffer reader class that does xyz transformation on data to construct polygons, but its structure may be different for lines and points.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
patrickarlt: @gowin20 https://stevage.github.io/geojson-spec/#section-3.1.6 bullet 4. GeoJSON organizes hole in polygons based on the orientation of the ring so to convert this into ArcGIS JSON you have the check the ring orientation accordingly.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
patrickarlt: I think that a follow up issue would be fine here especially since this is not exported.

packages/arcgis-rest-feature-service/src/query.ts
patrickarlt: I'm pretty sure that you will get back an error from the server anyway if you request an incorrect format. I've haven't seen a service without GeoJSON and PBF output in long time so I think we can reasonably sure that all services support them.

packages/arcgis-rest-feature-service/src/pbf-parser/arcGISPbfParser.ts
```diff
@@ -0,0 +1,403 @@
...
+    type: "Point",
+    coordinates: transformTuple(f.geometry.coords, transform)
+  };
+  const ret = {
+    x: p.coordinates[0],
+    y: p.coordinates[1]
+  };
+  // structure output according to arcgis point geometry spec
+  // istanbul ignore if else --@preserve
+  if (p.coordinates.length > 2) {
```
dixonyant, line 244: Need to know if there is a pbf feature service that returns 3D coordinates that can be used to test this branch.

packages/arcgis-rest-feature-service/src/pbf-parser/arcGISPbfParser.ts
```diff
@@ -0,0 +1,403 @@
...
+    const transformed = transformTuple([x, y], transform);
+    out.push(transformed);
+    prevX = x;
+    prevY = y;
+  }
+  return out;
+}
+
+/* istanbul ignore next --@preserve */
+function transformTuple(coords: any, transform: any) {
```
dixonyant, line 384: Need to find a polygon feature service that returns 3D coordinates to test this method.

packages/arcgis-rest-feature-service/src/pbf-parser/geoJSONPbfParser.ts
```diff
@@ -0,0 +1,309 @@
...
+  const objectIdField = featureResult.objectIdFieldName;
+
+  const out: EsriGeoJSONFeatureCollection = {
+    type: "FeatureCollection",
+    features: []
+  };
+
+  // if the spatial reference is web mercator, add the crs property
+  // otherwise default to assuming WGS84 (ESPG:4326) per geojson
+  // if feature service returns wkid 102100, latestWkid will be 3857
```
dixonyant: consumer may specify pbf outSR to something !(4326 or 3857)?

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    );
+
+    // the current pbf decoder does not define length property on String FieldTypes
+    expect(arcGIS.fields[3].length).toEqual(100);
+    expect(pbfArcGIS.fields[3].length).toBeUndefined();
+
+    // check that fields are equal for both
+    expect(arcGIS.fields.length).toEqual(pbfArcGIS.fields.length);
+    // check that features are equal for both
+    expect(arcGIS.features.length).toEqual(pbfArcGIS.features.length);
```
dixonyant, line 332: need to check for features number values toBeClose to similar to the geoJson tests.

packages/arcgis-rest-feature-service/src/pbf-parser/arcGISPbfParser.ts
```diff
@@ -0,0 +1,402 @@
...
+// Fields: PbfFeatureCollection doesnt decode to IField
+export function decodeField(
+  field: any,
+  fieldTypeMap: Record<number, string>,
+  sqlTypeMap?: Record<number, string>
+): IField {
+  // configure getters that return arcgis json default values for optional props
+  const optionalProps: Array<[string, (f: any) => any]> = [
+    ["alias", (f) => f.alias],
+    // TODO: ? is domain a value that needs to be decoded similar to type, or just parsed as is?
```
dixonyant: get featureservice that has domains

packages/arcgis-rest-feature-service/src/pbf-parser/arcGISPbfParser.ts
```diff
@@ -0,0 +1,404 @@
...
+  switch (featureType) {
+    case 0:
+      return "esriGeometryPoint";
+    case 2:
+      return "esriGeometryPolyline";
+    case 3:
+      return "esriGeometryPolygon";
+    // istanbul ignore next --@preserve
+    default:
+      throw new Error("Geometry type not supported.");
```
patrickarlt: I would change this to an instance of `ArcGISRequstError`

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    // expected undefined properties
+    expect(arcgis.fields[0].nullable).toBe(undefined);
+    expect(arcgis.fields[0].editable).toBe(undefined);
+    // sqlType not on IFields interface at the moment
+    expect((arcgis.fields[0] as any).sqlType).toBe(undefined);
+    // required IFeature properties (features)
+    expect(arcgis.features.length).toBe(1);
+  });
+
+  test("should convert a pbf single feature POINT to arcgis query features object", async () => {
```
patrickarlt, line 56: Do you want to also check that the X/Y gets converted properly in this test?

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    expect(arcgis.fields[3].defaultValue).toBe(null);
+    // expected undefined properties
+    expect(arcgis.fields[3].nullable).toBe(undefined);
+    expect(arcgis.fields[3].editable).toBe(undefined);
+    expect(arcgis.fields[3].length).toBe(undefined);
+    // sqlType not on IFields interface at the moment
+    expect((arcgis.fields[3] as any).sqlType).toBe(undefined);
+  });
+
+  test("should convert a pbf single feature LINE to arcgis query features object", async () => {
```
patrickarlt, line 95: Do you want to check that geometry get converted correctly here?

packages/arcgis-rest-feature-service/test/query.test.live.ts
```diff
@@ -0,0 +1,802 @@
...
+          outFields: ["*"],
+          resultRecordCount: 1
+        };
+
+        const trailsLinesPbfAsArcGISOptions: IQueryFeaturesOptions = {
+          ...trailsLinesJsonOptions,
+          f: "pbf-as-arcgis"
+        };
+
+        const arcGIS = (await queryFeatures(
```
patrickarlt, line 721: Since you have both the geometries here do you want to do a comparison here to make sure they match?

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -130,7 +135,7 @@ export interface IQueryAllFeaturesOptions extends ISharedQueryOptions {
    * NOTE: for "pbf" you must also supply `rawResponse: true`
    * and parse the response yourself using `response.arrayBuffer()`
    */
-  f?: "json" | "geojson";
+  f?: "json" | "geojson" | "pbf" | "pbf-as-geojson" | "pbf-as-arcgis";
```
gowin20: Is `pbf` an available output type for `queryAllFeatures`? My understanding is that in order to paginate through all features, we have to decode the PBF response and check its properties. The option should probably be removed from this interface.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -368,18 +478,18 @@ export async function queryAllFeatures(
 
     const returnedCount = response.features.length;
 
-    // check if the response has exceededTransferLimit handles both the standard json and geojson responses
     const exceededTransferLimit =
+      // ArcGIS JSON | pbf-as-arcgis: exceededTransferLimit is on the response object
       response.exceededTransferLimit ||
+      // GeoJSON | pbf-as-geojson: exceededTransferLimit is on properties in the response object
       (response as any).properties?.exceededTransferLimit;
```
gowin20, line 487: I believe in the case of `f: pbf`, there is no way to check `exceededTransferLimit` internally (you'd have to parse the response to do that). Seems best to just remove `pbf` as a `queryAllFeatures` option.

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.ts
dixonyant: this was a good catch, for arcgis, there was another nested level of array on single linestring vs multiline string, a diff that the live tests did not catch.

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: I don't believe we support it directly, but technically one could query it at their own risk :) I'll remove it from here, I think when I started the feature I was playing around with direct pbf queries but there's no realistic use for it without copying all the extra heavy-lifting that queryAllFeatures does to make it usable.

packages/arcgis-rest-feature-service/test/query.test.live.ts
dixonyant: I changed the way I check for precisions to be within a margin of difference which is set in an enum for 4326 and 3857. I feel it's more readable this way and more succinct for our purposes.

packages/arcgis-rest-feature-service/src/pbf-parser/arcGISPbfParser.ts
dixonyant: My philosophy on this was that since this file is technically a util/tool and not necessarily related to request, we'd throw a generic error here, as to not depend on or couple it to the request package, and then let the query/request/caller that consumes it catch and re-throw a ArcGISRequestError/custom error with all the extra information and status codes. However, we are importing several types from request in this file already, so it's starting to become tethered anyways, so I'm fine with using ArcGISRequestError here.

packages/arcgis-rest-feature-service/test/pbf-parser/arcGISPbfParser.test.ts
dixonyant: Added geometry structure checks for all geometry type in decoding tests as not just in precision.

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: I removed that option :D

packages/arcgis-rest-feature-service/test/query.test.live.ts
dixonyant: ...and did a precision tolerance comparison, which will return null if the arrays or structure is not the same and break the test.

Issue #1248, opened 2025-08-19
remove `requestOptions.request` as it should not be needed


tomwayson:
I implemented #1220 b/c I was having getting `jest.mock()` to work w/ our packages (any package really), however, I now have it working, and I think that, or `vi.mock()` is a much better option that this library providing a way to mock the low-level `request()` function.

As this will be a breaking change, it should be part of the v5 release.

Issue #1204, opened 2025-03-18
Remove withOptions() for v5


patrickarlt:
@gavinr @tomwayson and I discussed this at DevSummit. This feature war originally added for some functional programming workflows that never materialized. Removing it would reduce our overall surface area without really sacrificing functionality.

Issue #705, opened 2020-05-21
Custom portal not being used when set with `setDefaultRequestOptions()`


COV-GIS:
Setting a custom portal with [IRequestOptions](https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/#portal) via `setDefaultRequestOptions()` is not working for me; making calls to arcgis.com instead of the custom portal. Looks like it should be mixing it in: https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/src/request.ts#L37-L47 and https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/src/request.ts#L195. I've set other `IRequestOptions` with `setDefaultRequestOptions()` to test and they are being successfully mixed in. Is this not the proper method for setting a custom portal? Or is there something I'm missing in the portal package to do so? Or?

~~And a mostly unrelated question for you Esri folks that work with portal a lot. Is there a simple query to get items by folder id? I looked at how portal content fetches items for a folder. It makes several pretty complex queries and ultimately queries the item ids to display. I'm hoping there's an easier way but the lack of a function in the portal package has me thinking not. Thanks.~~

Edit: `ownerfolder: <ID>` works just fine if you remember to pass auth with the request.

jgravois:
> Setting a custom portal with IRequestOptions via setDefaultRequestOptions() is not working for me.

can you share a repro case?

COV-GIS:
Not at this time. But...

```javascript
setDefaultRequestOptions({
  portal: 'https://<URL>/portal/sharing/rest/'
});

searchItems('ownerfolder: ea19...91ac').then();

// request made to arcgis.com
```

```javascript
searchItems({
  q: 'ownerfolder: ea19...91ac',
  portal: 'https://<URL>/portal/sharing/rest/'
}).then();

// request made to custom portal url
```

And any custom request options are inaccessible. https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/src/request.ts#L15

If time permits I may PR `getDefaultRequestOptions()` and `getFolderItems()`.

jgravois:
thanks!

> If time permits I may PR `getDefaultRequestOptions()`.

[...]

PR #1284, opened 2026-01-05
Add deprecation tags and warnings

dixonyant:
Adding deprecation messages to these functionalities in v4 to start warning projects that are using ArcGIS REST JS for breaking changes in v5.

- [remove withOptions()](https://github.com/Esri/arcgis-rest-js/issues/1204)
- [remove setDefaultRequestOptions()](https://github.com/Esri/arcgis-rest-js/issues/705)
- [remove requestOptions.request](https://github.com/Esri/arcgis-rest-js/issues/1248)
- [remove rawResponse](https://github.com/Esri/arcgis-rest-js/issues/1045)

### DIFF
--- a/.changeset/crisp-rockets-sing.md
+++ b/.changeset/crisp-rockets-sing.md
@@ -1,0 +1,5 @@
+---
+"@esri/arcgis-rest-request": patch
+---
+
+Added deprecation messages for upcoming request package updates
--- a/package-lock.json
+++ b/package-lock.json
@@ -103,7 +103,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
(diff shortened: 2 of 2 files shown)


packages/arcgis-rest-request/test/request.test.ts
```diff
@@ -416,9 +416,11 @@ describe("request()", () => {
       true
     );
 
-    expect(warnSpy).toHaveBeenCalledTimes(1);
+    // expect(warnSpy).toHaveBeenCalledTimes(1);
+    // with deprecation messages warn is called twice more
+    expect(warnSpy).toHaveBeenCalledTimes(3);
```
gavinr-maps, line 421: Did you intend to commit this? it seems possibly unrelated?

gavinr-maps (approved):
Overall looks good to me!

packages/arcgis-rest-request/test/request.test.ts
dixonyant: it's kind of a weird test, it tests if the console outputted a different auth warning. But with the deprecation warning added, console.warn is called once for the auth warning and twice for the deprecation warning, as the setDefault... method is called twice in the test. I think the test should go away altogether when the method is deprecated and removed.

PR #1304, opened 2026-03-30
PBF Parser geometry Z coordinates and M values

dixonyant:


packages/arcgis-rest-feature-service/test/query.test.live.ts
```diff
@@ -466,8 +466,8 @@ describe("queryFeatures() and queryAllFeatures() live tests", () => {
         expect(arcgis.spatialReference.latestWkid).toBe(3857);
         expect(arcgis.fields.length).toBe(8);
         expect(arcgis.fieldAliases).toBe(undefined);
-        expect(arcgis.hasZ).toBe(false);
-        expect(arcgis.hasM).toBe(false);
+        expect(arcgis.hasZ).toBe(undefined);
```
dixonyant, line 469: hasZ and hasM values should not return if not queried to match json request behavior

patrickarlt (approved):
LGTM! Thanks for the tests!

PR #1307, opened 2026-04-04
Allow token check for decorated thumbnail to fail

patrickarlt:
This fixes a bug reported internally where `getItem()` will throw an error because it cannot get a token when using LDAP auth since the browser manages auth by sending cookies with the requests rather then storing a token.


dixonyant (approved):
LGTM!

PR #1308, opened 2026-04-06
Remove arcgis-rest-fetch and arcgis-rest-form-data

### LINKED ISSUE #1293: Remove node-fetch, arcgis-rest-fetch and arcgis-rest-form-data
Since we are going to be on Node 22+ for REST JS v5 we can fully rely on Nodes built in `fetch` and `FormData` so these old compatibility packages can be removed.

dixonyant:
In v5, we are supporting node 22+ only, and don't need polyfills to provide fetch, File, Blob, and FormData. In this PR we remove the compatibility packages and rely on node and browsers to provide them globally.


gavinr-maps (approved):
Looks good to me.

PR #1311, opened 2026-04-13
Update IRequestOptions interface to v5 spec

### LINKED ISSUE #880: add fetchOptions to IRequestOptions to replace fetch, httpMethod, credentials, etc
We should replace the individual [`IRequestOptions`](https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions) like `httpMethod` and `credentials` that control the [`fetch()` init object](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#syntax) with a single parameter like `fetchOptions` or `fetchInit` that will be passed directly to fetch. This should resolve issues like #810 and also allow us to remove passing in a custom implementation of `fetch()` via the `fetch` option.

Removing the existing parameters would be a breaking change. We could in theory add the new option first and deprecate the old ones until we later remove them as part of a larger major bump.

### LINKED ISSUE #1292: Update the `IRequestOptions` interface to seperate fetch options from ArcGIS REST JS options
For v5 we want to move most request and fetch options to separate `requestOptions` (option flags used in our internal `request` function) and `fetchOptions` (options passed to the internal call to `fetch`). The only things I would keep would at the top level would be `authentication` and `params`. This would mean our main `IRequestOptions` interface looks like this:

  ```js
  {
    params: Record<string, any>; // any additional params to append to the request
    authentication: IAuthenticationManager | string; // auth manager or access token
    requestOptions?: { // additional options for our intenal request method
      hideToken?: boolean; // put the token param in the header for GET requests
      suppressWarnings?: boolean // silence all internal console warnings from REST JS 
      portal?: string; // used internally by some requests to construct URLs, will override the `portal` set in the authentication option
      username?: string
      injectRequestHeaders?: boolean; // add a `arcgisRestRequestHeaders` property that returns the request headers to resolve https://github.com/Esri/arcgis-rest-js/issues/1181
    };
    /** 
     * anything you can pass to the options for fetch 
     * https://developer.mozilla.org/en-US/docs/Web/API/RequestInit 
     * REST JS may override or ignore these as it sees fit. REST JS
     * currently modifies the headers and credentials options.
     **/
    fetchOptions: RequestInit 

  }
  ```

This would make it so that we can easily copy options around inside higher level methods where we make multiple internal calls to different endpoints like `createApiKey()` by using `structuredClone()`

There are a few other things to note:

* This change should be backward compatible with the existing `IRequestOptions` so the new interface above should be added to it. Existing properties should all become optional and log a warning that they are deprecated when used.
* The only option that we are going to outright remove is `rawResponse`. Instead we should export a new `rawRequest()` method which will simply perform the request and return the `Response` object.
* There is no more `f` param on `request()` which will now only produce JSON. If you need a non-JSON response you can use `rawRequest()` and process the data yourself which would be the case for things like Feature layer queries as PBF etc...
Finally it is OK if we override some of the users `fetchOptions` if we need too. We just need to document around this. For example we currently modify various headers, the `credentials` options etc... because we need too but users should be able to set custom headers, abort signal etc...

Format

dixonyant:
Moving all request options into a legacy options interface that IRequestOptions will extend. IRequestOptions will only expose `params`, `authentication`, `portal`, `requestOptions`, and `fetchOptions`. All legacy options will be deprecated and some will be removed altogether in coming pr's.

Closes: #1292, #880


.changeset/seven-lands-lose.md
```diff
@@ -0,0 +1,15 @@
...
+"@esri/arcgis-rest-elevation": major
+"@esri/arcgis-rest-geocoding": major
+"@esri/arcgis-rest-request": major
+"@esri/arcgis-rest-routing": major
+"@esri/arcgis-rest-places": major
+"@esri/arcgis-rest-portal": major
+"@esri/arcgis-rest-auth": major
+---
+
+Changed IRequestOptions interface to have only necessary top level properties, moved fetch and request options to fetchOptions and requestFlags
```
patrickarlt: @dixonyant Can you enumerate all the options that changed here and what they changed to?

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -5,65 +5,55 @@ import { IAuthenticationManager } from "./IAuthenticationManager.js";
 // NOTE: the `requestOptionsKeys` array in ./append-custom-params.ts
 // must be kept in sync with this interface
 /**
- * Options for the `request()` method.
+ * Legacy request option properties kept for backwards compatibility.
+ *
+ * @deprecated Use the new `requestFlags` object on `IRequestOptions` where applicable.
```
patrickarlt:
```suggestion
 * @deprecated This class represents the v4 @linkcode{IRequestOptions} which should be used instead.
```

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -5,65 +5,55 @@ import { IAuthenticationManager } from "./IAuthenticationManager.js";
...
  */
-export interface IRequestOptions {
-  /**
-   * Additional parameters to pass in the request.
-   */
-  params?: IParams;
+export interface _ILegacyRequestOptions {
```
patrickarlt:
```suggestion
export interface ILegacyRequestOptions {
```

We don't really use `_` to represent private things in the API. I think that having "Legacy" here is enough.

packages/arcgis-rest-request/src/request.ts
patrickarlt, line 57: Note that this marked as deprecated so we probably could just remove this. I'm going to propose a new global config system in a separate issue.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -496,9 +500,6 @@ export function internalRequest(
           });
       }
       if (rawResponse) {
-        console.warn(
```
patrickarlt, line 499: `rawResponse` is actually deprecated to I think we need to rethink the roles of `internalRequest()` and `request()`.

Currently `request()` exists to handle the request option and all the logic happens in `internalRequest()`.

I think with our new model this will be different. We know `request()` will always be getting JSON so all the error and response processing, error retry, error checking etc... should probably happen in `request()`.

Then inside `internalRequest()` we can do all the processing that needs to happen before the request like default options, etc... basically anything that we don't need to know about the response for.

Another thing that might help would be to break out as much of this logic as possible into utility functions that we can test in smaller increments.

patrickarlt (changes requested):
@dixonyant I have a few small things in the PR plus a request about refactoring the `internalRequest()` logic. Which could be in this PR or the future PR where you add `rawRequest()`.

packages/arcgis-rest-request/src/request.ts
dixonyant: I 100% agree. Right now request is just calling internalRequest with a retry loop. Agreed that a lot of it can be broken down into helpers and updated to async/await syntax for legibility. I was going to handle gutting the rawResponse with adding rawRequest in the same future PR.

.changeset/seven-lands-lose.md
dixonyant: what will happen to "maxUrlLength"? @patrickarlt will it need to be removed or just moved somewhere else?

packages/arcgis-rest-request/src/request.ts
patrickarlt: Sounds good we can merge this is an you can follow up with:

* Add `rawRequest()`
* Refactor `internalRequest()` and `request()` logic to support `rawRequest()`

dixonyant: i'll put these in the rawRequest story

dixonyant (commented):
added comments

packages/arcgis-rest-request/src/request.ts
dixonyant: i removed this line in request but it's not showing this as outdated

.changeset/seven-lands-lose.md
```diff
@@ -0,0 +1,46 @@
...
+"@esri/arcgis-rest-request": major
+"@esri/arcgis-rest-routing": major
+"@esri/arcgis-rest-places": major
+"@esri/arcgis-rest-portal": major
+"@esri/arcgis-rest-auth": major
+---
+
+### What
+
+This is a breaking change to `IRequestOptions`.
```
patrickarlt, line 17: Technically the only breaking change will be the removal of `rawRequest` correct? Since this is mostly backward compatible it would be good to note this.

packages/arcgis-rest-request/src/utils/IRequestOptions.ts
```diff
@@ -5,65 +5,55 @@ import { IAuthenticationManager } from "./IAuthenticationManager.js";
...
  */
-export interface IRequestOptions {
-  /**
-   * Additional parameters to pass in the request.
-   */
-  params?: IParams;
+export interface ILegacyRequestOptions {
   /**
    * The HTTP method to send the request with.
+   * @deprecated This top-level option is legacy and will be removed in a future release.
```
patrickarlt:
```suggestion
   * @deprecated This top-level option is legacy and will be removed in a future release. Use `fetchOptions.httpMethod` in @linkcode{IRequestOptions} instead.
```

patrickarlt: @dixonyant it would be good to reference the new options and flags with a link.

patrickarlt (approved):
@dixonyant minor feedback on the release notes and some comments but I think this is good to go from a functionality perspective.

PR #1312, opened 2026-04-16
Remove request method from IRequestOptions properties 

### LINKED ISSUE #1248: remove `requestOptions.request` as it should not be needed

dixonyant:
Removed option to pass in custom request function to be called by RESTJS request() via the IRequestOptions request options. This was only used for testing and with new testing upgrades, this pattern should be no longer necessary. 

Closes #1248


patrickarlt (approved):
LGTM!

PR #1315, opened 2026-05-13
Request deprecation messages should honor suppressWarnings

dixonyant:
Deprecation messages should not warn when suppressWarnings option is true.

### DIFF
--- a/packages/arcgis-rest-request/src/request.ts
+++ b/packages/arcgis-rest-request/src/request.ts
@@ -45,9 +45,11 @@ export function setDefaultRequestOptions(
   options: IRequestOptions,
   hideWarnings?: boolean
 ) {
-  console.warn(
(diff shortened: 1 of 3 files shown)


packages/arcgis-rest-request/src/request.ts
```diff
@@ -45,7 +45,9 @@ export function setDefaultRequestOptions(
   options: IRequestOptions,
   hideWarnings?: boolean
 ) {
-  console.warn(`setDefaultRequestOptions() is deprecated. This will be removed in ArcGIS REST JS v5.0.`);
+  console.warn(
+    `setDefaultRequestOptions() is deprecated. This will be removed in ArcGIS REST JS v5.0.`
+  );
```
gowin20: Does this warning need to be suppressed as well? Not sure of the context here.

gowin20 (approved):
Looks good to me. Just one question - feel free to disregard if you have context I'm missing.

PR #1317, opened 2026-05-15
Introduce `rawRequest()` to replace `rawResponse`

dixonyant:
This PR introduces `rawRequest()` to replace `rawResponse: true` in rest-request. It also implements the new IRequestOptions interface across request and request tests. 
It outright removes support for `rawResponse` and `maxUrlLength` while retaining backwards-compatibility for legacy IRequestOptions. 

Users will have to update their code to use `requestFlags.ignoreMaxUrlLength` and manually request a raw response using `rawRequest()` and handle the response on their own.

**Raw Request:**
	- to be used to request a response as html, text, pbf, or BLOB.

**IRequestOptions:**
	- moved http related options to `requestOption.fetchOptions`
	- moved boolean flags to `requestOptions.requestFlags`
	- refactored `maxUrlLength` case to a new workflow that uses  `requestFlags.ignoreMaxUrlLength` when users want to force a GET request with a very long url length (due to long query params or other large stringified properties appended to url)
	- removed `request`from IRequestOptions as it was an artifact from jasmine era testing days

Added `normalizeDeprecatedRequestOptions()` as a way to support backwards compatibility for ILegacyRequestOptions (except for `rawResponse` and `maxUrlLength` which we are fully removing from request in this pr).


packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -78,7 +79,7 @@ export interface IQueryFeaturesOptions extends ISharedQueryOptions {
   returnExceededLimitFeatures?: boolean;
   /**
    * Response format. Defaults to "json"
-   * NOTE: for "pbf" you must also supply `rawResponse: true`
+   * NOTE: for "pbf" you must use the method `rawRequest()`
    * and parse the response yourself using `response.arrayBuffer()`
```
patrickarlt: @dixonyant since this is all handled internally for the user I don't think that we need to mention `rawRequest()` at all since it is used internally.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -270,24 +270,22 @@ export function queryPbfAsGeoJSONOrArcGIS(
  * ```
  *
  * @param requestOptions - Options for the request
- * @returns A Promise that will resolve with the feature or the [response](https://developer.mozilla.org/en-US/docs/Web/API/Response) itself if `rawResponse: true` was passed in.
+ * @returns A Promise that resolves with the feature by default, or with the native Response when `rawResponse` is `true`.
```
patrickarlt, line 271: @dixonyant I think that we should probably just remove `rawResponse` from `IGetFeatureOptions` and have this always return a `Promise<IFeature>`. If people ask we can introduce a `rawGetFeature()` method later.

packages/arcgis-rest-geocoding/src/bulk.ts
```diff
@@ -94,13 +95,17 @@ export function bulkGeocode(
     );
   }
 
+  if (options.rawResponse) {
+    return rawRequest(
+      `${cleanUrl(options.endpoint)}/geocodeAddresses`,
+      options
+    );
+  }
+
```
patrickarlt, line 104: @dixonyant since we don't want to support a `rawResponse` option anymore I think that we can remove this and add a `rawBulkGeocode()` method later is users need it.

packages/arcgis-rest-geocoding/src/geocode.ts
```diff
@@ -134,12 +135,13 @@ export function geocode(
     }
   }
 
+  if (typeof address !== "string" && address.rawResponse) {
+    return rawRequest(`${cleanUrl(endpoint)}/findAddressCandidates`, options);
+  }
+
```
patrickarlt, line 141: @dixonyant same here. This can also be removed and re-added as `rawGeocode()` later if needed.

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -110,6 +111,7 @@ export function getItemData(
 
   if (options.file) {
     options.params.f = null;
+    return rawRequest(url, options).then((response) => response.blob());
```
patrickarlt, line 114: @dixonyant I don't actually think that `options.file` is a valid option here. https://developers.arcgis.com/rest/users-groups-and-items/item-data/

I think that we can replace this with a `rawGetItemData()` function later for when users want to get the raw item data.

packages/arcgis-rest-request/src/utils/warn-deprecated-request-options.ts
```diff
@@ -30,19 +30,32 @@ const DEPRECATED_REQUEST_OPTION_REPLACEMENTS: Partial<
...
+  const unsupportedWarning = UNSUPPORTED_REQUEST_OPTION_WARNINGS[key];
+  if (unsupportedWarning) {
+    return unsupportedWarning;
+  }
+
   const replacement = DEPRECATED_REQUEST_OPTION_REPLACEMENTS[key];
-  return `${key} is deprecated as a top-level request option and will be removed in ArcGIS REST JS v5.0.${
-    replacement ? ` Use ${replacement} instead.` : ""
-  }`;
+  return `${key} is deprecated as a top-level request option and will be removed in ArcGIS REST JS v5.0. Use ${replacement} instead.`;
```
patrickarlt:
```suggestion
  return `${key} is deprecated as a top-level request option has been be removed in ArcGIS REST JS v5.0. Use ${replacement} instead.`;
```

We are in REST JS 5 now 😄

packages/arcgis-rest-request/src/request.ts
```diff
@@ -322,238 +335,266 @@ export function internalRequest(
...
+      if (token.length && requestFlags?.hideToken) {
         params.token = token;
+        // Remove existing header that was added before url query length was checked
+        delete requestHeaders["X-Esri-Authorization"];
       }
+    } else {
+      // just use GET
+      url = urlWithQueryString;
+    }
+  }
```
patrickarlt, line 464: @dixonyant the logic for this looks good but it might be a good idea to break the out into a helper function that we can test seperatly because the logic is a bit messy here.

patrickarlt (changes requested):
@dixonyant this LGTM but I would take the advice of the Copilot bot and handle the header object in other formats then just a plain object.

packages/arcgis-rest-request/src/request.ts
```diff
@@ -211,55 +206,105 @@ export function checkForErrors(
...
+}> {
+  const options = normalizeRequestOptions(requestOptions);
 
   const params: IParams = {
     ...{ f: "json" },
     ...options.params
   };
 
-  let originalAuthError: ArcGISAuthError = null;
+  const requestFlags = options.requestFlags;
```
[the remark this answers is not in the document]
dixonyant: made requestFlags?.hideToken instead

packages/arcgis-rest-portal/src/items/get.ts
dixonyant: It's supposedly on the IItemDataOptions extends IRequestOptions interface that is passed into this function, although I definitely agree it seems super funky this way and should be brought out like the other raw request ones. I will put this feedback in the story that pulls all of the raw requests out to their own methods.

packages/arcgis-rest-geocoding/src/geocode.ts
dixonyant: should I remove it in this PR or move this feedback to an item in the story that updates all the packages to honor v5 requestOptions?

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: I have a story up to bump all these packages to v5 and implement rawRequest versions in one swoop.

packages/arcgis-rest-geocoding/src/bulk.ts
dixonyant: I put this instance in the package update story as well

packages/arcgis-rest-portal/src/items/get.ts
dixonyant: this one i put in as well

packages/arcgis-rest-request/src/utils/normalize-deprecated-request-options.ts
```diff
@@ -0,0 +1,63 @@
...
+  const fetchOptions = {
+    ...(httpMethod !== undefined ? { method: httpMethod } : {}),
+    ...(credentials !== undefined ? { credentials } : {}),
+    ...(signal !== undefined ? { signal } : {}),
+    ...requestOptions.fetchOptions,
+    headers: {
+      ...(headers as any),
+      ...(requestOptions.fetchOptions?.headers as any)
+    }
+  };
```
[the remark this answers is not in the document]
dixonyant: I created helper methods to cover this in a more comprehensive way since it is needed in multiple places. We can reevaluate once all legacy shape options and normalization functions are removed if we need to keep it around still. (there should be one or two direct calls to merge headers in request standalone)

packages/arcgis-rest-request/src/request.ts
```diff
@@ -56,6 +57,7 @@ export function setDefaultRequestOptions(
 export function getDefaultRequestOptions() {
   return (
     (globalThis as any).DEFAULT_ARCGIS_REQUEST_OPTIONS || {
+      //fetchOptions: { method: "POST" },
       httpMethod: "POST",
       params: {
```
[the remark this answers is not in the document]
dixonyant: I updated both get and set defaultRequestOptions() functions to normalize options going into and out of the global namespace, removed the `httpMethod` and will use fetchOptions.method instead

packages/arcgis-rest-request/src/utils/warn-deprecated-request-options.ts
```diff
@@ -30,19 +30,32 @@ const DEPRECATED_REQUEST_OPTION_REPLACEMENTS: Partial<
...
+  if (unsupportedWarning) {
+    return unsupportedWarning;
+  }
+
   const replacement = DEPRECATED_REQUEST_OPTION_REPLACEMENTS[key];
-  return `${key} is deprecated as a top-level request option and will be removed in ArcGIS REST JS v5.0.${
-    replacement ? ` Use ${replacement} instead.` : ""
-  }`;
+  return `${key} is deprecated as a top-level request option has been be removed in ArcGIS REST JS v5.0. Use ${replacement} instead.`;
 }
```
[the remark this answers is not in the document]
dixonyant: this whole thing will go away with v5 release

packages/arcgis-rest-request/src/request.ts
```diff
@@ -322,238 +406,235 @@ export function internalRequest(
...
+): Promise<any> {
+  // -----------------------------
+  const suppressWarnings =
+    requestOptions.requestFlags?.suppressWarnings ??
+    requestOptions.suppressWarnings ??
+    false;
+  const formatIsNotJson =
+    requestOptions?.params?.f &&
+    requestOptions.params.f !== "json" &&
+    requestOptions.params.f !== "geojson";
```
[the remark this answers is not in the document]
dixonyant: fixed

packages/arcgis-rest-request/src/AuthenticationManagerBase.ts
```diff
@@ -88,8 +88,7 @@ class AuthenticationManagerBase {
       const options = {
         httpMethod: "GET",
         authentication: this,
-        ...requestOptions,
-        rawResponse: false
+        ...requestOptions
       } as IRequestOptions;
```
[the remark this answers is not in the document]
dixonyant: These and all the below instances are all part of this larger wip change that is going into the v5.0.0 branch. The focus of this PR is to handle options at the request/rawRequest level. there is a follow up story that covers all the package-level and repo-uses of LegacyRequestOptions. These will all be done by the time v5.0.0 -> main  merges into main.

packages/arcgis-rest-geocoding/src/geocode.ts
```diff
@@ -134,12 +135,13 @@ export function geocode(
     }
   }
 
+  if (typeof address !== "string" && address.rawResponse) {
+    return rawRequest(`${cleanUrl(endpoint)}/findAddressCandidates`, options);
+  }
```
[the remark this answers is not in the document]
dixonyant: I have this work grouped into a different v5 PR

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -365,6 +361,11 @@ export function queryFeatures(
     queryOptions.params?.f === "pbf-as-arcgis"
   ) {
     return queryPbfAsGeoJSONOrArcGIS(requestOptions.url, queryOptions);
+  } else if (queryOptions.params?.f === "pbf") {
+    return rawRequest(
+      `${cleanUrl(requestOptions.url)}/query`,
+      queryOptions
+    ) as Promise<any>;
   }
```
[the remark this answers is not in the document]
dixonyant: Fixing this in a different story

packages/arcgis-rest-geocoding/src/bulk.ts
```diff
@@ -94,13 +95,17 @@ export function bulkGeocode(
     );
   }
 
+  if (options.rawResponse) {
+    return rawRequest(
+      `${cleanUrl(options.endpoint)}/geocodeAddresses`,
+      options
+    );
+  }
```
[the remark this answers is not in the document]
dixonyant: different PR fixes this wip

packages/arcgis-rest-request/src/request.ts
dixonyant: done.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -270,7 +270,7 @@ export function queryPbfAsGeoJSONOrArcGIS(
  * ```
  *
  * @param requestOptions - Options for the request
- * @returns A Promise that will resolve with the feature or the [response](https://developer.mozilla.org/en-US/docs/Web/API/Response) itself if `rawResponse: true` was passed in.
+ * @returns A Promise that will resolve with the feature.
  */
 export function getFeature(
   requestOptions: IGetFeatureOptions
```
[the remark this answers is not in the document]
patrickarlt: @dixonyant copilot misunderstood tho goals here. Since we want to deprecate `rawResponse` as an option everywhere this should only return a `Promise<IFeature>` and we can introduce a `getRawFeature()` later (post v5) if people really need it.

So to resolve this `rawResponse` should actually be removed from `IGetFeatureOptions`.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -365,6 +361,11 @@ export function queryFeatures(
     queryOptions.params?.f === "pbf-as-arcgis"
   ) {
     return queryPbfAsGeoJSONOrArcGIS(requestOptions.url, queryOptions);
+  } else if (queryOptions.params?.f === "pbf") {
+    return rawRequest(
+      `${cleanUrl(requestOptions.url)}/query`,
+      queryOptions
+    ) as Promise<any>;
```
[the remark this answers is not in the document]
patrickarlt: @dixonyant this also seems off. The goal is that all methods should return JSON by default and this breaks the pattern by returning `Promise<any>` if `f=pbf` so if someone really wants `f=pbf` they will neet to use a new `rawQueryFeatures()` that will need to be added at v5

patrickarlt (changes requested):
@dixonyant There still seem to be a few cases where methods can return non JSON objects. These could get resolved now or in a future issue if you want to make a follow up issue.

packages/arcgis-rest-feature-service/src/query.ts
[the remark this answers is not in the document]
dixonyant: Yes it is off. This in-between behavior shouldn't even see the light of day since the follow-up PR is going to be all the removing of rawRequest calls from all functions that should be only returning JSON. They either get a replacement rawFunction call after v5 or with v5. Just the purpose of this PR is to replace the guts of all calls to `request({rawResponse})` with rawRequest for immediate backwards compatibility internally, and then I was going to remove fully and assert types in the next follow-up PR.

packages/arcgis-rest-geocoding/src/geocode.ts
dixonyant: removed in pr #1326 without replacement for now

packages/arcgis-rest-geocoding/src/bulk.ts
dixonyant: removed without replacement in #1326

packages/arcgis-rest-feature-service/src/query.ts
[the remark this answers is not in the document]
dixonyant: resolved in #1326

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: `pbf` has been changed to only be queryable with `queryFeaturesRaw()`

PR #1320, opened 2026-06-12
appendCustomParams normalize to IRequestOptions v5

dixonyant:
- added IRequestFlags interface


packages/arcgis-rest-request/src/request.ts
```diff
@@ -73,6 +75,46 @@ export function getDefaultRequestOptions(): IRequestOptions {
...
+        ...defaultRequestOptions.requestFlags,
+        ...normalizedRequestOptions.requestFlags
+      },
+      fetchOptions: {
+        ...defaultRequestOptions.fetchOptions,
+        ...normalizedRequestOptions.fetchOptions,
+        headers: {
+          ...(defaultRequestOptions.fetchOptions?.headers as any),
+          ...(normalizedRequestOptions.fetchOptions?.headers as any)
+        }
```
patrickarlt, line 112: @dixonyant I think that this needs to use the new merge header util that was added at some point. Unless `normalizedRequestOptions` and `defaultRequestOptions` both return the headers as an object?

patrickarlt (changes requested):
@dixonyant one change.

packages/arcgis-rest-request/src/request.ts
patrickarlt: Actually looking at this is is normalize to an object so this works.

PR #1326, opened 2026-06-26
Remove `rawResponse` support from REST JS packages

dixonyant:
## Description

This PR propagates the recent `rawResponse` deprecation changes across the library. `rawResponse` is no longer supported in `requestOptions` in `request()` and users must now use `rawRequest()` to get native responses. 
This PR also updates the library to use the new IRequestOptions shape `fetchOptions` object instead of legacy properties.

This change will allow for consistent and guaranteed return types across any wrapper functions that call `request`. This change also guarantees that all `request` responses return as `json`. This PR drops support for `requestOptions.rawResponse` across REST JS. 

Users will be able to get native responses through raw versions of wrapper methods like `getItemDataRaw()` or `queryFeaturesRaw()` and parse the response themselves. Format params that are not `json` (like `f=pbf` or binary return types like `blob` or `text`) will need to use methods that wrap `rawRequest()` and support those format types separately. 

## Breaking Changes

Query
- `getFeature()` no longer supports raw response.
- `queryFeatures()` can no longer be used to query features as `pbf` with `f=pbf`.

Geocode
- `bulkGeocode()` no longer supports raw response.
- `geocode()` no longer supports raw response.

Get
- `getItemData()` no longer supports raw, file, or binary responses, use `getItemDataRaw()`instead.
- deprecated `IItemDataOptions` since we don't support the `file` property in `getItemData()` anymore. Use `getItemDataRaw()` to get the native response instead.
- `getItemInfo()` no longer supports `rawResponse` as an option and uses `rawRequest()` as default behavior.
- `getItemResource()` no longer supports `rawResponse` as an option and uses `rawRequest()` as default behavior.

## Added Behavior

Query
- Added `queryFeaturesRaw()` to allow users to query features as a raw response and use  format params including `f=pbf` or non-json types. (`f=pbf-as-geojson` and `f=pbf-as-arcgis` remain unchanged and return as json from `queryFeatures()`)

Get
- Added `getItemDataRaw()` to allow users to handle non-json data manually.

### DIFF
--- a/.changeset/cold-teams-sit.md
+++ b/.changeset/cold-teams-sit.md
@@ -33,7 +33,7 @@ This change drops support for `rawResponse` from all packages, updates all packa
 - `queryFeatures()` can no longer be used to query features as pbf with `f=pbf`.
 - `bulkGeocode()` no longer supports raw response.
 - `geocode()` no longer supports raw response.
-- `getItemData()` no longer supports raw, file, or binary responses, use getItemDataRaw()instead.
+- `getItemData()` no longer supports raw, file, or binary responses, use `getItemDataRaw()` instead.
   deprecated IItemDataOptions since we don't support the file property in getItemData() anymore. Use getItemDataRaw() to get the native response instead.
 - `getItemInfo()` no longer supports rawResponse as an option and uses rawRequest() as default behavior.
 - `getItemResource()` no longer supports rawResponse as an option and uses rawRequest() as default behavior.
--- a/.changeset/cold-teams-sit.md
+++ b/.changeset/cold-teams-sit.md
@@ -1,0 +1,45 @@
+---
+"@esri/arcgis-rest-basemap-sessions": major
+"@esri/arcgis-rest-developer-credentials": major
+"@esri/arcgis-rest-elevation": major
+"@esri/arcgis-rest-feature-service": major
+"@esri/arcgis-rest-geocoding": major
+"@esri/arcgis-rest-places": major
+"@esri/arcgis-rest-portal": major
+"@esri/arcgis-rest-request": major
+---
+
+remove support for rawResponse and update packages to new IRequestOptions shape
+
+### Summary
+
+This change drops support for `rawResponse` from all packages, updates all packages to use the new `IRequestOptions` shape, and adds support for raw requests with new methods.
+
+- `rawResponse` support removed from REST JS
+- fetch methods now return JSON only. Non-JSON data can be fetched using `rawRequest()` or new wrapper methods that implement it.
+- added `queryFeaturesRaw()` and `getItemDataRaw()`
+- converted all `httpMethod` instances to the new `fetchOptions.method`
+
+### Why these changes were made
+
+- REST JS methods are being moved to return JSON only in most cases.
+- Generic types will be easier to implement and improve our TypeScript commitment.
+- Improve API surface area and our testing landscape.
+- Separate default function from more advanced uses.
+
+### Breaking Changes
+
+- `getFeature()` no longer supports raw response.
+- `queryFeatures()` can no longer be used to query features as pbf with `f=pbf`.
+- `bulkGeocode()` no longer supports raw response.
+- `geocode()` no longer supports raw response.
+- `getItemData()` no longer supports raw, file, or binary responses, use getItemDataRaw()instead.
+  deprecated IItemDataOptions since we don't support the file property in getItemData() anymore. Use getItemDataRaw() to get the native response instead.
+- `getItemInfo()` no longer supports rawResponse as an option and uses rawRequest() as default behavior.
+- `getItemResource()` no longer supports rawResponse as an option and uses rawRequest() as default behavior.
+
+### Migration guide
+
+- To get a raw response, construct your own request and use `rawRequest()` to fetch a native response.
+- For querying features as `pbf`, use `queryFeaturesRaw()` and supply `f=pbf` in the query options.
+- To manually set an HTTP method, supply `fetchOptions.method` in IRequestOptions instead of top-level `httpMethod`.
--- a/packages/arcgis-rest-feature-service/src/query.ts
+++ b/packages/arcgis-rest-feature-service/src/query.ts
@@ -208,7 +208,6 @@ function prepareQueryFeaturesOptions(
       "f"
     ],
     {
-      fetchOptions: { method: "GET" },
       params: {
         // set default query parameters
         where: "1=1",
@@ -499,7 +498,6 @@ export async function queryAllFeatures(
         "f"
       ],
       {
-        fetchOptions: { method: "GET" },
         params: {
           where: "1=1",
           outFields: "*",
--- a/packages/arcgis-rest-feature-service/src/query.ts
+++ b/packages/arcgis-rest-feature-service/src/query.ts
@@ -80,7 +80,7 @@ export interface IQueryFeaturesOptions extends ISharedQueryOptions {
   /**
    * Response format. Defaults to "json".
    */
-  f?: "json" | "geojson" | "pbf" | "pbf-as-geojson" | "pbf-as-arcgis";
+  f?: "json" | "geojson" | "pbf-as-geojson" | "pbf-as-arcgis";
   /**
    * someday...
    *
@@ -133,7 +133,7 @@ export interface IQueryAllFeaturesOptions extends ISharedQueryOptions {
   returnExceededLimitFeatures?: true;
   /**
    * Response format. Defaults to "json"
-   * NOTE: for "pbf" you must use the method `rawRequest()`
+   * NOTE: for "f=pbf" you must use the method `queryFeaturesRaw()`
    * and parse the response yourself using `response.arrayBuffer()`
    */
   f?: "json" | "geojson" | "pbf-as-geojson" | "pbf-as-arcgis";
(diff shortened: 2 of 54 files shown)


previnWong:
@dixonyant 

Right now we have rawResponse as true in this case.
image

Request is imported from @esri/arcgis-rest-request

Do I need to update this?  looks like I need to use rawRequest?

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -100,20 +102,19 @@ export const getItemBaseUrl = (
...
+  const options: IRequestOptions = {
+    params: {},
+    ...requestOptions,
+    fetchOptions: {
+      method: "GET",
+      ...requestOptions?.fetchOptions
+    }
   };
 
-  if (options.file) {
```
dixonyant, line 112: Removing options.file here and deprecating the interface that uses it to resolve comment here: https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3377126357

they can use `getItemDataRaw()` and parse the `response.blob()` themselves

packages/arcgis-rest-geocoding/src/geocode.ts
```diff
@@ -135,47 +135,43 @@ export function geocode(
     }
   }
 
-  if (typeof address !== "string" && address.rawResponse) {
```
dixonyant, line 138: addresses https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3377115217

packages/arcgis-rest-geocoding/src/bulk.ts
```diff
@@ -90,28 +90,20 @@ export function bulkGeocode(
...
     options.endpoint === ARCGIS_ONLINE_BULK_GEOCODING_URL
   ) {
-    return Promise.reject(
-      "bulk geocoding using the ArcGIS service requires authentication"
+    throw new Error(
+      "bulk geocoding using the ArcGIS service requires authentication."
     );
   }
 
-  if (options.rawResponse) {
```
dixonyant, line 98: addresses https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3377110766

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -133,7 +133,7 @@ export interface IQueryAllFeaturesOptions extends ISharedQueryOptions {
   returnExceededLimitFeatures?: true;
   /**
    * Response format. Defaults to "json"
-   * NOTE: for "pbf" you must use the method `rawRequest()`
+   * NOTE: for "f=pbf" you must use the method `queryFeaturesRaw()`
```
dixonyant, line 136: address feedback in https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3377077226

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -267,22 +336,22 @@ export function queryPbfAsGeoJSONOrArcGIS(
...
   const options: IGetFeatureOptions = {
-    ...{ httpMethod: "GET" },
-    ...requestOptions
+    ...requestOptions,
+    fetchOptions: {
+      method: "GET",
+      ...requestOptions.fetchOptions
+    }
   };
-  if (options.rawResponse) {
```
dixonyant, line 283: addresses https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3423478124

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -305,71 +374,29 @@ export function getFeature(
...
-    }
-  );
-
+  const queryOptions = prepareQueryFeaturesOptions(requestOptions);
   if (
     queryOptions.params?.f === "pbf-as-geojson" ||
     queryOptions.params?.f === "pbf-as-arcgis"
   ) {
     return queryPbfAsGeoJSONOrArcGIS(requestOptions.url, queryOptions);
-  } else if (queryOptions.params?.f === "pbf") {
```
dixonyant, line 364: removing pbf from `queryFeatures()` to address https://github.com/Esri/arcgis-rest-js/pull/1317#discussion_r3423522283 and users can use queryFeaturesRaw() in the case they want to use a native response or native response with `pbf`

dixonyant (commented):
@patrickarlt i tagged each of the changes with the comments on the previous pr #1317 feedback, although there were more changes here. This pr should be the one that covers  or finishes everything related to `rawResponse` for v5. most raw response support has been dropped except for queryFeaturesRaw() and getItemDataRaw() for now. Others can be added later with better typed support. processOptions and typed methods are coming next.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -156,6 +156,75 @@ export interface IQueryResponse {
   objectIds?: number[];
 }
 
+export interface IQueryFeaturesRawOptions
```
dixonyant, line 159: pulling out this options construction code shared between `queryFeatures` and `queryFeaturesRaw`

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -481,6 +507,11 @@ export async function queryAllFeatures(
       }
     );
 
+    queryOptions.fetchOptions = {
```
dixonyant, line 510: do we want to allow users to override or should this always be "GET" for query? @patrickarlt

packages/arcgis-rest-feature-service/src/queryRelated.ts
```diff
@@ -79,6 +78,11 @@ export function queryRelated(
     }
   );
 
+  options.fetchOptions = {
```
dixonyant, line 81: do we want to allow users to override or should this always be "GET" for queryRelated? @patrickarlt

packages/arcgis-rest-feature-service/test/query.test.ts
```diff
@@ -1348,7 +1353,50 @@ describe("queryFeatures(): pbf", () => {
...
     expect(response.ok).toBe(true);
 
-    const rawBuffer = await response.arrayBuffer();
-    expect(rawBuffer.byteLength).toBeGreaterThan(0);
+    const rawBuffer = (await response.arrayBuffer()) as ArrayBuffer;
+    // expect the raw buffer to have a byte length of 443, which is the length of the mock pbf file
+    expect(rawBuffer.byteLength).toBe(443);
+  });
+
+  test("queryFeatures with f=pbf should warn, but will query json and return typed json for f=pbf queries", async () => {
```
dixonyant, line 1361: test case showing if users try and force a pbf request to request they will be warned as unsupported and will still get json

packages/arcgis-rest-portal/src/items/get.ts
```diff
@@ -126,6 +127,44 @@ export function getItemData(
...
+  const options: IRequestOptions = {
+    params: {},
+    ...requestOptions,
+    fetchOptions: {
+      method: "GET",
+      ...requestOptions?.fetchOptions
+    }
+  };
+
+  options.params.f = null;
```
dixonyant, line 164: leaving this here for functional equivalence. What about `f=zip`? https://developers.arcgis.com/rest/users-groups-and-items/item-data/

@patrickarlt

.changeset/cold-teams-sit.md
```diff
@@ -0,0 +1,45 @@
...
+- Improve API surface area and our testing landscape.
+- Separate default function from more advanced uses.
+
+### Breaking Changes
+
+- `getFeature()` no longer supports raw response.
+- `queryFeatures()` can no longer be used to query features as pbf with `f=pbf`.
+- `bulkGeocode()` no longer supports raw response.
+- `geocode()` no longer supports raw response.
+- `getItemData()` no longer supports raw, file, or binary responses, use getItemDataRaw()instead.
```
patrickarlt:
```suggestion
- `getItemData()` no longer supports raw, file, or binary responses, use `getItemDataRaw()` instead.
```

packages/arcgis-rest-feature-service/src/query.ts
patrickarlt: ~I don't think we actually need to set this. Since the user might spec a query that exceeds the character limit we might want to just let this use the default behavior~

I remembered that we only cache things that use `GET` so we do want to force it here.

packages/arcgis-rest-feature-service/src/queryRelated.ts
patrickarlt: According to the REST API doc this method only supports `GET` so lets leave it.

packages/arcgis-rest-places/src/getCategories.ts
```diff
@@ -24,11 +24,18 @@ type successResponse =
  */
 export interface IGetCategoriesResponse extends successResponse {}
 
+type IRequestOptionsWithoutHttpMethod = Omit<
+  IRequestOptions,
+  "fetchOptions"
+> & {
+  fetchOptions?: Omit<RequestInit, "method">;
+};
+
```
patrickarlt, line 33: @dixonyant I notice that this is repeated in both `getCategories.ts` and `getCategory.ts` should we maybe move this type to a shared utility in the request package?

patrickarlt: This would probably be valuable anywhere we want to force GET or POST.

patrickarlt (commented):
@dixonyant I think that this looks good! I had a few comments that could be handled elsewhere in follow up issues but I think that this mostly wraps up dev work on v5!

packages/arcgis-rest-places/src/getCategories.ts
dixonyant: we might as well

Issue #1199, opened 2025-03-08
`findPlacesNearPoint` does not use abort signal


mpayson:
### Describe the bug

`findPlacesNearPoint` does not use the abort signal that is passed to it

### Reproduction

The following code snippet logs the response, rather than an abort error

```js
import { ApiKeyManager } from "@esri/arcgis-rest-request";
import { findPlacesNearPoint } from "@esri/arcgis-rest-places";

const apiKey = "<replace me>";

const abortController = new AbortController();
findPlacesNearPoint({
  x: -73.735152,
  y: 40.9384275,
  radius: 1600 * 5, // 5 miles in meters
  categoryIds: ["4d4b7105d754a06372d81259"], // Schools
  authentication: ApiKeyManager.fromKey(apiKey),
  signal: abortController.signal,
})
  .then((response) => {
    console.log(response);
  })
  .catch((error) => {
    console.error(error);
  });

abortController.abort();
```

### Logs

```shell

```

### System Info

```shell
"@esri/arcgis-rest-places": "1.2.0"
```

patrickarlt:
@gavinr it looks like `signal`, `credentials`, `hideToken` and `suppressWarnings` are all missing from https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-request/src/utils/append-custom-params.ts#L14 which is why they aren't getting passed thorough down to `request` internally.

Since this could have a lot of long reaching implications do we want to hold this until REST JS v5 when we refactor options anyway?

gavinr-maps:
@patrickarlt sounds reasonable to me. I've created a new milestone to start tracking v5 items: https://github.com/Esri/arcgis-rest-js/milestone/3

PR #1328, opened 2026-07-09
Add `processOptions()` method

### LINKED ISSUE #657: custom params get dropped in generic search
I was trying to do a `searchItems` with custom params, but they kept getting stripped out. 

It looks like `genericSearch` (which is used by `searchItems`) only allows these params:
https://github.com/Esri/arcgis-rest-js/blob/13bbea4ee995c1a7ba6936ea4ad16c3ff887927f/packages/arcgis-rest-portal/src/util/generic-search.ts#L29-L35

even though `ISearchOptions` type definition looks like it's possible to add other params:
https://github.com/Esri/arcgis-rest-js/blob/13bbea4ee995c1a7ba6936ea4ad16c3ff887927f/packages/arcgis-rest-portal/src/util/search.ts#L8-L13

should item search support all these parameters?
https://developers.arcgis.com/rest/users-groups-and-items/search.htm#GUID-A2C8BFD1-89B2-483F-9275-65D2786F6A6C

dixonyant:
This pull request adds `processOptions()` to replace the `appendCustomParams()` method, utilizing only v5 `IRequestOptions` and deprecates `appendCustomParams()`.

- adds `processOptions(requestOptions, { paramKeys[], extractKeys[], defaultOptions }): IRequestOptions`;
- implements `processOptions()` across packages
- deprecates `appendCustomParams()` with console warnings for users.
- adds tests to assert top level request options pass through `processOptions` to request (should allow us to close #657 and #1199)

### DIFF
--- a/packages/arcgis-rest-developer-credentials/src/updateApiKey.ts
+++ b/packages/arcgis-rest-developer-credentials/src/updateApiKey.ts
@@ -16,7 +16,7 @@ import {
   buildExpirationDateParams
 } from "./shared/helpers.js";
 import { getItem, getPortalUrl, updateItem } from "@esri/arcgis-rest-portal";
-import { appendCustomParams, request } from "@esri/arcgis-rest-request";
+import { processOptions, request } from "@esri/arcgis-rest-request";
 import {
   IApp,
   IGetAppInfoOptions,
@@ -98,11 +98,15 @@ export async function updateApiKey(
     };
     const appResponse = await getRegisteredAppInfo(getAppOption);
     const clientId = appResponse.client_id;
-    const options = appendCustomParams(
-      { ...appResponse, ...requestOptions }, // object with the custom params to look in
-      ["privileges", "httpReferrers"] // keys you want copied to the params object
-    );
-    options.params.f = "json";
+
+    const mergedOptions = {
+      ...appResponse,
+      ...requestOptions
+    };
+    const { requestOptions: options } = processOptions(mergedOptions, {
+      paramKeys: ["privileges", "httpReferrers"],
+      extractKeys: []
+    });
 
     // encode special params value (e.g. array type...) in advance in order to make encodeQueryString() works correctly
     stringifyArrays(options);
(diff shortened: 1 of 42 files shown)


packages/arcgis-rest-developer-credentials/src/updateApiKey.ts
```diff
@@ -98,11 +98,15 @@ export async function updateApiKey(
     };
     const appResponse = await getRegisteredAppInfo(getAppOption);
     const clientId = appResponse.client_id;
-    const options = appendCustomParams(
-      { ...appResponse, ...requestOptions }, // object with the custom params to look in
-      ["privileges", "httpReferrers"] // keys you want copied to the params object
-    );
-    options.params.f = "json";
```
dixonyant, line 105: request now forces f=json for all requests

packages/arcgis-rest-developer-credentials/test/shared/registerApp.test.ts
```diff
@@ -169,8 +169,7 @@ describe("registerApp()", () => {
       redirect_uris: [],
       httpReferrers: ["https://www.esri.com/en-us/home"],
       privileges: [],
-      authentication: authOnline,
-      fetchOptions: { method: "GET" }
```
dixonyant, line 173: This was converted from `httpMethod: "GET"` to the new fetchOptions, and started breaking the tests. The default option of "POST" (set before this line and after in `registerApp()` overrode "GET" because of preference for the default value of `fetchOptions.method` new shape over legacy key of `httpMethod`. I'm not exactly sure why this was "GET" from the beginning as all signs point to this method defaulting to POST and there being no assertion for a "GET" in the test, with the fetch-mock being constructed to match only POST. My analysis is that this is incorrect and should be removed?

packages/arcgis-rest-elevation/test/findElevationAtManyPoints.test.ts
```diff
@@ -22,24 +21,62 @@ describe("findElevationAtManyPoints()", () => {
     });
 
     const [url, options] = fetchMock.lastCall("*");
-
-    expect(response.result).toEqual(atManyPointsDefaultResult.result);
+    expect(url).toContain("/elevation/at-many-points");
```
dixonyant, line 24: This test needed to be stubbed out to truly test falsifiable results. before, it was essentially just asserting fetch-mock return behavior without any end to end verification. Now we test that data goes out a certain way and is return in a specific way.

PR #1329, opened 2026-07-18
Add generic types to REST JS and implement `processOptions()` for strict typing with dynamic options

dixonyant:
This PR adds stronger typing across REST JS by adding generic types to `request()` and a new utility method `processOptions()` to shape custom request options with types to support the different request wrappers across the API.

Process Options utility:

`processOptions()` replaces the `appendCustomParams()` method, utilizing only v5 `IRequestOptions` and deprecates `appendCustomParams()` and bringing strong typing to callers.

- added `processOptions(requestOptions, { paramKeys[], extractKeys[], defaultOptions }): IRequestOptions`;
- replace `appendCustomParams()` across packages with `processOptions()`
- deprecates `appendCustomParams()` with console warnings for users.
- adds tests to assert top level request options pass through `processOptions` to request 
- this change should resolve: 
- #657 
- #1199

Generic types:
- `request()` defaults return type to `unknown` unless specified or inherently typed. 
- `rawRequest()` returns a `Response` object.
- `queryFeatures()` and  `queryAllFeatures()` now utilizes overrides for different query formats.

arcgis-rest-elevation
- populate findElevationAtManyPoints tests.


packages/arcgis-rest-feature-service/test/query.test.ts
```diff
@@ -88,6 +88,39 @@ describe("getFeature() and queryFeatures()", () => {
...
+          geometry: null,
+          properties: { OBJECTID: 1 }
+        }
+      ]
+    });
+
+    const response = (await queryFeatures(requestOptions)) as {
+      type: string;
+      features: Array<{ properties: { OBJECTID: number } }>;
+    };
```
patrickarlt: @dixonyant something looks wrong here. You shouldn't have to force a type override on the result here. The override on `queryFeatures()` for `f: "geojson"` SHOULD work and you should get the `EsriGeoJSONFeatureCollection` type automatically.

packages/arcgis-rest-feature-service/src/query.ts
```diff
@@ -378,17 +381,48 @@ export function getFeature(
...
-    queryOptions.params?.f === "pbf-as-arcgis"
-  ) {
-    return queryPbfAsGeoJSONOrArcGIS(requestOptions.url, queryOptions);
+): Promise<IQueryFeaturesResponse | IQueryResponse>;
+function queryFeatures(
+  requestOptions: IQueryFeaturesOptions
+): Promise<
+  IQueryFeaturesResponse | IQueryResponse | EsriGeoJSONFeatureCollection
+> {
+  const processedOptions = processQueryFeaturesOptions(requestOptions);
```
patrickarlt, line 442: @dixonyant this looks correct and matches up with my example in https://github.com/Esri/arcgis-rest-js/issues/821#issuecomment-4492893427 so the return type must not be getting passed through properly.

patrickarlt (changes requested):
@dixonyant 2 comments.

packages/arcgis-rest-feature-service/src/query.ts
dixonyant: right, the type union for `f` is getting expanded to the largest type (includes f=pbf) through the `processQueryFeaturesOptions` , (because the f -> params pipeline decomposes it to an arbitrary string, down from a type) so the user has to declare an explicit f type in the declaration. I'm going to try and make the process query function handle it by either returning it through extract keys or declaring a specific processQueryFeaturesOptions override for each options type,insltead of intersecting the two options types

dixonyant: update: this is still correct, the caller to this method was widening the `f` param to string by not declaring the options object inline, declaring the `f` param as a type literal using const assertion, or declaring the object type + type literal

packages/arcgis-rest-feature-service/test/query.test.ts
dixonyant: Yes, this is now true, but has to be done through one of the 4 following:

Inline:
```ts
const response = await queryFeatures({
  url: "https://.../FeatureServer/0",
  f: "geojson"
});
// response is EsriGeoJSONFeatureCollection
```
Const Assertion:
```ts
const options = {
   url: "https://.../FeatureServer/0",
   f: "geojson" as const
};

const response = await queryFeatures(options);
// response is EsriGeoJSONFeatureCollection
```
Type Assertion + Literal:
```ts
const typedOptions: IQueryFeaturesOptions & { f: "geojson" } = {
  url: "https://.../FeatureServer/0",
  f: "geojson"
};

const typedResponse = await queryFeatures(typedOptions);
// typedResponse is EsriGeoJSONFeatureCollection
```
`satisfies` keyword:
```ts
const options = {
  url: "https://.../FeatureServer/0",
  f: "geojson"
} satisfies IQueryFeaturesOptions;

const response = await queryFeatures(options);
// response is EsriGeoJSONFeatureCollection
```

patrickarlt (approved):
@dixonyant this looks great! Lets merge this so we can proceed with the v5 release.

PR #1335, opened 2026-08-11
v4.x Maintenance Release

dixonyant:
when this merges it should bump the version to 4.11.1 and release to dist-tag lts-v4.x


patrickarlt (approved):
LGTM. Copilot comments could be addressed

